PR #142, opened 2024-07-17
fix(docker): specify platform in docker-compose.yml

prasad-takale-eminds:


vbabuEM (approved):
Prasad, I was just curious. This AMD64 works also for Intel 64?

prasad-takale-eminds:
> Prasad, I was just curious. This AMD64 works also for Intel 64?

@vbabuEM Yes it will work for both as this is a configuration change to file.

PR #143, opened 2024-07-22
Update the docker-compose file to create the docker logs volume for g…

prasad-takale-eminds:
…ateway and registry


vbabuEM (commented):
Since these are marked as external, dont you want to create them somewhere?

vbabuEM (approved):

Issue #74, opened 2024-05-07
ONIX-GUI not working in Amazon Linux


Mishalabdullah:
# Problem
The installation script `start.sh` does not include commands for installing on Amazon Linux and other Linux distributions, mainly because the script only adds Ubuntu repositories for downloading Docker.

# Goal
The script should work for all Linux distributions.

## Technical Skills Needed:
- Node.js
- React.js

## Category
Bug

## Sub-Category
Frontend, Backend

mayurvir:
@Mishalabdullah  Can you add approach to solve it here as discussed?

viraj89:
Hi @harshcrop - can you please check within the community and take further into the community.

PR #195, opened 2024-08-30
Added Docker Install Support For  Amazon Linux GUI  Script

Mishalabdullah:
This PR will resolve the #74 issue.

In the fix, the GUI script will make use of the `package manager.sh` file to download and manage docker.


nirmay:
Too old and now irrelevant for the new ONIX design. Closing.

PR #399, opened 2025-02-13
Feat/client services

MayurWitsLab:


shared/utils/logger.go
ashishkgGoogle, line 1: move this to shared/log.go and change the package name to log.

then follow the below pattern 
Add a method in this package Info(ctx, string message)
Now wherever you want to log something Import this pacakge and then you can simply call log.Info()

Similarly for Debug, Error, warn and Fatal

ashishkgGoogle (changes requested):
Hey You seemed to have used the wrong branch to for your feature branch from , your need to use the follwoing branch :https://github.com/beckn/beckn-onix/tree/beckn-onix-v1.0-develop

cmd/clientSideReciever/handler/home.go
ashishkgGoogle, line 1: We will have a single handler , so all this should go to  cmd/clientSideReciever/handler.go in package main itself.

cmd/clientSideReciever/handler/home.go
```diff
@@ -13,9 +12,7 @@ type PublishRequest struct {
 
 // HomeHandler handles the home route
 func HomeHandler(w http.ResponseWriter, r *http.Request) {
```
ashishkgGoogle, line 14: Since there will be just one handler we will call it just handler.

cmd/clientSideReciever/handler/home.go
```diff
@@ -24,16 +21,8 @@ func HomeHandler(w http.ResponseWriter, r *http.Request) {
...
-		fmt.Println("req.Message", req.Message)
-		err := handlers.PublishMessage(req.Message)
-		if err != nil {
-			http.Error(w, "Failed to publish message here", http.StatusInternalServerError)
-			return
-		}
-
 		w.WriteHeader(http.StatusOK)
-		fmt.Fprintf(w, "Message published successfully")
+		fmt.Fprintf(w, "Message recieved successfully")
```
ashishkgGoogle, line 25: This is not needed, please just add a log.

cmd/clientSideReciever/handler/home.go
```diff
@@ -13,9 +12,7 @@ type PublishRequest struct {
 
 // HomeHandler handles the home route
 func HomeHandler(w http.ResponseWriter, r *http.Request) {
-	if r.Method == "GET" {
-		fmt.Fprintf(w, "This is a GET request")
-	} else if r.Method == "POST" {
+	if r.Method == "POST" {
 		var req PublishRequest
```
ashishkgGoogle, line 16: What is this ?

cmd/clientSideReciever/handler/home.go
```diff
@@ -24,16 +21,8 @@ func HomeHandler(w http.ResponseWriter, r *http.Request) {
...
 
-		// Publish message
-		fmt.Println("req.Message", req.Message)
-		err := handlers.PublishMessage(req.Message)
-		if err != nil {
-			http.Error(w, "Failed to publish message here", http.StatusInternalServerError)
-			return
-		}
-
 		w.WriteHeader(http.StatusOK)
```
ashishkgGoogle, line 24: this does not seem right, If the method is not post we are returning 200 ok, where as we should be retuning error.

cmd/clientSideReciever/main.go
```diff
@@ -1,22 +1,46 @@
...
+	handlers "beckn-onix/cmd/clientSideReciever/handler"
+	shared "beckn-onix/shared"
+	"context"
+	"flag"
+	"fmt"
+	"log"
 	"net/http"
 )
 
 func main() {
```
ashishkgGoogle, line 14: func main() should only call run(), Please refer to this , https://grafana.com/blog/2024/02/09/how-i-write-http-services-in-go-after-13-years/

cmd/clientSideReciever/main.go
```diff
@@ -1,22 +1,46 @@
...
-	// Load configuration
-	config.LoadEnv()
-	port := config.GetPort()
+
+	// Define the command-line flag for the YAML file path
+	configPath := flag.String("config", "config.yaml", "./config.yaml")
```
ashishkgGoogle, line 17: this should be a package level variable

cmd/clientSideReciever/main.go
```diff
@@ -1,22 +1,46 @@
...
+	configuration, err := config.InitConfig(ctx, *configPath)
+	if err != nil {
+		log.Fatalf("Error initializing config: %v", err)
+	}
+
+	// Use the config to initialize the server
+	fmt.Printf("App Name: %s\n", configuration.AppName)
+	fmt.Printf("Server Port: %d\n", configuration.ServerPort)
+	fmt.Printf("Database Host: %s\n", configuration.DBHost)
+	fmt.Printf("Database Port: %d\n", configuration.DBPort)
```
ashishkgGoogle, line 33: this should be a debug log and we should just print the read config struct.

cmd/clientSideReciever/handler/home.go
MayurWitsLab: I made changes in clientReciever for review.

cmd/clientSideReciever/handler/home.go
MayurWitsLab: okay

PR #404, opened 2025-02-20
test cases , method split for config and server

MayurWitsLab:
coverage: 97.1% of statements


cmd/clientSideHandler/main.go
```diff
@@ -0,0 +1,86 @@
...
+	"context"
+	"flag"
+	"fmt"
+	"net/http"
+	"os"
+
+	// BecknOnix packages
+	log "beckn-onix/log"
+
+	// Third-party packages
```
ashishkgGoogle, line 14: these comments are not needed

cmd/clientSideHandler/main.go
```diff
@@ -0,0 +1,86 @@
...
+
+func requestHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	log.Log.Info("Received request:", r.Method, r.URL.Path, r.Header)
+	w.WriteHeader(http.StatusOK)
+	fmt.Fprintln(w, "Message received successfully")
```
ashishkgGoogle, line 31: remove this

index.html
ashishkgGoogle, line 1: why is this needed ?

README.md
```diff
@@ -1 +1,32 @@
+# Beckn-ONIX
 
+Beckn-ONIX is [FIDE](https://fide.org/) project aimed at easing setup and maintainance of a [Beckn](https://becknprotocol.io/) Network using reference implementations. Objectives include setting up reliable, configurable and fast Beckn network as a virtual appliance. This initiative is independent of the evolution of the Beckn protocol. This effort is also aimed at inviting contributions from the community to create secure, reliable builds for production environments.
+
+> Disclaimer : Beckn-onix is a reference implementation of the Beckn-onix stack. It is a reference application only and has not been tested for production environmens. However, implementers can fork this repository and build it for scale. The maintainer of this repository holds no liabillity for deployments of this application in production environments.
```
ashishkgGoogle, line 5: remove all this

cmd/clientSideHandler/main.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+var configPath string
+
+func main() {
+	flag.StringVar(&configPath, "config", "../../config/clientSideHandler-config.yaml", "../../config/clientSideHandler-config.yaml")
+	flag.Parse()
+
+	if err := run(context.Background(), configPath); err != nil {
+		log.Log.Error("Application failed:", err)
```
ashishkgGoogle, line 84: Add a todo here to change this to a fatal log once the logging package is added.

cmd/clientSideHandler/main_test.go
```diff
@@ -0,0 +1,297 @@
...
+)
+
+
+
+func createTempConfig(t *testing.T, data string) string {
+	t.Helper()
+	tmpFile, err := os.CreateTemp("", "config.yaml")
+	if err != nil {
+		t.Fatalf("Failed to create temp config: %v", err)
+	}
```
ashishkgGoogle, line 20: Don't create test data in unit tests, add a directory for test data and add a test config file in it and use it from there, it will simplify add more tests as well as extending it in future.

cmd/clientSideHandler/main.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return fmt.Errorf("error initializing config: %w", err)
+	}
+
+	port := fmt.Sprintf(":%d", configuration.Port)
+	http.HandleFunc("/", requestHandler)
+
+	server := &http.Server{Addr: port}
+
+	// Run server in a goroutine
```
ashishkgGoogle, line 45: all comments should end with a full stop.

cmd/clientSideHandler/main.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	defer file.Close()
+
+	var config config
+	decoder := yaml.NewDecoder(file)
+	err = decoder.Decode(&config)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshal config data: %v", err)
+	}
+	if config.AppName == "" || config.Port == 0 {
```
ashishkgGoogle, line 70: Please move this to a validate method, as we add more fields to the config this section will grow and become unmanageable if left here. Also please separate the checks and return specific errors. for string fields instead of comparing with an empty string a better check would be. 'len(strings.Trim(filed)) ==0'

cmd/clientSideHandler/main_test.go
```diff
@@ -0,0 +1,297 @@
...
+	if _, err := tmpFile.Write([]byte(data)); err != nil {
+		t.Fatalf("Failed to write to temp config: %v", err)
+	}
+	tmpFile.Close()
+	return tmpFile.Name()
+}
+
+
+func TestInitConfig(t *testing.T) {
```
ashishkgGoogle, line 29: Please split it into 2 tests, one for success and one for error

cmd/clientSideHandler/main_test.go
```diff
@@ -0,0 +1,297 @@
...
+		{
+			name:        "Error - Missing Required Fields",
+			configData:  `appName: ""`,
+			expectError: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			tempFilePath := "../../config/clientSideHandler-config.yaml"
```
ashishkgGoogle, line 62: instead of doing this please add yaml files to a test data directory and read it. it will make addign new test cases much simpler. 
Once you have added a validate method add success and failure cases for it as well.

cmd/clientSideHandler/main_test.go
```diff
@@ -0,0 +1,297 @@
...
+				if err == nil {
+					t.Errorf("Expected error, got nil")
+				}
+			} else {
+				if err != nil {
+					t.Errorf("Did not expect error, got %v", err)
+				}
+				if cfg.AppName != tc.expected.AppName {
+					t.Errorf("Expected appName %s, got %s", tc.expected.AppName, cfg.AppName)
+				}
```
ashishkgGoogle, line 87: This is not the scalable way to do this at the config will grow much bigger in the future, Instead please use cmp.diff for this comparing the entire config object against an expected object.

cmd/clientSideHandler/main_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == "POST" {
+
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte("Message received successfully"))
+		} else {
+			http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		}
+	})
```
ashishkgGoogle, line 127: Why is this needed, from the below tests it seems that we are testing this handler instead of the actual handler used by the server.

cmd/clientSideHandler/main_test.go
```diff
@@ -0,0 +1,297 @@
...
+			}
+
+			if rec.Body.String() != tc.expectedResponse {
+				t.Errorf("Expected response body: %q, got: %q", tc.expectedResponse, rec.Body.String())
+			}
+		})
+	}
+}
+
+
```
ashishkgGoogle, line 148: remove unnecessary empty lines.

cmd/clientSideHandler/main_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+
+func TestRun(t *testing.T) {
```
ashishkgGoogle, line 151: please split this into success and failure cases.

cmd/clientSideHandler/main_test.go
```diff
@@ -0,0 +1,297 @@
...
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			if !tt.expectError {
+				go func() {
+					if err := run(ctx, configPath); err != nil && !tt.expectError {
+						t.Errorf("Unexpected error: %v", err)
+					}
+				}()
+				time.Sleep(500 * time.Millisecond)
```
ashishkgGoogle, line 203: instead of sleep, consider using a channel to signal the end of the go routine,

log/log.go
ashishkgGoogle, line 1: move this to beckn-onyx/shared/log/log.go

ashishkgGoogle (changes requested):
I have added comments only to the handler side , please make sure that the same comments are also handled for the Receiver code

ashishkgGoogle (changes requested):

cmd/clientSideHandler/main.go
MayurWitsLab: done

cmd/clientSideHandler/main_test.go
MayurWitsLab: done

PR #426, opened 2025-03-18
Ft/logging module

MayurWitsLab:


ashishkgGoogle (changes requested):
please add separate test for each exported method, configure it to write to a file, and test the logged message in the file

Deepa-Mulchandani (approved):

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/rs/zerolog"
+	"gopkg.in/natefinch/lumberjack.v2"
+)
+
+type Level string
+type DestinationType string
```
poojajoshi2, line 20: Add a blank line here.

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+	newLogger = zerolog.New(multiwriter).
+		Level(logLevels[config.level]).
+		With().
+		Timestamp().
+		Caller().
+		Logger()
+
+	cfg = config
+	return newLogger, nil
+}
```
poojajoshi2, line 170: Add a blank line here

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+			return
+		}
+
+		logger, initErr = getLogger(c)
+		if initErr != nil {
+			return
+		}
+	})
+	return initErr
+
```
poojajoshi2, line 184: remove this blank line

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+}
```
poojajoshi2, line 185: add a blank line

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+
+func logEvent(ctx context.Context, level zerolog.Level, msg string, err error) {
+	event := logger.WithLevel(level)
+
+	if err != nil {
+		event = event.Err(err)
+	}
+	addCtx(ctx, event)
+	event.Msg(msg)
+}
```
poojajoshi2, line 248: Add a blank line

poojajoshi2 (changes requested):

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+	FatalLevel: zerolog.FatalLevel,
+	PanicLevel: zerolog.PanicLevel,
+}
+
+type Config struct {
+	level        Level         `yaml:"level"`
+	destinations []Destination `yaml:"destinations"`
+	contextKeys  []any         `yaml:"contextKeys"`
+}
+
```
sakshamGoogle, line 54: The config struct fields should be exported

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+	once   sync.Once
+)
+
+var (
+	ErrInvalidLogLevel   = errors.New("invalid log level")
+	ErrLogDestinationNil = errors.New("log Destinations cant be empty")
+	ErrMissingFilePath   = errors.New("file path missing in destination config for file logging")
+)
+
+func (config *Config) Validate() error {
```
sakshamGoogle, line 67: Validate method should not be exported

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+		if err := c.Validate(); err != nil {
+			return
+		}
+
+		logger, initErr = getLogger(c)
+		if initErr != nil {
+			return
+		}
+	})
+	return initErr
```
sakshamGoogle, line 183: If config validation fails, error will never be reported as err is never returned 
err can be replaced initErr and can be handled properly

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+			}
+
+			fmt.Printf("writing test log to file: %v\n", config)
+			lumberjackLogger := &lumberjack.Logger{
+				Filename:   filePath,
+				MaxSize:    500, // Default size in MB if not overridden
+				MaxBackups: 15,  // Number of backups
+				MaxAge:     30,  // Days to retain
+				Compress:   false,
+			}
```
sakshamGoogle, line 131: Lumberjack has its own default values...so no need to define our own

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+	return nil
+}
+
+var defaultConfig = Config{
+	level: InfoLevel,
+	destinations: []Destination{
+		{Type: Stdout},
+	},
+	contextKeys: []any{"userID", "requestID"},
+}
```
sakshamGoogle, line 104: Don't add any context keys in default config

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+	defer func() {
+		if closer, ok := multiwriter.(io.Closer); ok {
+			closer.Close()
+		}
+	}()
+	newLogger = zerolog.New(multiwriter).
+		Level(logLevels[config.level]).
+		With().
+		Timestamp().
+		Caller().
```
sakshamGoogle, line 165: Remove line 165 caller is not needed

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func addCtx(ctx context.Context, event *zerolog.Event) {
+	for _, key := range cfg.contextKeys {
+		val, ok := ctx.Value(key).(string)
+		if !ok {
+			continue
+		}
+		keyStr := key.(string)
+		event.Str(keyStr, val)
```
sakshamGoogle, line 266: use event.Any() here instead of Str

pkg/log/log.go
MayurWitsLab: done

pkg/log/log.go
MayurWitsLab: done

pkg/log/log.go
MayurWitsLab: done

pkg/log/log.go
MayurWitsLab: done

pkg/log/log.go
MayurWitsLab: done

pkg/log/log.go
MayurWitsLab: done

pkg/log/log.go
MayurWitsLab: done

pkg/log/log.go
MayurWitsLab: done

pkg/log/log.go
```diff
@@ -0,0 +1,278 @@
...
+		case Stdout:
+			writers = append(writers, os.Stdout)
+		case File:
+			filePath := dest.Config["path"]
+			dir := filepath.Dir(filePath)
+			if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+				return newLogger, fmt.Errorf("failed to create log directory: %v", err)
+			}
+
+			fmt.Printf("writing test log to file: %v\n", config)
```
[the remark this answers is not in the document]
MayurWitsLab: done

pkg/log/log.go
MayurWitsLab: done

pkg/log/log.go
MayurWitsLab: done

pkg/log/log.go
MayurWitsLab: done

ashishkgGoogle (changes requested):

pkg/log/log_test.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+	err = InitLogger(config)
+	if err != nil {
+		t.Fatalf("failed to initialize logger: %v", err)
+	}
+	return testLogFilePath
+}
+
+func readLogFile(t *testing.T, logPath string) []string {
+	file, err := os.Open(logPath)
```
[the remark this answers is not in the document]
MayurWitsLab: It was done to read file line by line instead of while file at once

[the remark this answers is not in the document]
MayurWitsLab: I have updated these now

pkg/log/log_test.go
```diff
@@ -0,0 +1,535 @@
...
+		lines = append(lines, scanner.Text())
+	}
+	if err := scanner.Err(); err != nil {
+		t.Fatalf("failed to read log file: %v", err)
+	}
+
+	return lines
+}
+
+func parseLogLine(t *testing.T, line string) map[string]interface{} {
```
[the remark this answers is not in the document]
MayurWitsLab: this is needed for parsing the stringified json , it was being used miltiple time so the function was created

pkg/log/log.go
MayurWitsLab: done

pkg/log/log.go
```diff
@@ -0,0 +1,271 @@
...
+	"path/filepath"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/rs/zerolog"
+	"gopkg.in/natefinch/lumberjack.v2"
+)
+
+type Level string
```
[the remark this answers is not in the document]
MayurWitsLab: done

pkg/log/log.go
```diff
@@ -0,0 +1,271 @@
...
+
+type DestinationType string
+
+type Destination struct {
```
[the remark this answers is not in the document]
MayurWitsLab: done

pkg/log/log_test.go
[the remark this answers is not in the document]
MayurWitsLab: done

pkg/log/log_test.go
```diff
@@ -0,0 +1,535 @@
...
+		t.Fatalf("Failed to parse log line: %v", err)
+	}
+	return logEntry
+}
+
+func TestDebug(t *testing.T) {
+	logPath := setupLogger(t, DebugLevel)
+	ctx := context.WithValue(context.Background(), userID, "12345")
+	Debug(ctx, "Debug message")
+	lines := readLogFile(t, logPath)
```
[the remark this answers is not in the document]
MayurWitsLab: we cannot declare with exact string that we are expecting as that contains time also , time will not match when compared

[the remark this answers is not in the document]
MayurWitsLab: > This is not the right way to do this.
> 
> Declare a []byte with exact string that you need as below: var expected:= []byte("Expected log message") var got:= readLogFile(t) if got != expected { t.fatal }
> 
> if you used T.tempDir() it will always return a directory, so you don't have to worry about your tests interfaring with each other as well.

we cannot declare with exact string that we are expecting as that contains time also , time will not match when compared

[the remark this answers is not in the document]
MayurWitsLab: have updated with the approach we discussed

ashishkgGoogle (approved):

PR #427, opened 2025-03-18
Feat/encryption

rupinder-syngh:
- Added encryption plugin


rupinder-syngh:
Adding a new PR for the same

PR #432, opened 2025-03-19
Feature/routing plugin

tanyamadaan:
Routing plugin


pkg/plugin/implementation/router/cmd/plugin.go
```diff
@@ -0,0 +1,31 @@
...
+// RouterProvider provides instances of Router.
+type RouterProvider struct{}
+
+// New initializes a new Router instance.
+func (rp RouterProvider) New(ctx context.Context, config map[string]string) (definition.Router, func() error, error) {
+	if ctx == nil {
+		return nil, nil, errors.New("context cannot be nil")
+	}
+
+	// Parse the routing_config key from the config map
```
ashishkgGoogle, line 20: fix the comment, key name is wrong.

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+		switch rule.RoutingType {
+		case "url":
+			if rule.Target.URL == "" {
+				return fmt.Errorf("invalid rule: url is required for routingType 'url'")
+			}
+			if _, err := url.ParseRequestURI(rule.Target.URL); err != nil {
+				return fmt.Errorf("invalid URL in rule: %w", err)
+			}
+		case "msgq":
+			if rule.Target.TopicID == "" {
```
ashishkgGoogle, line 107: We should not call it TopicID as that would make it specific to the gcp pubsub. We should call it Publisher ID

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// Route determines the routing destination based on the request context.
+func (r *Router) Route(ctx context.Context, url *url.URL, body []byte) (*definition.Route, error) {
+
+	// Parse the body to extract domain and version
+	var requestBody struct {
+		Context struct {
+			Domain  string `json:"domain"`
+			Version string `json:"version"`
```
ashishkgGoogle, line 127: this is something we are doing in schema validator as well, can we just pull this into the model.go and use it from there in both places ?

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+			BppURI  string `json:"bpp_uri,omitempty"`
```
ashishkgGoogle, line 128: When using Acronyms in filed names , all letters should in same case.
So the Fileds should be BAPURI and BPPURI

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+			BapURI  string `json:"bap_uri,omitempty"`
+		} `json:"context"`
+	}
+	if err := json.Unmarshal(body, &requestBody); err != nil {
+		return nil, fmt.Errorf("error parsing request body: %w", err)
```
ashishkgGoogle, line 133: this should return model.BadReq

we should return a badRequest error from this PR https://github.com/beckn/beckn-onix/pull/437/files#

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+				}
+			}
+		}
+	}
+
+	// If domain and version match but endpoint is not found, return an error
+	return nil, fmt.Errorf("endpoint '%s' is not supported for domain %s and version %s", endpoint, requestBody.Context.Domain, requestBody.Context.Version)
+}
+
+// getMatchingRules returns all rules that match the given domain and version
```
ashishkgGoogle, line 177: all comments should end with a full stop.

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// Extract the endpoint from the URL
+	endpoint := path.Base(url.Path)
+
+	// Collect all matching rules for the domain and version
+	matchingRules := r.getMatchingRules(requestBody.Context.Domain, requestBody.Context.Version)
+
+	// If no matching rules are found, return an error
+	if len(matchingRules) == 0 {
+		return nil, fmt.Errorf("no matching routing rule found for domain %s and version %s", requestBody.Context.Domain, requestBody.Context.Version)
```
ashishkgGoogle, line 144: When need to decide weather this will be cosidered a bad request or notFound error and return corresponding error from here

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+	}
```
ashishkgGoogle, line 145: we should also check that endpoint and the request.context.action are same. else we should return a bad request.

pkg/plugin/definition/router.go
```diff
@@ -0,0 +1,24 @@
+package definition
+
+import (
+	"context"
+	"net/url"
+)
+
+// Route defines the structure for the Route returned
+type Route struct {
```
ashishkgGoogle, line 9: we should move this to model.go.

Already being done in this pr https://github.com/beckn/beckn-onix/pull/437/files#

pkg/plugin/definition/router.go
```diff
@@ -0,0 +1,24 @@
...
+	RoutingType string // "url" or "msgq"
```
ashishkgGoogle, line 10: this should be just Type

pkg/plugin/definition/router.go
```diff
@@ -0,0 +1,24 @@
...
+	TopicID     string // For message queues
```
ashishkgGoogle, line 11: this should be PublisherID

pkg/plugin/definition/router.go
```diff
@@ -0,0 +1,24 @@
...
+	TargetURL   string // For API calls
```
ashishkgGoogle, line 12: this can be just Target or Just URL

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+
+// RoutingConfig represents the structure of the routing configuration file.
+type routingConfig struct {
+	RoutingRules []routingRule `yaml:"routingRules"`
+}
+
+// Router implements Router interface
+type Router struct {
+	config *Config
+	rules  []routingRule
```
ashishkgGoogle, line 30: instead of storing a slice rules , if we store it as :
map[string]map[string]map[string]route it will make finding the route much faster and cleaner.

essentially map Domain to=>{map version to=>{map endpoint to route.}}
we can create this from config in load rules and store this . then retriving will simply look like :
route:= rules[domain][version][endpoint]
ofcourse we will have to add nil checks.

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+					return &definition.Route{
+						RoutingType: rule.RoutingType,
+						TopicID:     rule.Target.TopicID,
+					}, nil
+				case "url":
+					return &definition.Route{
+						RoutingType: rule.RoutingType,
+						TargetURL:   rule.Target.URL,
+					}, nil
+				case "bpp":
```
ashishkgGoogle, line 162: we should define enums for this.

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+					return handleRouting(rule, requestBody.Context.BppURI, endpoint, "bpp")
+				case "bap":
+					return handleRouting(rule, requestBody.Context.BapURI, endpoint, "bap")
+				default:
+					return nil, fmt.Errorf("unsupported routingType: %s", rule.RoutingType)
```
ashishkgGoogle, line 167: this should be a not found error.

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+// handleRouting handles routing for bap and bpp routing type
+func handleRouting(rule routingRule, uri, endpoint string, routingType string) (*definition.Route, error) {
+	if uri == "" {
+		if rule.Target.URL != "" {
+			return &definition.Route{
+				RoutingType: routingType,
+				TargetURL:   rule.Target.URL,
+			}, nil
+		} else {
+			return nil, fmt.Errorf("no target URI or URL found for %s routing type and %s endpoint", routingType, endpoint)
```
ashishkgGoogle, line 197: This error will reach the caller, perhapes we should reword it to make it easier to understand for them. they would not understand what is target uri what is routing type.

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+
+// handleRouting handles routing for bap and bpp routing type
+func handleRouting(rule routingRule, uri, endpoint string, routingType string) (*definition.Route, error) {
+	if uri == "" {
+		if rule.Target.URL != "" {
+			return &definition.Route{
+				RoutingType: routingType,
+				TargetURL:   rule.Target.URL,
+			}, nil
+		} else {
```
ashishkgGoogle, line 196: this else is not needed as we are returning from the above if.

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+		if rule.Domain == domain && rule.Version == version {
+			matchingRules = append(matchingRules, rule)
+		}
+	}
+	return matchingRules
+}
+
+// handleRouting handles routing for bap and bpp routing type
+func handleRouting(rule routingRule, uri, endpoint string, routingType string) (*definition.Route, error) {
+	if uri == "" {
```
ashishkgGoogle, line 190: a simpler implementation would be as follows :
target:= strings.TrimSpace(rule.Target.URL)
if len(strings.TrimSpace(uri)) != 0 {
target =strings.TrimSpace( uri)
}
if len(target) == 0{
  return not found error
}
return &definition.Route{
				RoutingType: routingType,
				TargetURL:   target,
			}, nil

ashishkgGoogle (changes requested):

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+
+	"gopkg.in/yaml.v3"
+)
+
+// Config holds the configuration for the Router plugin.
+type Config struct {
+	RoutingConfig string `json:"routingConfig"`
+}
+
+// RoutingConfig represents the structure of the routing configuration file.
```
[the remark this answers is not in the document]
tanyamadaan: The config defines the plugin config, which has a path defined for routingConfig.
This routingConfig is a yaml file which has routingRules defined.

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+type routingConfig struct {
+	RoutingRules []routingRule `yaml:"routingRules"`
```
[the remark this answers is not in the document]
tanyamadaan: A sample routingConfig looks like below:

routingRules:
  - domain: "ONDC:TRV10"
    version: "2.0.0"
    routingType: "url"
    target:
      url: "https://services-backend/trv/v1"
    endpoints:
      - on_select
.......


routingRules: immediately communicates that this section defines routing rules
More explicit than just rules: which could mean anything

pkg/plugin/definition/router.go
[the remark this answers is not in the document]
tanyamadaan: This PR is not yet merged. Can take this up in integration task. We will have to anyways do it for all the plugins.

pkg/plugin/definition/router.go
tanyamadaan: This PR is not yet merged. Can take this up in integration task. We will have to anyways do it for all the plugins.

pkg/plugin/definition/router.go
tanyamadaan: TopicID is Broker-Neutral: Works for Kafka (topics), RabbitMQ (Exchanges + Routing Keys), PubSub (topics),  Matches AWS SNS/SQS terminology

pkg/plugin/implementation/router/cmd/plugin.go
tanyamadaan: Done

pkg/plugin/definition/router.go
tanyamadaan: Changing it to TargetType for clarity.

pkg/plugin/definition/router.go
tanyamadaan: Changing it to publisherId, due to the nomenclature being used by publisher plugin.

pkg/plugin/implementation/router/router.go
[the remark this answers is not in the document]
tanyamadaan: No change required

pkg/plugin/implementation/router/router.go
tanyamadaan: Will make changes, once model PR is merged.

pkg/plugin/implementation/router/router.go
tanyamadaan: Modified.

pkg/plugin/implementation/router/router.go
tanyamadaan: Will take up as part of integration task.

pkg/plugin/implementation/router/router.go
tanyamadaan: It should be a not found error. Will integrate with model.go as part of integration test.

pkg/plugin/implementation/router/router.go
tanyamadaan: This will be done as part of schema validation.

pkg/plugin/implementation/router/router.go
tanyamadaan: It is actually for any other routing type, which does not match the four enums defined.
Its added as part of validation. Removed from here.

pkg/plugin/implementation/router/router.go
tanyamadaan: This is required to return in case both bpp_uri and target.url does not exist. Anyways the implementation has got some changes here.

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// New initializes a new Router instance with the provided configuration.
+// It loads and validates the routing rules from the specified YAML file.
+// Returns an error if the configuration is invalid or the rules cannot be loaded.
+func New(ctx context.Context, config *Config) (*Router, func() error, error) {
+	// Check if config is nil
+	if config == nil {
+		return nil, nil, fmt.Errorf("config cannot be nil")
+	}
```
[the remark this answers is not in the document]
tanyamadaan: This is a single validation. Rest validations are already part of validation method.

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,232 @@
...
+					URL:        rule.Target.URL,
+				}
+			case targetTypeBPP, targetTypeBAP:
+				route = &definition.Route{
+					TargetType: rule.TargetType,
+					URL:        rule.Target.URL, // Fallback URL if URI not provided in request
+				}
+			}
+
+			fmt.Print(r.rules)
```
ashishkgGoogle, line 125: please remove print lines

pkg/plugin/implementation/router/cmd/plugin_test.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	if err := os.WriteFile(tempPath, content, 0644); err != nil {
+		t.Fatalf("Failed to create temp config: %v", err)
+	}
+
+	return tempPath
+}
+
+// TestRouterProviderSuccess tests the RouterProvider implementation for success cases.
+func TestRouterProviderSuccess(t *testing.T) {
```
ashishkgGoogle, line 36: since there is just once case here, we can have a simple test here insted of a table driven one.

ashishkgGoogle (approved):
Left a few minor comments resolve them if you can

pkg/plugin/implementation/router/testData/bpp_receiver.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+  - domain: "ONDC:TRV11"
+    version: "2.0.0"
+    targetType: "url"
+    target:
+      url: "https://services-backend/trv/v1"
+    endpoints:
+      - select
+      - init
+      - confirm
```
ashishkgGoogle, line 30: remove empty lines from the yaml, also we don't need to quote string in yaml file

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,232 @@
...
+		if len(target) == 0 {
+			return nil, fmt.Errorf("could not determine destination for endpoint '%s': neither request contained a BAP URI nor was a default URL configured in routing rules", endpoint)
+		}
+		route = &definition.Route{
+			TargetType: route.TargetType,
+			URL:        target,
+		}
+	}
+
+	return route, nil
```
ashishkgGoogle, line 231: before returning url we should append the endpoint to it

pkg/plugin/implementation/router/router.go
```diff
@@ -0,0 +1,232 @@
...
+// Target contains destination-specific details.
+type target struct {
+	URL         string `yaml:"url,omitempty"`         // URL for "url" or gateway endpoint for "bpp"/"bap"
+	PublisherID string `yaml:"publisherId,omitempty"` // For "msgq" type
+}
+
+// TargetType defines possible target destinations.
+const (
+	targetTypeURL  = "url"  // Route to a specific URL
+	targetTypeMSGQ = "msgq" // Route to a message queue
```
ashishkgGoogle, line 50: can we change this to 
targetTypePublisher = "publisher"

This will make it consistent with the rest of the application as we are defining the plugin as a publisher as well.

ashishkgGoogle (changes requested):

ashishkgGoogle (approved):

PR #437, opened 2025-03-24
feat: test file for response and error module

MayurWitsLab:


pkg/model/error.go
ashishkgGoogle, line 1: add a comment for each exported type.

pkg/model/model.go
```diff
@@ -0,0 +1,107 @@
...
+const (
+	AuthHeaderSubscriber          string = "Authorization"
+	AuthHeaderGateway             string = "X-Gateway-Authorization"
+	UnaAuthorizedHeaderSubscriber string = "WWW-Authenticate"
+	UnaAuthorizedHeaderGateway    string = "Proxy-Authenticate"
+)
+
+type contextKey string
+
+var MsgIDKey = contextKey("message_id")
```
ashishkgGoogle, line 39: this should be a const string.

pkg/response/response.go
```diff
@@ -13,131 +17,93 @@ const (
...
+				Status: model.StatusNACK,
 			},
+			Error: err,
 		},
 	}
-
-	return json.Marshal(response)
+	data, jsonErr := json.Marshal(resp)
+	if jsonErr != nil {
+		http.Error(w, "failed to marshal response", http.StatusInternalServerError)
```
ashishkgGoogle, line 67: log the error here, then in response error message should be : fmt.Sprintf("Internal server error, MessageID: %s", ctx.Value(model.MsgIDKey), same for the below error as well

pkg/response/response.go
```diff
@@ -13,131 +17,93 @@ const (
 	InvalidRequestErrorType   ErrorType = "INVALID_REQUEST"
 )
```
ashishkgGoogle, line 19: remove this

pkg/response/response.go
```diff
@@ -13,131 +17,93 @@ const (
...
-		Status string `json:"status,omitempty"`
-	} `json:"ack,omitempty"`
-	Error *Error `json:"error,omitempty"`
-}
-
-type BecknResponse struct {
-	Context map[string]interface{} `json:"context,omitempty"`
-	Message Message                `json:"message,omitempty"`
-}
+type errorResponseWriter struct{}
```
ashishkgGoogle, line 20: why do we need this ?

pkg/response/response_test.go
ashishkgGoogle, line 1: Something wrong here

pkg/model/error_test.go
```diff
@@ -0,0 +1,153 @@
...
+		Code:    "404",
+		Paths:   "/api/v1/user",
+		Message: "User not found",
+	}
+
+	expected := "Error: Code=404, Path=/api/v1/user, Message=User not found"
+	actual := err.Error()
+
+	if actual != expected {
+		t.Errorf("expected %s, got %s", expected, actual)
```
ashishkgGoogle, line 22: We should follow this pattern for success tests :

   t.Errorf("%s\err.Error() = %s, want %s",
                   actual, expected)}

For error tests it should be if gotErr != test.wantErr {
    t.Errorf("func(%q) = %v, want error presence = %v", test.input, err, test.wantErr)
}

ashishkgGoogle (changes requested):

pkg/response/response.go
MayurWitsLab: we are using it in test cases to test error

pkg/model/coverage.out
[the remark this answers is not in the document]
MayurWitsLab: done

pkg/model/error.go
MayurWitsLab: done

pkg/model/model.go
MayurWitsLab: done

pkg/response/response.go
MayurWitsLab: have removed the err check part as discussed

pkg/response/response.go
MayurWitsLab: done

pkg/model/error_test.go
MayurWitsLab: have updated this in the same test file

pkg/response/response.go
ashishkgGoogle: then this should be declared in the test file.

pkg/response/response_test.go
ashishkgGoogle: I see the code this was pointing to has been updated, Now I can't remember what it was about.

pkg/response/response.go
MayurWitsLab: done

pkg/model/error_test.go
MayurWitsLab: updated the pattern

pkg/model/error_test.go
```diff
@@ -0,0 +1,168 @@
...
+			notFoundErr.Error(), err.Error())
+	}
+}
+
+func TestRole_UnmarshalYAML_ValidRole(t *testing.T) {
+	var role Role
+	yamlData := []byte("bap")
+
+	err := yaml.Unmarshal(yamlData, &role)
+	assert.NoError(t, err)
```
ashishkgGoogle, line 157: We do not want to use asserts for this, please add a todo here to fix this later

pkg/response/response_test.go
```diff
@@ -1,303 +1,248 @@
...
+
+			var w http.ResponseWriter
+			if tt.useBadWrite {
+				w = &errorResponseWriter{} // Simulate write error
+			} else {
+				w = httptest.NewRecorder()
+			}
+
+			// Force marshal error if needed
+			if tt.useBadJSON {
```
ashishkgGoogle, line 217: this is not the right way for handling different error cases, add a todo here to fix this later

ashishkgGoogle (approved):

pkg/response/response_test.go
MayurWitsLab: done

pkg/model/error_test.go
MayurWitsLab: > We do not want to use asserts for this, please add a todo here to fix this later

done

Deepa-Mulchandani (approved):

ashishkgGoogle (approved):

PR #448, opened 2025-04-03
Bug Fix for ReqPreProcessor plugin

tanyamadaan:


pkg/plugin/implementation/reqpreprocessor/reqpreprocessor.go
```diff
@@ -10,103 +10,70 @@ import (
...
-		return true
-	default:
-		return false
-	}
-}
-
 func validateConfig(cfg *Config) error {
 	if cfg == nil {
 		return errors.New("config cannot be nil")
 	}
```
ashishkgGoogle, line 74: this is fine for now, but we should add a todo to revisit this once we implement gateway and registery

pkg/plugin/implementation/reqpreprocessor/reqpreprocessor_test.go
```diff
@@ -87,71 +98,113 @@ func TestNewUUIDSetterSuccessCases(t *testing.T) {
...
 	}
 }
 
-func TestNewUUIDSetterErrorCases(t *testing.T) {
+func TestNewPreProcessorErrorCases(t *testing.T) {
 	tests := []struct {
 		name         string
 		config       *Config
-		requestBody  map[string]any
+		requestBody  interface{}
```
ashishkgGoogle, line 124: We should change this to any

pkg/plugin/implementation/reqpreprocessor/reqpreprocessor_test.go
ashishkgGoogle, line 1: These tests do not seem correct to me. Please add a todo here to separate the tests for middleware creation and execution, I will take it up separately.

ashishkgGoogle (approved):

pkg/plugin/implementation/reqpreprocessor/reqpreprocessor.go
tanyamadaan: Gateway and Registry does not need to update fields. Why would we need this.

pkg/plugin/implementation/reqpreprocessor/reqpreprocessor_test.go
tanyamadaan: Correct, however, just added tests to enhance coverage after bug fix. Did not modify the test execution part. Added Todo.

PR #457, opened 2025-04-21
Fix/reg lookup

MayurWitsLab:


pkg/plugin/implementation/registery/cmd/plugin_test.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+			if cleanup != nil {
+				t.Fatal("Expected cleanup function to be nil but got a non-nil function")
+			}
+		})
+	}
+}
+
+func TestParseConfig(t *testing.T) {
+	tests := []struct {
```
ashishkgGoogle, line 73: please split the test for success and error

pkg/plugin/implementation/registery/cmd/plugin_test.go
```diff
@@ -0,0 +1,143 @@
...
+					t.Fatal("Expected error but got none")
+				}
+				return
+			}
+
+			if err != nil {
+				t.Fatalf("Unexpected error: %v", err)
+			}
+
+			if cfg.RegistryURL != tt.expectedCfg.RegistryURL || cfg.RetryMax != tt.expectedCfg.RetryMax {
```
ashishkgGoogle, line 138: instead of this please use cmp.diff

pkg/plugin/implementation/registery/registery.go
```diff
@@ -0,0 +1,99 @@
...
+	if config.RetryWaitMin < 0 {
+		return errors.New("RetryWaitMin must be non-negative")
+	}
+	if config.RetryWaitMax < 0 {
+		return errors.New("RetryWaitMax must be non-negative")
+	}
+	if config.RetryWaitMin > config.RetryWaitMax {
+		return errors.New("RetryWaitMin cannot be greater than RetryWaitMax")
+	}
+	if config.RetryMax < 0 {
```
ashishkgGoogle, line 45: in the main config we are just read url and retryMax, can you please add parsing logic for all of these as well. thanks

pkg/plugin/implementation/registery/registery.go
```diff
@@ -0,0 +1,99 @@
...
+	r := &registryLookup{
+		client: client,
+		config: config,
+	}
+	return r, nil, nil
+}
+
+// Lookup performs a POST request to the /lookUp endpoint of the registry service with retry logic.
+func (r *registryLookup) Lookup(ctx context.Context, subscription *model.Subscription) ([]model.Subscription, error) {
+	lookupURL := fmt.Sprintf("%s/lookUp", r.config.LookupURL)
```
ashishkgGoogle, line 71: this is not needed as the config will have the entire lookup url

pkg/plugin/implementation/registery/registery_test.go
```diff
@@ -0,0 +1,189 @@
...
+	"time"
+
+	"github.com/beckn/beckn-onix/pkg/model"
+	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+func TestRegistryLookupSuccess(t *testing.T) {
+	tests := []struct {
+		name           string
```
ashishkgGoogle, line 18: we don't need to add a table driven test when there is just one case.

pkg/plugin/implementation/registery/registery_test.go
```diff
@@ -0,0 +1,189 @@
...
+		})
+	}
+}
+
+
+func TestValidateConfig(t *testing.T) {
+	tests := []struct {
+		name        string
+		config      *Config
+		expectedErr string
```
ashishkgGoogle, line 141: please split the error and success cases in different tests

pkg/plugin/manager.go
```diff
@@ -387,3 +387,23 @@ func unzip(src, dest string) error {
 
 	return nil
 }
+
+// RegistryLookup retrieves the registryLookup plugin instance.
```
ashishkgGoogle, line 391: add tests for this

pkg/plugin/implementation/registery/registery.go
```diff
@@ -0,0 +1,99 @@
+package registery
+
+import (
+	"bytes"
```
Deepa-Mulchandani, line 4: can we fix the name registery everywhere it should be registry

MohitKatare-protean: Resolved

PR #466, opened 2025-04-29
Publisher plugin implementation

MohitKatare-protean:
Publisher plugin Implementation

### DIFF
--- a/pkg/plugin/implementation/publisher/cmd/plugin.go
+++ b/pkg/plugin/implementation/publisher/cmd/plugin.go
@@ -2,22 +2,18 @@ package main
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/beckn/beckn-onix/pkg/log"
 	"github.com/beckn/beckn-onix/pkg/plugin/definition"
 	"github.com/beckn/beckn-onix/pkg/plugin/implementation/publisher"
-	"github.com/rabbitmq/amqp091-go"
 )
 
 // publisherProvider implements the PublisherProvider interface.
 // It is responsible for creating a new Publisher instance.
 type publisherProvider struct{}
 
-// New creates a new Publisher instance based on the provided configuration map.
-// It also returns a cleanup function to close resources and any potential errors encountered.
+// New creates a new Publisher instance based on the provided configuration.
 func (p *publisherProvider) New(ctx context.Context, config map[string]string) (definition.Publisher, func() error, error) {
-	// Step 1: Map config
 	cfg := &publisher.Config{
 		Addr:       config["addr"],
 		Exchange:   config["exchange"],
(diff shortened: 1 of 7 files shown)


pkg/plugin/implementation/publisher/cmd/plugin.go
tanyamadaan, line 1: Here , you have added the logic for creating connection to publisher directly in plugin.go, but the intended design was to have this logic in publisher.go with plugin.go just a thin wrapper.

Move the logic to new function in publisher.go and plugin.go should handle config mapping and calling publisher.New()

tanyamadaan, line 1: Here , you have added the logic for creating connection to publisher directly in plugin.go, but the intended design was to have this logic in publisher.go with plugin.go just a thin wrapper.

Move the logic to new function in publisher.go and plugin.go should handle config mapping and calling publisher.New()

pkg/plugin/implementation/publisher/publisher_test.go
```diff
@@ -0,0 +1,220 @@
...
+			username: "",
+			password: "",
+			config:   &Config{Addr: "localhost:5672"},
+			wantErr:  true,
+		},
+		{
+			name:     "Missing config address",
+			username: "guest",
+			password: "guest",
+			config:   &Config{}, // this won't error unless Validate() is called separately
```
tanyamadaan, line 69: validate is called within New function right? why this -> // this won't error unless Validate() is called separately

pkg/plugin/implementation/publisher/publisher_test.go
```diff
@@ -0,0 +1,220 @@
...
+		},
+	}
+
+	// Set valid credentials
+	os.Setenv("RABBITMQ_USERNAME", "guest")
+	os.Setenv("RABBITMQ_PASSWORD", "guest")
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			url, err := GetConnURL(tt.config)
```
tanyamadaan, line 39: Call New function, why GetConnURL is called here?

pkg/plugin/implementation/publisher/publisher_test.go
```diff
@@ -0,0 +1,220 @@
...
+				os.Setenv("RABBITMQ_USERNAME", tt.username)
+			}
+
+			if tt.password == "" {
+				os.Unsetenv("RABBITMQ_PASSWORD")
+			} else {
+				os.Setenv("RABBITMQ_PASSWORD", tt.password)
+			}
+
+			url, err := GetConnURL(tt.config)
```
tanyamadaan, line 88: Same as comment in positive test case.

pkg/plugin/implementation/publisher/publisher_test.go
```diff
@@ -0,0 +1,220 @@
...
+			}
+
+			if err == nil && url == "" {
+				t.Errorf("expected non-empty URL, got empty string")
+			}
+		})
+	}
+}
+
+func TestValidateSuccess(t *testing.T) {
```
tanyamadaan, line 100: Same as above comment

pkg/plugin/implementation/publisher/publisher_test.go
```diff
@@ -0,0 +1,220 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := Validate(tt.config); err != nil {
+				t.Errorf("expected no error, got: %v", err)
+			}
+		})
+	}
+}
+
+func TestValidateFailure(t *testing.T) {
```
tanyamadaan, line 123: Same as comment in GetConnURL.

tanyamadaan (changes requested):

pkg/plugin/implementation/publisher/publisher_test.go
MohitKatare-protean: yes, but in this test we are only calling GetConnURL which does not invoke validate.

pkg/plugin/implementation/publisher/publisher_test.go
MohitKatare-protean: this test was only intended to verify the connection URL generation logic via GetConnURL.

pkg/plugin/implementation/publisher/publisher_test.go
```diff
@@ -0,0 +1,350 @@
...
+			name:   "Missing Exchange",
+			config: &Config{Addr: "localhost:5672"},
+		},
+		{
+			name:   "Empty Addr and Exchange",
+			config: &Config{Addr: " ", Exchange: " "},
+		},
+	}
+
+	for _, tt := range tests {
```
tanyamadaan, line 147: Validate expected errors as well.

pkg/plugin/implementation/publisher/publisher_test.go
```diff
@@ -0,0 +1,350 @@
...
+			name: "Valid config with vhost",
+			config: &Config{
+				Addr:   "localhost:5672/myvhost",
+				UseTLS: false,
+			},
+		},
+	}
+
+	// Set valid credentials
+	os.Setenv("RABBITMQ_USERNAME", "guest")
```
tanyamadaan, line 35: Use t.Setenv(...)

tanyamadaan: Same for other similar cases.

tanyamadaan (changes requested):

pkg/plugin/implementation/publisher/publisher.go
```diff
@@ -0,0 +1,196 @@
...
+
+// GetConnURL constructs the RabbitMQ connection URL using the config and environment credentials.
+func GetConnURL(cfg *Config) (string, error) {
+	user := os.Getenv("RABBITMQ_USERNAME")
+	pass := os.Getenv("RABBITMQ_PASSWORD")
+	if user == "" || pass == "" {
+		return "", model.NewBadReqErr(fmt.Errorf("missing RabbitMQ credentials in environment"))
+	}
+
+	parts := strings.SplitN(cfg.Addr, "/", 2)
```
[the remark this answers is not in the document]
MohitKatare-protean: I have changed this line with strings.SplitN(strings.TrimSpace(cfg.Addr), "/", 2) also added the test case for it.

pkg/plugin/implementation/publisher/publisher.go
```diff
@@ -0,0 +1,196 @@
...
+	encodedUser := url.QueryEscape(user)
+	encodedPass := url.QueryEscape(pass)
+	encodedVHost := url.QueryEscape(vhost)
+	protocol := "amqp"
+	if cfg.UseTLS {
+		protocol = "amqps"
+	}
+
+	connURL := fmt.Sprintf("%s://%s:%s@%s/%s", protocol, encodedUser, encodedPass, hostPort, encodedVHost)
+	log.Debugf(context.Background(), "Generated RabbitMQ connection URL: %s", connURL)
```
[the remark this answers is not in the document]
MohitKatare-protean: changed line to log useful context detail for debugging without exposing username and password.

pkg/plugin/implementation/publisher/publisher.go
```diff
@@ -0,0 +1,196 @@
...
+type Config struct {
+	Addr       string
+	Exchange   string
+	RoutingKey string
+	Durable    bool
+	UseTLS     bool
+}
+
+// Channel defines the interface for publishing messages to RabbitMQ.
+type Channel interface {
```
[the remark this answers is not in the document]
MohitKatare-protean: Exported the Channel interface its used for both publisher and plugin packages. it was necessary for mocking the AMQP channel during the test cases.

tanyamadaan (approved):

PR #470, opened 2025-04-30
fix: logs level

MayurWitsLab:


pkg/plugin/manager.go
```diff
@@ -86,13 +86,14 @@ func plugins(ctx context.Context, cfg *ManagerConfig) (map[string]onixPlugin, er
...
 	}
+	log.Infof(ctx, "Successfully loaded %d plugins", len(plugins))
 
 	return plugins, nil
 }
 
 // loadPlugin attempts to load a plugin from the given path and logs the execution time.
 func loadPlugin(ctx context.Context, path, id string) (onixPlugin, time.Duration, error) {
-	log.Debugf(ctx, "Loading plugin: %s", id)
+	log.Infof(ctx, "Loading plugin: %s", id)
```
tanyamadaan, line 96: For Debug, add plugins/ modules initialized with their configs

core/module/module.go
```diff
@@ -29,7 +29,7 @@ var handlerProviders = map[handler.Type]Provider{
 // It iterates over the module configurations, retrieves appropriate handler providers,
 // and registers the handlers with the HTTP multiplexer.
 func Register(ctx context.Context, mCfgs []Config, mux *http.ServeMux, mgr handler.PluginManager) error {
-	log.Debugf(ctx, "Registering modules with config: %#v", mCfgs)
+	log.Infof(ctx, "Registering modules with config: %#v", mCfgs)
```
tanyamadaan, line 32: As defined in logging strategy doc, please add 
1. Info level: names of modules/plugins initialised
2. Debug Level: Modules and plugins initialised with their configs. (for debugging)

core/module/handler/stdHandler.go
```diff
@@ -193,29 +193,42 @@ func loadKeyManager(ctx context.Context, mgr PluginManager, cache definition.Cac
 // initPlugins initializes required plugins for the processor.
 func (h *stdHandler) initPlugins(ctx context.Context, mgr PluginManager, cfg *PluginCfg, regURL string) error {
 	var err error
+	log.Infof(ctx, "Initializing Cache plugin: %s", cfg.Cache)
```
tanyamadaan, line 196: Aren't these duplicate of whats added in plugin/manager.go - Line No. 96.
Same for all logs added here.

core/module/handler/stdHandler.go
```diff
@@ -193,29 +193,42 @@ func loadKeyManager(ctx context.Context, mgr PluginManager, cache definition.Cac
...
 	if h.publisher, err = loadPlugin(ctx, "Publisher", cfg.Publisher, mgr.Publisher); err != nil {
 		return err
 	}
+
+	log.Infof(ctx, "Initializing Signer plugin: %s", cfg.Signer)
 	if h.signer, err = loadPlugin(ctx, "Signer", cfg.Signer, mgr.Signer); err != nil {
 		return err
 	}
 
-	log.Debugf(ctx, "All required plugins successfully loaded for stdHandler")
```
tanyamadaan, line 218: Lets keep this as is.

core/module/client/registery.go
```diff
@@ -75,7 +76,7 @@ func (c *registryClient) Lookup(ctx context.Context, subscription *model.Subscri
 		return nil, fmt.Errorf("failed to create request: %w", err)
 	}
 	req.Header.Set("Content-Type", "application/json")
-
+	log.Infof(ctx, "Request forwarded to %s", lookupURL)
```
tanyamadaan, line 79: Not clear at all. For info, it should be clear on what task performed. For ex. log.Infof(ctx, "Lookup request sent to %s", lookupURL)

core/module/client/registery.go
```diff
@@ -48,7 +49,7 @@ func (c *registryClient) Subscribe(ctx context.Context, subscription *model.Subs
 		return fmt.Errorf("failed to create request: %w", err)
 	}
 	req.Header.Set("Content-Type", "application/json")
-
+	log.Infof(ctx, "Request forwarded to %s", subscribeURL)
```
tanyamadaan, line 52: Same as lookup comment below

tanyamadaan (changes requested):

core/module/handler/stdHandler.go
NidhaSV: Yes, these look like a repetition. A minor difference I see is line no. 96 logs the plugin ID(`mgr.Cache`), but here it is the name and cfg

NidhaSV: Wait, this `loadPlugin` call  doesn't really go to line 96 of plugin/manager.go. This actually goes to the generic function `loadPlugin[T any]` at line 158 of this same file.

PR #479, opened 2025-06-03
Initial commit for Lookup call in subscription Handler

AshwiniK-protean:


core/module/client/registery.go
```diff
@@ -99,3 +100,31 @@ func (c *registryClient) Lookup(ctx context.Context, subscription *model.Subscri
 
 	return results, nil
 }
+
+// RegistryClientInterface defines the contract for making subscription requests to a Beckn registry.
+type RegistryClientInterface interface {
```
ashishkgGoogle, line 105: why do we need this interface here? In golang we define interfaces in consumer packages

core/module/client/registery.go
```diff
@@ -99,3 +100,31 @@ func (c *registryClient) Lookup(ctx context.Context, subscription *model.Subscri
...
+	CreateRequest(ctx context.Context, method, endpoint string, body []byte) (*http.Response, error)
+}
+
+var _ RegistryClientInterface = (*registryClient)(nil)
+
+// CreateRequest creates an HTTP request with retry capability
+func (c *registryClient) CreateRequest(ctx context.Context, method, endpoint string, body []byte) (*http.Response, error) {
```
ashishkgGoogle, line 112: This method's name should be createSubscription

ashishkgGoogle: The input to this method should be *model.Subscription request, 

The method should alway be POST for this so we don't need it as a paramter and the endpoint should be created by appending "subscribed" the the baseUrl in the struct similar to lookup.

Also add a separate method for UpdateSubscription, where everything else remains the same just the method should be changes to patch.

core/module/client/registery.go
```diff
@@ -99,3 +100,31 @@ func (c *registryClient) Lookup(ctx context.Context, subscription *model.Subscri
...
+	if c.config == nil || c.client == nil {
+		return nil, errors.New("client or config is not initialized")
```
ashishkgGoogle, line 114: this check should be part of the newRegistryClient method and should not be required here.

core/module/client/registery.go
```diff
@@ -99,3 +100,31 @@ func (c *registryClient) Lookup(ctx context.Context, subscription *model.Subscri
...
+	}
+	url := fmt.Sprintf("%s/%s", c.config.RegisteryURL, endpoint)
```
ashishkgGoogle, line 116: this should be 
	url := fmt.Sprintf("%s/%s", c.config.RegisteryURL,"subscribe")

core/module/handler/subscription_status_check.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/beckn/beckn-onix/core/module/client"
+	"github.com/beckn/beckn-onix/pkg/log"
+	"github.com/beckn/beckn-onix/pkg/model"
+)
+
+// SubscriptionStatusChecker manages asynchronous status lookups with retries.
+type SubscriptionStatusChecker struct {
+	Client          *client.Config
+	MaxRetries      int
+	RetryDelay      time.Duration
```
ashishkgGoogle, line 20: Why are all these fields exported?

core/module/handler/subscription_status_check.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// fetchSubscriptionStatus performs the actual HTTP request for lookup.
+func (s *SubscriptionStatusChecker) fetchSubscriptionStatus(ctx context.Context, lookupReq model.LookupRequest) (string, error) {
+	reqBody, err := json.Marshal(lookupReq)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal lookup request: %w", err)
+	}
+
+	resp, err := s.registeryClient.CreateRequest(ctx, http.MethodPost, "lookup", reqBody)
```
ashishkgGoogle, line 78: Why are we not using the lookup method available in the registryClient ?

core/module/handler/subscription_status_check.go
```diff
@@ -0,0 +1,99 @@
...
+	return &SubscriptionStatusChecker{
+		Client:     clientConfig,
+		MaxRetries: retries,
+		RetryDelay: delay,
+	}
+}
+
+// LookupStatus attempts multiple registry lookups until it finds a "SUBSCRIBED" response.
+func (s *SubscriptionStatusChecker) LookupStatus(ctx context.Context, lookupReq model.LookupRequest) (bool, error) {
+	for attempt := 1; attempt <= s.MaxRetries; attempt++ {
```
ashishkgGoogle, line 58: Since we are using retrriaybleHttp Client in this is not needed

core/module/handler/subscription_status_check.go
```diff
@@ -0,0 +1,99 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("unexpected response code: %d", resp.StatusCode)
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("failed to read response body: %w", err)
+	}
+
+	var lookupResp model.LookupResponse
```
ashishkgGoogle, line 93: We do not need a sepearte struct for lookup response, the type for lookup response is []model.Subscription

pkg/model/model.go
```diff
@@ -184,3 +184,17 @@ type Message struct {
 type Response struct {
 	Message Message `json:"message"`
 }
+
+// LookupRequest is the payload sent to registry's /lookup
+type LookupRequest struct {
```
ashishkgGoogle, line 189: We should use the model.Subscription Object for this instead of using a new Request Type.

ashishkgGoogle (changes requested):

PR #520, opened 2025-09-19
feat(router): add excludeAction option for URL routing

nirmalnr:
- Add ExcludeAction field to target struct in routing configuration
- When excludeAction is true, skip appending action to URL path
- Provides flexibility for routing scenarios where action should not be part of URL path
- Maintains backward compatibility by defaulting to existing behaviour
- Allows users to match the exact behaviour of the legacy protocol server which did not append action to the URL when forwarding.

This change allows routing rules to be configured with excludeAction: true when the target URL should not have the action appended to its path.

### DIFF
--- a/pkg/plugin/implementation/router/router.go
+++ b/pkg/plugin/implementation/router/router.go
@@ -42,6 +42,7 @@ type routingRule struct {
 type target struct {
 	URL         string `yaml:"url,omitempty"`         // URL for "url" or gateway endpoint for "bpp"/"bap"
 	PublisherID string `yaml:"publisherId,omitempty"` // For "msgq" type
+	ExcludeAction bool `yaml:"excludeAction,omitempty"` // For "url" type to exclude appending action to URL path
 }
 
 // TargetType defines possible target destinations.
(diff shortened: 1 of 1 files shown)


pkg/plugin/implementation/router/router.go
```diff
@@ -115,7 +116,9 @@ func (r *Router) loadRules(configPath string) error {
 				if err != nil {
 					return fmt.Errorf("invalid URL in rule: %w", err)
 				}
-				parsedURL.Path = joinPath(parsedURL, endpoint)
+				if !rule.Target.ExcludeAction {
```
ashishkgGoogle, line 119: Please add unit tests for this. Also instead of making this a double negative check, making it a positive check will make it more readable.
instead of "if !rule.Target.ExcludeAction"  use "if rule.Target.AppendAction"

nirmalnr: I initially made it as AppendAction, but thought will use ExcludeAction as if it's excluded in the yaml file, it will be false by default. But agree on the readability. Will update and add test cases.

ashishkgGoogle: What you are suggesting makes sense, this is fine, Backwards compatibility is more important.

nirmalnr: Added unit tests for the following:
- For target type `URL`:
    - `ExcludeAction` explicitly set `true`
    - `ExcludeAction` explicitly set to `false`
    - `ExcludeAction` omitted
- `ExcludeAction` when target type is `BPP`
- `ExcludeAction` when target type is `BAP`
- `ExcludeAction` when target type is `Publisher`

PR #545, opened 2025-10-29
Issue 540 - fix: remove hardcoded value and add test cases

ameersohel45:
Changes baseURL to url and adds required registryName parameter for configurable registry lookup paths. Removes status mapping from is_revoked field. also added test cases for the same. 

Issue : https://github.com/Beckn-One/beckn-onix/issues/540


nirmay (approved):
approved.

PR #555, opened 2025-11-17
Issue 552 - Improve schema2validator Performance by Building In-Memory Action→Schema Lookup Map

ameersohel45:
This update optimizes the `schema2validator` plugin by eliminating the repeated traversal of the OpenAPI spec during every request.

Previously, even though the spec was cached with a TTL, each validation still required extracting `context.action` and scanning the entire spec to find the corresponding schema. This added unnecessary overhead.

## What Changed
- Added an **in-memory `action → schema` map** that is built during the spec load phase.  
- Cached the map alongside the loaded OpenAPI spec for reuse.  
- Updated the validation flow to perform **constant-time schema lookups** using this map.  

## Outcome
These changes significantly reduce validation latency, simplify lookup logic, and improve overall runtime performance by avoiding repeated full-path scans.

Issue : https://github.com/Beckn-One/beckn-onix/issues/552

### DIFF
--- a/pkg/plugin/implementation/schemav2validator/schemav2validator.go
+++ b/pkg/plugin/implementation/schemav2validator/schemav2validator.go
@@ -31,8 +31,9 @@ type schemav2Validator struct {
 
 // cachedSpec holds a cached OpenAPI spec.
 type cachedSpec struct {
-	doc      *openapi3.T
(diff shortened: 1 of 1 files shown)


nirmay (approved):
approved.

PR #572, opened 2025-12-02
fix the test case in main_test.go and module_test.go

manendrapalsingh:
Fix the import and add the mock for the test file
`func (m *MockPluginManager) Registry(ctx context.Context, cfg *plugin.Config) (definition.RegistryLookup, error) {
	return nil, nil
}`

`func (m *mockPluginManager) Registry(ctx context.Context, cfg *plugin.Config) (definition.RegistryLookup, error) {
	return nil, nil
}`


nirmay (approved):
approved.

Issue #594, opened 2026-01-16
Enable NPs control over transaction eligibility based on NFO affiliation


nirmalnr:
### Problem Statement

Network Participants (NPs) should have the freedom to decide who they transact with based on Network Facilitating Organization (NFO) affiliation. Specifically, an NP should be able to choose whether to transact with all NPs listed in beckn registries or restrict transactions only to NPs that are affiliated with one or more explicitly selected NFOs.

When an NP is removed from an NFO registry, other NPs should be able to detect this change and decide whether to continue transacting with that NP based on their configured trust preferences. If the NP has configured they only want to transact with NPs belonging to a specific NFO, any requests from NPs not part of the NFO should be rejected.

### Current State

- NFOs publish registries of affiliated NPs.
- NPs generally transact with other NPs without applying configurable policies based on NFO affiliation.
- There is no standardized mechanism for an NP to express or enforce preferences around trusted NFOs.

### Proposed Change

Introduce a mechanism that allows NPs to:
* Configure transaction policies that specify whether they:
  * transact with all NPs in the registry, or
  * transact only with NPs affiliated with one or more selected NFOs.
* Evaluate NFO affiliation dynamically using registry data when determining transaction eligibility.
* Automatically reflect registry updates (additions, removals, or modifications) in transaction decisions without manual intervention.

### Benefits

* Gives NPs autonomy over their trust and governance model.
* Supports diverse network participation policies without enforcing a single global rule.
* Ensures transaction decisions remain aligned with up-to-date registry information.
* Improves network safety and governance transparency.

### Acceptance Criteria

* NPs can configure whether to transact with all NPs or restrict transactions based on selected NFOs.
* Transaction eligibility checks evaluate current NFO affiliation using registry data.
* Registry changes are reflected in transaction decisions automatically.
* Default behavior preserves backward compatibility for NPs that do not configure restrictions.

nirmalnr:
### Proposed Solution

Currently the below lookup is done to fetch the keys for signature validation:

> **GET** `<registry_url>/dedi/lookup/{subscriber_id}/subscribers.beckn.one/{record_id}`

The response will be of the below format:
```json
{
	"message": "ok",
	"data": {
		"namespace": "example-np",
		"namespace_id": "76EU8M5Zkwgu8g...",
		"registry_id": "76EU8ATmvmimFmR1...",
		"registry_name": "subscription-details",
[...]

PR #595, opened 2026-01-20
feat: add allowed parent namespaces for signature validation

nirmalnr:
Added an optional `allowedParentNamespaces` allowlist to the DeDi registry plugin and enforce it during signature validation by checking the registry lookup response’s parent_namespaces. The allowlist is supplied as a comma‑separated string in plugin config and parsed at plugin init. Lookup now rejects entries that are not part of any allowed parent namespace, with matching tests and documentation updates.

The configuration for registry plugin should have an optional configuration allowedParentNamespaces which will be a csv of verified domains that the NP want to exclusively transact with. Example:

```yaml
      plugins:
        registry:
          id: dediregistry
          config:
            url: https://api.testnet.beckn.one/registry/dedi
            registryName: subscribers.beckn.one
            timeout: 10
            allowedParentNamespaces: commerce-network.org, retail-collective.org
```

Compared to the initial proposal, the configuration was changed from a list to a CSV. Plugin configurations across the system are parsed into map[string]string, and YAML lists do not unmarshal cleanly into string values without modifying this shared configuration type. Using CSV avoids a broader refactor and remains consistent with existing plugin configuration handling.

This fixes issue #594

### DIFF
--- a/pkg/plugin/implementation/dediregistry/dediregistry.go
+++ b/pkg/plugin/implementation/dediregistry/dediregistry.go
@@ -152,13 +152,6 @@ func (c *DeDiRegistryClient) Lookup(ctx context.Context, req *model.Subscription
 		return nil, fmt.Errorf("invalid response format: missing details field")
 	}
 
-	parentNamespaces := extractStringSlice(data["parent_namespaces"])
-	if len(c.config.AllowedParentNamespaces) > 0 {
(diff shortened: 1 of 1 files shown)


nirmalnr:
Updated the error message as `registry entry with subscriber_id '<subscriber_id>' does not belong to any configured parent namespaces (registry.config.allowedParentNamespaces)`

PR #606, opened 2026-01-30
Update registry URLs in example configuration files to use beckn.io domain

nirmalnr:
Updated registry service URLs in example configuration files from
`https://api.testnet.beckn.one/registry` and `https://api.beckn.one/registry` 
to `https://api.dev.beckn.io/registry` and `https://api.beckn.io/registry` respectively


nirmay (approved):
approved.

PR #615, opened 2026-02-26
Implement Policy Enforcer Plugin

711ayush711:
- Added a new Policy Enforcer plugin to evaluate incoming messages against OPA policies.
- Configurable via YAML with options for policy sources, actions, and query.
- Integrated into existing configuration files for BAP and BPP.
- Updated related tests and documentation for the new functionality.
- Enhanced plugin manager to support Policy Enforcer instantiation.
- Note on config shape (no YAML arrays for URLs): The plugin config in beckn-onix is typed as map[string]string, so each value is a single string, not a structured list. For now we intentionally keep the simpler comma‑separated policyUrls pattern
- Added documentation for using YAML folded scalar (>-) to keep long comma-separated policyUrls values readable across multiple lines.

### DIFF
--- a/pkg/plugin/implementation/opapolicychecker/enforcer_test.go
+++ b/pkg/plugin/implementation/opapolicychecker/enforcer_test.go
@@ -855,6 +855,32 @@ result := {
 	}
 }
 
+func TestEvaluator_NonStructuredMapResult_Ignored(t *testing.T) {
+	policy := `
+package policy
+
+import rego.v1
+
+result := {
+  "action": "confirm",
+  "status": "ok"
+}
+`
+	dir := writePolicyDir(t, "policy.rego", policy)
+	eval, err := NewEvaluator([]string{dir}, "data.policy.result", nil, false, 0)
+	if err != nil {
+		t.Fatalf("NewEvaluator failed: %v", err)
+	}
+
+	violations, err := eval.Evaluate(context.Background(), []byte(`{}`))
+	if err != nil {
+		t.Fatalf("Evaluate failed: %v", err)
+	}
+	if len(violations) != 0 {
+		t.Fatalf("expected non-structured map result to be ignored, got %v", violations)
+	}
+}
+
 // --- Bundle Tests ---
 
 // buildTestBundle creates an OPA bundle .tar.gz in memory from the given modules.
--- a/pkg/plugin/implementation/opapolicychecker/evaluator.go
+++ b/pkg/plugin/implementation/opapolicychecker/evaluator.go
@@ -332,14 +332,8 @@ func extractViolations(rs rego.ResultSet) ([]string, error) {
 					}
 				}
 			case map[string]interface{}:
-				// Check for structured result: {"valid": bool, "violations": [...]}
 				if vs := extractStructuredViolations(v); vs != nil {
 					violations = append(violations, vs...)
-				} else {
-					// Fallback: OPA sometimes returns sets as maps with string keys
-					for key := range v {
-						violations = append(violations, key)
-					}
 				}
 			}
 		}


nirmalnr:
Currently, the plugin expects policy query output in a violations-oriented shape. It extracts violations from list/set-like results (`[]interface{}` with string items, or map keys from `map[string]interface{}`). If the query returns `true`/`false` (a common allow/deny policy pattern), no violations are extracted, so the request is treated as compliant and passes. This means a policy that returns false to indicate denial will not be enforced as a denial in the current implementation. So I think the policy query output structure should be standardized explicitly.

config/onix/adapter.yaml
```diff
@@ -98,6 +113,10 @@ modules:
           id: schemavalidator
           config:
             schemaDir: /mnt/gcs/configs/schemas
+        policyEnforcer:
+          id: policyenforcer
+          config:
+            policyPaths: "./policies"
```
nirmay: I would really like the example config to show how to load policy from a url, and not take a short cut from the file.
This is critical because policy files are maintained by network orchestrators and they really a simple way to get those updated files to NPs, so, let us host a simple file somewhere in beckn github and show that to be used here.

pkg/plugin/definition/policyEnforcer.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+
+	"github.com/beckn-one/beckn-onix/pkg/model"
+)
+
+// PolicyEnforcer interface for policy enforcement on incoming messages.
+type PolicyEnforcer interface {
+	Run(ctx *model.StepContext) error
```
nirmay: It might be useful to revise this to more relevant verb, this will allow for any extension needed to include new interface functions in the future.

pkg/plugin/implementation/policyenforcer/README.md
```diff
@@ -0,0 +1,113 @@
+# Policy Enforcer Plugin
+
+OPA/Rego-based policy enforcement for beckn-onix adapters. Evaluates incoming beckn messages against configurable policies and NACKs non-compliant requests.
```
nirmay: policy enforcement is not for onix adapter. We need to introduce this properly. what is being enforced? who is created those rules??
@711ayush711 @ameetdesh

pkg/plugin/implementation/policyenforcer/README.md
```diff
@@ -0,0 +1,105 @@
...
+## Configuration
+
+All config keys are passed via `map[string]string` in the adapter YAML config.
+
+| Key | Required | Default | Description |
+|-----|----------|---------|-------------|
+| `policyUrls` | At least one of `policyUrls`, `policyDir`, or `policyFile` required | — | Comma-separated list of URLs, local file paths, or directory paths to `.rego` files |
+| `policyPaths` | | `./policies` | Local directory or path containing `.rego` files |
+| `policyFile` | | — | Single local `.rego` file path |
+| `query` | No | `data.policy.violations` | Rego query returning violation strings |
```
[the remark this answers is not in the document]
nirmay: One critical thing we are missing here is that the policy applies to a single network. In a network of networks world, each app (and onix sitting next to it) will be working within multiple networks. now, the policy is specific to a network, so config has to map the opa/rego files to a network only. 
I would imagine that for inbound/outbound messages, this is indicated by networdId and not domain, but this needs to be clarified with larger team. @ameetdesh @711ayush711

policies/compliance.rego
```diff
@@ -0,0 +1,18 @@
+package policy
```
nirmay: this is completely wrong place for this file. It should strictly be inside the implementation folder. when adapter instance chooses not to use the plugin, there should be absolutely no trace of it anywhere. Given that the rego appraoch is for the plugin and not anything outside of it, there should be no rego files outside on plugin/implementation/opapolicyenforcer folder.

by the way, the example file is not even a hello world, does not even have any rules.
@ameetdesh @711ayush711

nirmay:
Some more comments, @ameetdesh @711ayush711 

---

## Code Review — PR #615: Implement Policy Enforcer Plugin

### 🔴 Critical Issues

**1. Silent pass-through when `policyEnforcer` is `nil` at runtime**

In `step.go`, `newEnforcePolicyStep` correctly guards against a nil `policyEnforcer` and returns an error. But in `stdHandler.go`, the plugin is loaded with `loadPlugin` using `cfg.PolicyEnforcer` — which is `omitempty`. If the config block is absent, the plugin is never loaded and `h.policyEnforcer` stays `nil`. When the step `"policyEnforcer"` is referenced in the steps list, `newEnforcePolicyStep(nil)` will return an error at startup — which is the right behavior. But the concern is whether all config files that omit the plugin but still reference the step name will fail loudly at startup vs silently at request time. This needs a clear test case covering that scenario.

**2. No HTTP timeout on remote policy fetches**

`policyPaths` supports remote URLs. There's no evidence of a configurable (or even hardcoded) HTTP timeout for fetching `.rego` files from URLs. In production, a slow or unreachable policy server would cause the plugin initialization to hang, potentially blocking adapter startup entirely. A timeout with a sensible default (e.g. 10s) is required, ideally configurable.

**3. No policy reload / cache invalidation mechanism**

Policies are loaded at startup. In a production network, policies authored by network orchestrators need to be updatable without restarting the adapter. There's no TTL, file-watcher, or refresh interval mechanism. At minimum, this constraint should be clearly documented as a known limitation. Ideally, a `refreshIntervalSeconds` config option should be introduced.

**4. `go.mod` version bump is unrelated and potentially destabilizing**

The PR bumps `go 1.24.0` → `go 1.24.6` and upgrades several indirect dependencies (prometheus, otel, etc.) as a side-effect of adding the OPA dependency. These should be separated into a distinct PR or at least explicitly called out. Mixing dependency upgrades with feature work makes rollback harder.

---

### 🟡 Design & Correctness Issues

**5. Violation extraction logic doesn't handle `bool` results**

Based on the diff's comment in `enforcer.go` — `model.NewBadReqErr` is only called when violations are extracted from a list or map. If an OPA policy returns a plain `false` (a common deny pattern), the plugin treats the request as compliant and passes it through. This is a silent policy bypass bug. The query output contract must be explicitly defined and enforced — the plugin should either require a `violations` shape, or explicitly handle the `bool` case and treat `false` as a denial.

**6. `policyPaths` is a comma-separated string parsed from `map[string]string`**

The config type constraint (`map[string]string`) forces a comma-separated encoding for multiple paths. This is a workaround for a structural limitation and is fragile — paths containing commas (unusual but valid on some systems) would silently break. The YAML folded-scalar trick documented in comments helps readability, but doesn't fix the underlying parsing risk. This deserves a comment in the parsing code explicitly validating that no individual parsed path contains unexpected characters.

**7. `newEnforcePolicyStep` is a trivial function that adds no value**

```go
func newEnforcePolicyStep(policyEnforcer definition.PolicyEnforcer) (definition.Step, error) {
    if policyEnforcer == nil {
        return nil, fmt.Errorf("invalid config: PolicyEnforcer plugin not configured")
    }
    return policyEnforcer, nil
}
```

The `PolicyEnforcer` interface already extends `definition.Step` (it implements `Run`). This wrapper just does a nil-check and returns the interface as-is. While the nil check is useful, naming this `newEnforcePolicyStep` gives it an implied constructor role that is misleading. Either rename it to `validatePolicyEnforcer` or inline the nil check at the callsite.

**8. `policies/compliance.rego` at repo root is incorrect placement**

A `.rego` example file has been placed at `policies/compliance.rego` at the repository root. This creates a conceptual coupling between the repo structure and a plugin that is entirely opt-in. Any operator who doesn't use this plugin will still see this `policies/` directory alongside `config/`, `schemas/`, etc., making it appear to be a required or first-class directory. The example belongs inside `pkg/plugin/implementation/policyenforcer/testdata/` or similar.

---

### 🟡 Testing Gaps

**9. Mock `PolicyEnforcer` always returns `nil, nil`**

Both mock implementations in `main_test.go` and `module_test.go` return `(nil, nil)` for `PolicyEnforcer`. This means zero test coverage for the `policyEnforcer` step path in integration tests. At minimum, one test scenario should inject a non-nil mock that simulates a violation and verify the step chain is correctly short-circuited.

**10. Benchmark tests don't assert correctness**

The benchmark test (`benchmark_test.go`) measures evaluation and compilation time under varying rule counts, which is good. However, benchmarks that only measure performance without also asserting that the evaluation result is correct can mask behavioral regressions. At least a basic correctness assertion should be included alongside the timing measurement.

**11. No test for the URL fetching path**

The `policyPaths` URL fetching is a critical path (especially given it's the recommended usage). There's no test — even with an `httptest.Server` — verifying that a policy loaded from a remote URL is correctly compiled and evaluated. This should be a required test given the security implications.

---

### 🟢 Low Priority / Style

**12. Inconsistent casing: `policyEnforcer` vs `policyenforcer`**

The plugin `id` is `policyenforcer` (all lowercase), while the config block key and step name are `policyEnforcer` (camelCase). While this follows existing conventions in the repo (e.g., `schemavalidator` vs `schemaValidator`), it's worth a comment in the README explicitly explaining this distinction to avoid confusion for operators writing their own configs.

**13. Indentation-only change in `main_test.go`**

Lines 191–194 in `main_test.go` show a diff that is purely indentation. This adds noise to the PR diff and should be reverted.

---

**Summary:** The core structure is reasonable and the OPA integration is architecturally sound. The two most important things to fix before merge are the silent pass-through bug for boolean policy results (#5), and the lack of a fetch timeout for remote URLs (#2). The rest are important but addressable in follow-ups if documented.

Issue #617, opened 2026-02-27
fix: ineffective nil guard on log provider in EmitAuditLogs


nirmay:
## Problem

In `pkg/telemetry/audit.go`, the nil check on the logger provider never triggers:

```go
provider := global.GetLoggerProvider()
if provider == nil {   // this never triggers
    logger.Warnf(ctx, "failed to emit audit logs, logs disabled")
    return
}
```

`global.GetLoggerProvider()` always returns a no-op provider (never `nil`) when OtelSetup is not configured. As a result, when logs are disabled, `EmitAuditLogs` silently emits to the no-op provider with no warning logged.

## Suggested fix

Either:

**Option A — remove the guard entirely**, accepting that the no-op provider silently discards logs (correct behaviour, no misleading warning needed).

**Option B — use a package-level flag** set by `otelsetup` when a real log provider is registered, and gate `EmitAuditLogs` on that flag.

## Context

Identified during review of PR #605 (Network Observability / OTLP migration).

nirmay:
merged.

Issue #616, opened 2026-02-27
fix: meter provider singleton race in HTTPMetrics and StepMetrics


nirmay:
## Problem

`httpMetricsOnce` (`core/module/handler/http_metric.go:36`) and `stepMetricsOnce` (`core/module/handler/step_metrics.go:33`) permanently capture `otel.GetMeterProvider()` on first invocation via `sync.Once`.

If any HTTP request reaches the handler before the `otelsetup` plugin has registered the global meter provider, both singletons bind permanently to the no-op provider. All metrics are silently dropped for the lifetime of the process — no errors, no warnings.

## Steps to reproduce

1. Start the adapter with `otelsetup` enabled.
2. If a request arrives (or a test runs) before `otelsetup.New()` completes and calls `otel.SetMeterProvider(...)`, the no-op provider is captured.
3. All subsequent `onix_http_request_count` and step execution metrics are discarded.

## Suggested fix

Either:

**Option A — drop `sync.Once`, resolve lazily per call**
```go
func GetHTTPMetrics(ctx context.Context) (*HTTPMetrics, error) {
    return newHTTPMetrics() // otel.GetMeterProvider() is safe to call repeatedly
}
```
The OTel SDK meter provider and its instruments are safe to create multiple times from the same provider; the SDK deduplicates by instrument name.

**Option B — document and enforce init order**
Add an explicit comment and/or startup assertion in `main.go` guaranteeing that the `otelsetup` plugin is initialised before the HTTP server starts accepting requests.

Option A is simpler and removes the fragility entirely.

## Context

Identified during review of PR #605 (Network Observability / OTLP migration).

nirmay:
merged.

PR #656, opened 2026-04-14
fix: remove sync.Once race and fix nil guard on log provider

nirmay:
### Issue #616 — Metric instrument caching: from sync.Once to provider-identity cache

**Root cause:** The original `sync.Once` singletons for all five metric getters fired exactly once. If a request or constructor ran before `otelsetup` registered the real `MeterProvider`, the no-op instruments were cached permanently with no escape hatch.

**First fix attempt (removed):** Drop `sync.Once` and call `newXxx()` on every `Get*Metrics` invocation, relying on the OTel SDK's name-based deduplication. This eliminated the permanent trap but introduced a new problem: per-request and per-operation allocation on the hot paths (`ServeHTTP` calls `GetHTTPMetrics` twice per request; every cache op calls `GetCacheMetrics`).

**Final fix — provider-identity cache:** Each of the five metric getters now holds a package-level cache struct `{ sync.RWMutex, MeterProvider, *Metrics }`. On every call:

1. **Fast path (RLock):** if `otel.GetMeterProvider()` returns the same pointer as the cached provider, return cached instruments immediately. Cost: one atomic load + pointer compare — negligible at any RPS.
2. **Slow path (write lock, double-checked):** rebuild `newXxx()` only when `otel.SetMeterProvider` has been called since the last build, then update the cached pointer and instruments.

**Why this is not sync.Once:** `sync.Once` fires exactly once — stale no-op instruments are cached forever. The provider-identity check invalidates the cache whenever the provider pointer changes. A call racing with `otel.SetMeterProvider` binds to whichever provider it observes; the very next call after the real provider is registered rebinds correctly. There is no permanent trap.

All five getters follow the identical pattern for consistency — including construction-time getters (`GetStepMetrics`, `GetHandlerMetrics`, `GetMetrics`) — so a constructor that runs before the real provider is set will self-heal on the next call rather than leaving the owning struct permanently bound to no-op instruments.

### Issue #617 — Ineffective nil guard on log provider

`global.GetLoggerProvider()` never returns nil — it always returns a no-op provider when no real provider is registered, making the previous `if provider == nil` check dead code.

**Fix:** Introduced a package-level `atomic.Bool` (`telemetry.SetLogsEnabled`) that `otelsetup` sets to `true` only after successfully wiring the real log provider via `global.SetLoggerProvider`. `EmitAuditLogs` gates on `LogsEnabled()` instead of the nil check, and `otelsetup`'s shutdown closure resets it to `false` after tearing down the log provider — preventing emission through a closed provider if telemetry is restarted with logs disabled.

### Test improvements

- `pkg/telemetry/test_helper.go`: Added `RecordingLogExporter` (in-memory log recorder) and `NewTestProviderWithLogs` helper that wires a real SDK log provider, sets `SetLogsEnabled(true)`, and resets the flag on `Shutdown`.
- `pkg/telemetry/audit_test.go` (new): Tests for `LogsEnabled` toggle, `EmitAuditLogs` early-return when disabled, record emission and attribute presence when enabled, and flag reset on provider shutdown.
- `core/module/handler/step_metrics_test.go`: Fixed `TestGetStepMetrics_ConcurrentAccess` — removed stale same-pointer assertion left over from the `sync.Once` era; replaced with per-instrument non-nil checks and a 20-goroutine concurrent stress test.

## Test plan

- [ ] `go test -race ./core/module/handler/... ./pkg/telemetry/...` — all pass, no races detected
- [ ] `go build ./pkg/plugin/implementation/otelsetup` — clean build
- [ ] `go.mod` remains at `go 1.24.6` — no version bump

### DIFF
--- a/core/module/handler/http_metric.go
+++ b/core/module/handler/http_metric.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"sync"
 
 	"github.com/beckn-one/beckn-onix/pkg/telemetry"
 	"go.opentelemetry.io/otel"
@@ -33,11 +34,42 @@ func newHTTPMetrics() (*HTTPMetrics, error) {
 	return m, nil
 }
 
-// GetHTTPMetrics returns a fresh HTTPMetrics bound to the current global meter
-// provider. otel.GetMeterProvider() is safe to call repeatedly; the SDK
-// deduplicates instruments by name, so there is no double-registration risk.
-func GetHTTPMetrics(ctx context.Context) (*HTTPMetrics, error) {
-	return newHTTPMetrics()
+// httpMetricsCache caches the HTTPMetrics for the current global MeterProvider.
+// When otel.SetMeterProvider is called (e.g. during startup or test setup) the
+// provider pointer changes, the cache misses, and instruments are rebound once.
+// Fast-path cost on every request: one atomic load + RLock + two pointer compares.
+var httpMetricsCache struct {
+	mu       sync.RWMutex
+	provider metric.MeterProvider
+	m        *HTTPMetrics
+}
+
+// GetHTTPMetrics returns HTTPMetrics bound to the current global MeterProvider,
+// rebuilding only when the provider has been replaced since the last call.
+func GetHTTPMetrics(_ context.Context) (*HTTPMetrics, error) {
+	current := otel.GetMeterProvider()
+
+	httpMetricsCache.mu.RLock()
+	if httpMetricsCache.provider == current && httpMetricsCache.m != nil {
+		m := httpMetricsCache.m
+		httpMetricsCache.mu.RUnlock()
+		return m, nil
+	}
+	httpMetricsCache.mu.RUnlock()
+
+	httpMetricsCache.mu.Lock()
+	defer httpMetricsCache.mu.Unlock()
+	// Double-check after acquiring the write lock.
+	if httpMetricsCache.provider == current && httpMetricsCache.m != nil {
+		return httpMetricsCache.m, nil
+	}
+	m, err := newHTTPMetrics()
+	if err != nil {
+		return nil, err
+	}
+	httpMetricsCache.provider = current
+	httpMetricsCache.m = m
+	return m, nil
 }
 
 // StatusClass returns the HTTP status class string (e.g. 200 -> "2xx").
--- a/core/module/handler/http_metric.go
+++ b/core/module/handler/http_metric.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
-	"sync"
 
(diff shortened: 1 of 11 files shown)


pkg/plugin/implementation/otelsetup/otelsetup.go
```diff
@@ -173,6 +173,7 @@ func (Setup) New(ctx context.Context, cfg *Config) (*telemetry.Provider, error)
 		processor := logsdk.NewBatchProcessor(logExporter)
 		logProvider = logsdk.NewLoggerProvider(logsdk.WithProcessor(processor), logsdk.WithResource(resAudit))
 		global.SetLoggerProvider(logProvider)
+		telemetry.SetLogsEnabled(true)
```
nirmalnr, line 176: Reset the log-enabled flag when tearing telemetry down

`SetLogsEnabled(true)` is now the only guard that lets `EmitAuditLogs` talk to the global logger provider, but this shutdown path never flips it back to `false`. After `provider.Shutdown(...)`, or after reinitializing telemetry with logs disabled, `EmitAuditLogs` will still treat logging as enabled and emit through a stale/shut-down global provider. The new test helper already resets the flag on shutdown; the production setup needs the same lifecycle handling.

core/module/handler/http_metric.go
```diff
@@ -41,11 +33,11 @@ func newHTTPMetrics() (*HTTPMetrics, error) {
...
 
+// GetHTTPMetrics returns a fresh HTTPMetrics bound to the current global meter
+// provider. otel.GetMeterProvider() is safe to call repeatedly; the SDK
+// deduplicates instruments by name, so there is no double-registration risk.
 func GetHTTPMetrics(ctx context.Context) (*HTTPMetrics, error) {
-	httpMetricsOnce.Do(func() {
-		httpMetricsInstance, httpMetricsErr = newHTTPMetrics()
-	})
-	return httpMetricsInstance, httpMetricsErr
+	return newHTTPMetrics()
```
nirmalnr: HTTP metric instrument creation moved onto the request hot path

`GetHTTPMetrics` now rebuilds the instrument wrapper on every call. That looks harmless in isolation, but `stdHandler.ServeHTTP` calls it for every request and `RecordHTTPRequest` calls it again before recording, so this PR turns one startup-time race fix into repeated per-request instrument lookup/allocation. The SDK deduplicates registrations, but it does not make these calls free. I’d avoid doing `newHTTPMetrics()` on every request and instead rebind once when the real provider is installed, or cache by provider identity.

nirmalnr (changes requested):
Do let me know if you want me to pick these up.

nirmay:
Thanks @nirmalnr for catching that, I have now taken a different approach to fix the original problem, and it is now summarised in the original PR summary.
please review.

nirmalnr (approved):

PR #661, opened 2026-04-15
OPA policy checker: add network policy config and signature verification

### LINKED ISSUE #642: OPA policy checker plugin does not support verification of signed policy artifacts
## Summary

The [`opapolicychecker`](https://github.com/beckn/beckn-onix/tree/main/pkg/plugin/implementation/opapolicychecker) plugin currently supports `type: url | file | dir | bundle` with a single `location` field, but it does not provide any config structure for signature verification. The docs also explicitly call out that bundle signature verification is not implemented yet.

We need to extend the plugin config so policy artifacts can optionally be verified before they are loaded:

- for `file`, `url`, and `dir`, the user should be able to point to the detached signature location
- for `bundle`, the signature should be read from inside the bundle itself
- when verification is enabled, the user should also provide a public key lookup URL so the plugin can fetch the public key used for verification
- the user should also be able to disable verification explicitly

## Current config

Today the plugin config is flat:

```yaml
checkPolicy:
  id: opapolicychecker
  config:
    type: file
    location: ./pkg/plugin/implementation/opapolicychecker/testdata/example.rego
    query: "data.policy.result"
```

This is not sufficient for signature verification because:

- `location` only identifies the policy source
- there is no place to specify detached signature location for `file`, `url`, or `dir`
- there is no place to specify the public key lookup URL
- there is no explicit way to turn verification on or off

## Proposed config change

Keep the existing `location` field as-is and add an optional `verification` block.

Suggested shape:

```yaml
checkPolicy:
  id: opapolicychecker
  config:
    type: file
    location: ./policies/network.rego
    verification:
      enabled: true
      publicKeyLookupUrl: https://fabric.nfh.global/lookup/example-nfo.com/key-registry/policy-key
      signatureLocation: ./policies/network.rego.sig
    query: "data.policy.result"
```

For a remote file:

```yaml
checkPolicy:
  id: opapolicychecker
  config:
    type: url
    location: https://nfo.example.org/policies/network.rego
    verification:
      enabled: true
      publicKeyLookupUrl: https://fabric.nfh.global/lookup/example-nfo.com/key-registry/policy-key
      signatureLocation: https://nfo.example.org/policies/network.rego.sig
    query: "data.policy.result"
```

For a directory:

```yaml
checkPolicy:
  id: opapolicychecker
  config:
    type: dir
    location: ./policies/
    verification:
      enabled: true
      publicKeyLookupUrl: https://fabric.nfh.global/lookup/example-nfo.com/key-registry/policy-key
      signatureLocation: ./policies/policies.sig
    query: "data.policy.result"
```

For a bundle:

```yaml
checkPolicy:
  id: opapolicychecker
  config:
    type: bundle
    location: https://nfo.example.org/policies/bundle.tar.gz
    verification:
      enabled: true
      publicKeyLookupUrl: https://fabric.nfh.global/lookup/example-nfo.com/key-registry/policy-key
    query: "data.retail.validation.result"
```

For no verification:

```yaml
checkPolicy:
  id: opapolicychecker
  config:
    type: file
    location: ./policies/network.rego
    verification:
      enabled: false
    query: "data.policy.result"
```

## Validation rules

- `location` remains mandatory for all source types
- `verification` is optional
- if `verification.enabled` is `false`, no signature verification is performed
- if `verification.enabled` is `true`:
- for `type: file | url | dir`, both `publicKeyLookupUrl` and `signatureLocation` are required
- for `type: bundle`, `publicKeyLookupUrl` is required
- for `type: bundle`, `signatureLocation` must not be set, because the signature is expected inside the bundle

### When verification is disabled

- load the policy artifact directly from `location`
- do not attempt signature verification

### When verification is enabled for `file`, `url`, or `dir`

- load the policy artifact from `location`
- load the detached signature from `verification.signatureLocation`
- fetch the public key from `verification.publicKeyLookupUrl`
- verify the artifact before compiling or evaluating policy
- fail closed if signature verification fails or the public key cannot be resolved

### When verification is enabled for `bundle`

- load the bundle from `location`
- read signature metadata from inside the bundle
- fetch the public key from `verification.publicKeyLookupUrl`
- verify the bundle before loading it into OPA
- fail closed if embedded signature verification fails or the public key cannot be resolved

## Why this structure

- keeps the current `location` field unchanged
- avoids adding redundant fields that can conflict with `type`
- makes verification optional
- keeps bundle verification distinct without introducing extra config states
- minimizes migration cost for existing users

## Backward compatibility

This change is additive if `verification` is optional. Existing configs should continue to work unchanged, with the default behavior remaining equivalent to no verification unless `verification.enabled: true` is set.

## Acceptance criteria

- config supports optional signature verification for `file`, `url`, `dir`, and `bundle`
- `file`, `url`, and `dir` support detached signature location
- `bundle` supports embedded signature verification
- all verified source types support `publicKeyLookupUrl`
- `bundle` rejects `signatureLocation`
- policy load fails closed when verification is enabled and verification fails
- plugin docs are updated with the new config structure and examples

nirmalnr:
## Summary

This PR extends `opapolicychecker` to support network-aware policy selection, signed policy artifacts, and a cleaner configuration model.

## Changes

- add network-policy-based configuration using `networkPolicyConfig`
- load all configured policies at startup and select the correct one at request time using `context.networkId` / `context.network_id`
- policy enforcement is opt-in by `network_id`; unmatched networks are skipped unless `default` is defined
- support `default` as the fallback policy entry
- an exact network match with `enabled: false` intentionally overrides `default`
- remove the old flat single-policy mode from `opapolicychecker`
- simplify policy source types to:
  - `file`
  - `dir`
  - `bundle`
- make `file` support both local file paths and remote URLs
- make `bundle` support both local bundle paths and remote bundle URLs
- document `dir` as test/dev-only and not recommended for production use

## Signature Verification

- add optional signature verification per policy entry
- support detached signature verification for `type: file`
- support OPA native signed bundle verification for `type: bundle`
- reject signature verification for `type: dir`
- move verification config into a nested `verification` block
- make bundle verification algorithm configurable via `verification.algorithm`
- default bundle verification algorithm to `ES256`

## DeDi Public Key Lookup

- change public key resolution to support DeDi lookup responses
- extract public key details from `data.details`
- support DeDi `keyFormat: base64`
- continue to support raw PEM / certificate parsing where applicable

## Config Cleanup

- replace `refreshIntervalSeconds` with `refreshInterval`
- parse refresh interval using Go duration strings such as:
  - `30s`
  - `5m`
  - `24h`

## Logging

- include `networkID` in policy violation and evaluation logs
- include request context details in policy logs:
  - `bap_id`
  - `bpp_id`
  - `message_id`
  - `transaction_id`
  - `action`
  - `timestamp`
- log loaded network policies at startup

## Docs and Examples

- update `opapolicychecker` README
- update example adapter configs to use `networkPolicyConfig`
- add example network policy config files for local and onix example setups

## Validation

Tested with focused local/unit coverage for:
- network policy config parsing
- duration-based refresh interval parsing
- nested verification block parsing
- DeDi-style public key response parsing
- local file signature verification
- local bundle loading
- signed local bundle verification

Manually tested:
- Single rego file signature verification
- Bundle file signature verification
- Policy selection based on network ID
- Error when wrong public key provided

fixes #643 Support network-specific policy configuration
fixes #642 OPA policy checker plugin does not support verification of signed policy artifacts


nirmay:
Here's my review of **beckn/beckn-onix#661** — "OPA policy checker: add network policy config and signature verification" against latest main (`54ae28e`).

---

## Summary

This PR replaces the flat single-policy mode in `opapolicychecker` with a network-keyed policy model, adds optional signature verification for file and bundle policies, and upgrades the refresh interval config to use Go duration strings. The architecture is solid and the test coverage is good. A few issues are worth addressing.

---

## Issues

### 1. Missing test: `enabled: false` on a network-specific entry overriding `default`

The PR description explicitly documents this behavior:
> An exact network match with `enabled: false` intentionally overrides `default`

And `selectedPolicy` ([enforcer.go:1032](pkg/plugin/implementation/opapolicychecker/enforcer.go)) implements it — it returns the exact match even if disabled, and `CheckPolicy` then skips evaluation. But there is **no test** covering this case. The existing `TestEnforcer_NetworkPolicyDisabled` only tests the global `enabled: false` on the top-level config. Add a test like:

```go
// configPath with:
//   retail.network/production: enabled: false
//   default: enabled: true, type: dir, ...
// Then request with networkId=retail.network/production should PASS
// (not fall through to default, not error)
```

This is a subtle, intentional behavior that's easy to regress.

---

### 2. Variadic `verification` in `NewEvaluator` creates a silent nil ambiguity

```go
// evaluator.go
func NewEvaluator(..., verification ...*ArtifactVerificationConfig) (*Evaluator, error)
```

`loadPolicy` calls it passing `policyConfig.Verification` directly:
```go
NewEvaluator(..., policyConfig.Verification)
```

When `Verification` is nil, this passes `(*ArtifactVerificationConfig)(nil)` as a variadic element, so `len(verification) == 1` and `verifyCfg = nil`. The nil checks inside `newRegoEvaluator`/`parseBundleArchive` handle this safely, but it's non-obvious. Consider using an explicit `*ArtifactVerificationConfig` parameter instead of variadic, or at minimum add a comment.

---

### 3. `parsePublicKeyValue` with `base64` strips whitespace using `strings.Fields` — edge case

```go
clean := strings.Join(strings.Fields(value), "")
```

`strings.Fields` splits on all Unicode whitespace. If the DeDi API ever returns a URL-safe base64 (`+` → `-`, `/` → `_`) or unpadded base64, `base64.StdEncoding.DecodeString` will silently fail. The README says only `keyFormat: base64` (standard) is supported, but a comment in code would prevent future surprises. Not blocking.

---

### 4. `resolveVerificationPublicKey` always attempts JSON parse for local PEM files

```go
func parseVerificationPublicKeyResponse(data []byte) (any, error) {
    if err := json.Unmarshal(data, &response); err == nil && response.Data.Details.PublicKey != "" {
        return parsePublicKeyValue(...)
    }
    return parseVerificationPublicKey(data)  // fallback: PEM/DER
}
```

For local PEM files (used in all current tests), JSON unmarshal fails, and parsing falls through to PEM. This works but the dual-use of `resolveVerificationPublicKey` for both DeDi HTTP responses and local PEM files is implicit. Worth a brief comment distinguishing the two paths — or separate them if local-file-only mode grows in usage.

---

### 5. No test for DeDi HTTP public key lookup (online path)

All signature verification tests use local file paths for `PublicKeyLookupURL`. The DeDi HTTP path is exercised only in `TestParseVerificationPublicKeyResponse_DeDiBase64Key`, which tests the JSON parsing in isolation. There's no integration test with a mock HTTP server serving a DeDi-style response and feeding it through `NewEvaluator`. Recommend adding one (similar to how `TestEnforcer_RemoteFilePolicy` uses `httptest.NewServer`).

---

### 6. `config/onix/opa-network-policies.yaml` references test data in a production config

```yaml
# config/onix/opa-network-policies.yaml
networkPolicies:
  default:
    type: file
    location: /app/pkg/plugin/implementation/opapolicychecker/testdata/example.rego
```

This ships a production example config pointing at `testdata/`. The `example.rego` is probably permissive (allow-all) by design for local testing, but it's a footgun for operators who deploy `onix/adapter.yaml` without customizing the policy. At minimum add a comment warning that this must be replaced before production use.

---

## Minor / Nits

- **Breaking change clarity**: The PR removes `type`, `location`, `query`, `actions`, `refreshIntervalSeconds` from top-level config entirely. Existing deployments will fail at startup with `'networkPolicyConfig' is required`. This should be called out in the PR description as a migration requirement (the description doesn't use the word "breaking").

- **`policyEntryKnownKeys` validation doesn't reject deep nesting**: If a user puts an unexpected nested map under a non-`verification` key (e.g., `actions: {a: b}`), `loadNetworkPolicies` returns `"must be a scalar value"`. This is correct behavior but is not tested.

- **`logLoadedPolicy` iterates `e.policies` without lock at startup** — this is fine because it's called from `New()` before the enforcer is shared, but worth noting it's not under `evaluatorMu`.

- **Typo in `isBecknDirection`** (pre-existing): `"reciever"` — pre-existing, not introduced by this PR.

---

## What's well done

- The `selectedPolicy` concurrency model is correct: RLock for read, Lock for atomic swap on reload, pointer-based return so old `loadedPolicy` objects outlive the swap safely.
- Fail-closed startup: if any network policy entry is invalid or fails to load, startup fails (not silent degradation).
- Ed25519 signing in `verifyDetachedSignature` correctly passes the full `content` to `ed25519.Verify` rather than a pre-hashed digest.
- The `networkPolicies` YAML flattening of `verification.*` sub-keys is clean and avoids a custom YAML type.
- Hot-reload atomicity is preserved in the new multi-policy model.

711ayush711:
Reviewed. All issues already captured in the existing review above

nirmalnr:
### 1. Missing test: `enabled: false` on a network-specific entry overriding `default`

Addressed.

- Added `TestEnforcer_NetworkPolicyDisabledExactMatchOverridesDefault` in `pkg/plugin/implementation/opapolicychecker/enforcer_test.go`
- This verifies that an exact `network_id` match with `enabled: false` skips evaluation and does not fall through to `default`

### 2. Variadic `verification` in `NewEvaluator` creates a silent nil ambiguity

Addressed.

- Changed `NewEvaluator` in `pkg/plugin/implementation/opapolicychecker/evaluator.go` to take an explicit `*ArtifactVerificationConfig` instead of a variadic parameter
- Updated all call sites and tests accordingly
- This removes the nil/len ambiguity entirely

### 3. `parsePublicKeyValue` with `base64` strips whitespace using `strings.Fields` — edge case

Addressed with clarification, no behavior change.

- Added an inline code comment in `pkg/plugin/implementation/opapolicychecker/evaluator.go`
- The comment now explicitly states that `keyFormat=base64` currently expects standard padded base64 and does not support URL-safe or alternate encodings
- Also updated the README to state the same constraint

### 4. `resolveVerificationPublicKey` always attempts JSON parse for local PEM files

Addressed with clarification.

- Added a comment in `pkg/plugin/implementation/opapolicychecker/evaluator.go` explaining the fallback path:
  - DeDi-style JSON response parsing is attempted first
  - raw PEM/DER key material is accepted as a fallback for local files and non-DeDi endpoints

### 5. No test for DeDi HTTP public key lookup (online path)

Addressed.

- Added `TestEvaluator_FileVerificationWithDeDiHTTPPublicKeyLookup` in `pkg/plugin/implementation/opapolicychecker/enforcer_test.go`
- The test serves a DeDi-style JSON response over HTTP and verifies that the evaluator can use it for detached file signature verification

### 6. `config/onix/opa-network-policies.yaml` references test data in a production config

Addressed.

- Added a warning comment in `config/onix/opa-network-policies.yaml`
- It now explicitly says the example policy path must be replaced before production use

## Minor / Nits

### Breaking change clarity

Addressed.

- Added a migration note in `pkg/plugin/implementation/opapolicychecker/README.md`
- The README now explicitly states that `networkPolicyConfig` is required and that older top-level keys like `type`, `location`, `query`, `actions`, and `refreshIntervalSeconds` are no longer supported

### `policyEntryKnownKeys` validation doesn't reject deep nesting

Addressed.

- Added `TestLoadNetworkPolicies_RejectsUnexpectedNestedMap` in `pkg/plugin/implementation/opapolicychecker/enforcer_test.go`
- This verifies that unexpected nested maps under non-`verification` keys are rejected with the expected scalar-value validation error

### `logLoadedPolicy` iterates `e.policies` without lock at startup

No change.

- This is still safe in the current lifecycle because startup logging happens before the enforcer is shared
- Did not add locking here in this PR

### Typo in `isBecknDirection` (pre-existing): `"reciever"`

No change.

- This is pre-existing and unrelated to the policy checker changes in this PR
- Left it out to avoid expanding scope
- Analysing existing logic here. Will update as a separate issue/PR if to be changed.

nirmay:
@nirmalnr 
## Some suggestions:

1. selectedPolicy reads body twice per request — minor perf/allocation concern
CheckPolicy calls selectedPolicy(ctx.Body) (which calls extractNetworkID) and then formatRequestLogContext(ctx.Body) — both json.Unmarshal the same body. For high-throughput paths this doubles JSON parsing. A minor optimization: parse once and pass down. Not a blocker but worth noting.

enforcer.go

2. selectedPolicy can return a disabled policy that still has ev == nil — potential panic path avoided only by ordering
In CheckPolicy:

policy, selectedNetworkID := e.selectedPolicy(ctx.Body)

```
policyConfig := policy.config
ev := policy.evaluator

if !policyConfig.Enabled {
    return nil  // ← guard is here
}

if ev == nil {
    return model.NewBadReqErr(...)  // ← catches unenabled policies that slipped through?
```

The ev == nil check is only reached if Enabled is true but the evaluator is nil. This can happen if loadPolicy returned a non-nil loadedPolicy with evaluator == nil (which it does for enabled: false entries). The code path through CheckPolicy correctly returns early on !policyConfig.Enabled, so this is safe. However, the control flow is fragile — a future code change could break this ordering silently. Consider asserting ev != nil earlier or restructuring to make the invariant explicit.

3. policyEntryKnownKeys uses flattened dot-notation keys but YAML is parsed as a nested map — inconsistency in validation
policyEntryKnownKeys includes "verification.enabled", "verification.publicKeyLookupUrl", etc. These are the flattened keys that exist after loadNetworkPolicies converts the nested verification block. However, parsePolicyConfig also iterates cfg to collect RuntimeConfig for keys not in policyEntryKnownKeys. Since the flattened dot-keys are included in policyEntryKnownKeys, they're correctly excluded from RuntimeConfig. This is correct but subtle. A comment explaining that policyEntryKnownKeys contains post-flattening keys would help future maintainers.

4. config/opa-network-policies.yaml has no production-use comment — risk of misuse
config/opa-network-policies.yaml references testdata/example.rego as the default policy with no warning comment. The onix/ variant does have a comment (# Example only. Replace this...), but the local config doesn't. This is the file that will be used in local dev setups and could accidentally get deployed.

Suggestion: Add a comment to config/opa-network-policies.yaml matching the one in config/onix/opa-network-policies.yaml.

You may choose to file separate issues for some of these.

nirmalnr:
#### 1. selectedPolicy reads body twice per request — minor perf/allocation concern

Addressed.

- Refactored `pkg/plugin/implementation/opapolicychecker/enforcer.go` so `CheckPolicy` now parses request `context` once per request
- Added a shared `parsedRequestContext` helper path and reused it for:
  - network policy selection
  - body action fallback
  - request log context formatting
- `selectedPolicy` now takes the already-parsed `networkID` instead of the raw body

#### 2. selectedPolicy can return a disabled policy that still has ev == nil — potential panic path avoided only by ordering

Addressed.

- Made the disabled-policy/evaluator invariant clearer in `CheckPolicy`
- The code now checks `!policyConfig.Enabled` before reading `policy.evaluator`
- Kept the explicit comment that disabled policies intentionally do not initialize an evaluator in `loadPolicy`

#### 3. policyEntryKnownKeys uses flattened dot-notation keys but YAML is parsed as a nested map — inconsistency in validation

Addressed.

- Added a clarifying comment in `pkg/plugin/implementation/opapolicychecker/enforcer.go`
- The comment now explains that `policyEntryKnownKeys` is matched against the post-flattening key shape produced by `loadNetworkPolicies`

#### 4. config/opa-network-policies.yaml has no production-use comment — risk of misuse

Addressed.

- Added the same warning comment to `config/opa-network-policies.yaml` that already existed in `config/onix/opa-network-policies.yaml`
- It now explicitly says the example test policy path must be replaced before production use

nirmay:
<html><head></head><body>
<p>Here's my updated review of PR #661, covering the <strong>7 new commits</strong> landed since the last review (April 16–21).</p>
<hr>
<h2>What Was Addressed ✅</h2>
<p>All 6 issues and the minor notes from the previous review have been fixed:</p>

Previous issue | Resolution
-- | --
Missing test: enabled: false exact match overrides default | TestEnforcer_NetworkPolicyDisabledExactMatchOverridesDefault added
Variadic verification parameter ambiguity | Changed to explicit *ArtifactVerificationConfig everywhere; benchmarks and tests updated
base64 format not documented as standard padded | Inline comment + README clarification added
parseVerificationPublicKeyResponse tries JSON first without comment | Comment added explaining local-file/non-DeDi fallback
No mock HTTP test for DeDi public key lookup | TestEvaluator_FileVerificationWithDeDiHTTPPublicKeyLookup added
Example configs pointing at testdata/ in production | Warning comment added to both opa-network-policies.yaml files
Breaking change not highlighted | README now opens with explicit callout of removed top-level keys


<p>The <code>refactor: parse request context once</code> commit is also a nice improvement — <code>parseRequestContext</code> does a single <code>json.Unmarshal</code> for networkID, action, and all log fields, replacing three separate unmarshal calls.</p>
<hr>
<h2>New Issues</h2>
<h3>1. <code>selectedPolicy</code> second return value is now redundant</h3>
<pre><code class="language-go">// enforcer.go
func (e *PolicyEnforcer) selectedPolicy(networkID string) (*loadedPolicy, string) {
    ...
    return policy, networkID   // networkID was already known at the call site
}
</code></pre>
<p>After the refactor, <code>networkID</code> is extracted before the call and passed in. Returning it back is dead weight. The only downstream use is log formatting. Either return just <code>*loadedPolicy</code> and use the already-available <code>reqCtx.NetworkID</code> for logs, or document why the return is kept. Minor, but a little confusing given the function signature change.</p>
<hr>
<h3>2. <code>enabled: false</code> entries still require <code>type</code>, <code>location</code>, <code>query</code></h3>
<p><code>parsePolicyConfig</code> enforces all three fields as required even when <code>enabled: false</code>. In <code>loadPolicy</code> the evaluator is never created for disabled entries, so the field values are parsed but never used. A user trying to "disable" a network with a minimal entry like:</p>
<pre><code class="language-yaml">networkPolicies:
  retail.network/dev:
    enabled: false
</code></pre>
<p>will get a startup error: <code>'type' is required</code>. There's no test or docs for this edge case. Consider either:</p>
<ul>
<li>Short-circuiting <code>parsePolicyConfig</code> when <code>enabled: false</code> before the required-field checks, <strong>or</strong></li>
<li>Adding a note to the README that disabled entries still need valid <code>type</code>/<code>location</code>/<code>query</code>.</li>
</ul>
<hr>
<h3>3. Network policy selection only tested with <code>networkId</code> (camelCase)</h3>
<p><code>TestEnforcer_NetworkPolicySelection</code> and <code>TestEnforcer_NetworkPolicyDefaultFallback</code> build requests with <code>"networkId"</code> (camelCase). The snake_case variant <code>"network_id"</code> is covered by <code>TestParseRequestContext_NetworkIDVariants</code> in isolation, but there's no end-to-end enforcer test where the request body uses <code>"network_id"</code> and policy selection routes correctly. Low risk given <code>parseRequestContext</code> is well-tested, but worth a one-liner variant in <code>TestEnforcer_NetworkPolicySelection</code>.</p>
<hr>
<h2>Nits</h2>
<ul>
<li>
<p><strong><code>get("action", "action")</code></strong> in <code>parseRequestContext</code> passes <code>"action"</code> as both snake and camel key. The comment says it handles both, but since they're identical strings the second lookup is a no-op. Correct but reads oddly — a comment like <code>// "action" has no camelCase variant</code> would help.</p>
</li>
<li>
<p><strong><code>debugLogging</code> removed from per-policy entries</strong> — the README correctly removes it from the supported per-entry fields list, but the top-level config table still says <code>debugLogging</code> applies to all policies. That's fine and accurate, just confirm that's the intended UX (one global debug switch, not per-network).</p>
</li>
</ul>
<hr>
<h2>Overall Assessment</h2>
<p>The PR is in good shape. The core architecture is sound, concurrency is correct, all my prior issues are resolved, and the new refactor is cleaner. The three new issues above are all minor — <strong>issues 1 and 3 are nits</strong>, <strong>issue 2 is the only one worth a decision</strong> before merge (fix or document). No blockers.</p></body></html>

nirmay:
merged, 
minor nits can be resolved as separate bugs subsequently.

nirmalnr:
Thanks for the review & merge! For the final review comments:

**New Issues raised:**
- **`selectedPolicy` redundant return value** — Tracked in #689.
- **Missing `network_id` snake_case end-to-end test + `get("action", "action")` comment** — Tracked in #690.


**Points not raised as issues:**
- **`enabled: false` requiring type/location/query** — By design. The `enabled` flag is intended for quickly toggling an existing, fully-configured policy entry (e.g. during testing). A user who wants to exclude a network would simply omit the entry entirely rather than defining a minimal disabled stub, so requiring valid fields when `enabled: false` is acceptable.
- **`debugLogging` in README config table** — Intentional. `debugLogging` is a single global switch that applies to all policies, not a per-network setting. The README accurately reflects this UX and no change is needed.

PR #670, opened 2026-04-20
fix(observability): use Beckn context.action for request metric label

SurbhiAgarwal1:
# fix(observability): address mentor feedback on action extraction and observability metrics

The objective of this fix is to ensure that HTTP request metrics and tracing spans use the actual Beckn `context.action` value (e.g., `search`, `select`) instead of the raw URL path, while maintaining a robust and standard-compliant implementation.

### Summary of Changes:
1. **Verified & Fixed `stdHandler.go`**:
2.     - **Early 500 Error**: Added an immediate `http.Error` response if `io.ReadAll(r.Body)` fails, ensuring request state isn't corrupted.
3.     - **Log Message Corrected**: Changed the log message to `failed to read request body` (removed "for metrics") as requested.
4.     - **Added Documentation**: Included the explanatory comment about why the body must be read before establishing the trace context.
5.     - **Span Identity**: Confirmed `spanName` uses `r.URL.Path`, while the Beckn action is stored as a span attribute.
2. **Verified & Added Tests in `stdHandler_test.go`**:
3.     - **Flow Test**: Added `TestServeHTTP_ActionResolution` to verify that `ServeHTTP` correctly extracts the action from a valid body and falls back to the URL path for invalid/empty bodies.
4.     - **Error Test**: Added `TestServeHTTP_BodyReadError` to verify the 500 response.
5.     - **All Tests Passed**: Verified that both the handler and preprocessor tests pass locally.
3. **Cleaned & Pushed**:
4.     - Confirmed that `model.go` and `reqpreprocessor.go` are clean of any accidental whitespace or stale changes.
5.     - Reverted the `spanName` to `r.URL.Path` to follow OTel standards, while still providing the Beckn-specific action as a searchable attribute.
6.

### DIFF
--- a/core/module/handler/stdHandler.go
+++ b/core/module/handler/stdHandler.go
@@ -107,11 +107,14 @@ func (h *stdHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		r.Header.Del("X-Role")
 	}()
 
-	// Read body early to extract Beckn action for metrics/tracing.
+	// Read body early to extract the Beckn action, which is needed for both
+	// the trace context span attributes and for the HTTP request metrics.
+	// This must happen before span creation so the attributes are available.
 	body, err := io.ReadAll(r.Body)
 	if err != nil {
-		log.Errorf(r.Context(), err, "failed to read request body for metrics: %v", err)
-		body = nil
+		log.Errorf(r.Context(), err, "failed to read request body: %v", err)
+		http.Error(w, "failed to read request body", http.StatusInternalServerError)
+		return
 	}
 	r.Body = io.NopCloser(bytes.NewBuffer(body))
 
--- a/core/module/handler/stdHandler.go
+++ b/core/module/handler/stdHandler.go
@@ -3,6 +3,7 @@ package handler
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -106,6 +107,19 @@ func (h *stdHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		r.Header.Del("X-Role")
 	}()
 
+	// Read body early to extract Beckn action for metrics/tracing.
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		log.Errorf(r.Context(), err, "failed to read request body for metrics: %v", err)
+		body = nil
+	}
+	r.Body = io.NopCloser(bytes.NewBuffer(body))
+
+	action := extractBecknAction(body)
+	if action == "" {
+		action = r.URL.Path
+	}
+
 	// to start a new trace
 	propagator := otel.GetTextMapPropagator()
 	traceCtx := propagator.Extract(r.Context(), propagation.HeaderCarrier(r.Header))
@@ -127,13 +141,13 @@ func (h *stdHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	httpMeter, _ := GetHTTPMetrics(r.Context())
 	if httpMeter != nil {
 		recordOnce = func() {
-			RecordHTTPRequest(r.Context(), wrapped.statusCode, r.URL.Path, string(h.role), senderID, receiverID)
+			RecordHTTPRequest(r.Context(), wrapped.statusCode, action, string(h.role), senderID, receiverID)
 		}
 		wrapped.record = recordOnce
 	}
 
 	// set beckn attribute
-	setBecknAttr(span, r, h)
+	setBecknAttr(span, r, h, action)
 
 	stepCtx, err := h.stepCtx(r, w.Header())
 	if err != nil {
@@ -388,7 +402,7 @@ func (h *stdHandler) resolveDirection(ctx context.Context) (senderID, receiverID
 	return remoteID, selfID
 }
 
-func setBecknAttr(span trace.Span, r *http.Request, h *stdHandler) {
+func setBecknAttr(span trace.Span, r *http.Request, h *stdHandler, action string) {
 	senderID, receiverID := h.resolveDirection(r.Context())
 	attrs := []attribute.KeyValue{
 		telemetry.AttrRecipientID.String(receiverID),
@@ -396,6 +410,7 @@ func setBecknAttr(span trace.Span, r *http.Request, h *stdHandler) {
 		attribute.String("span_uuid", uuid.New().String()),
 		attribute.String("http.request.method", r.Method),
 		attribute.String("http.route", r.URL.Path),
+		telemetry.AttrAction.String(action),
 	}
 
 	if trxID, ok := r.Context().Value(model.ContextKeyTxnID).(string); ok {
@@ -421,6 +436,21 @@ func setBecknAttr(span trace.Span, r *http.Request, h *stdHandler) {
 	}
 
 	span.SetAttributes(attrs...)
+}
+
+func extractBecknAction(body []byte) string {
+	if len(body) == 0 {
+		return ""
+	}
+	var req struct {
+		Context struct {
+			Action string `json:"action"`
+		} `json:"context"`
+	}
+	if err := json.Unmarshal(body, &req); err != nil {
+		return ""
+	}
+	return req.Context.Action
 }
 
 func errString(e error) string {
--- a/core/module/handler/stdHandler_test.go
+++ b/core/module/handler/stdHandler_test.go
@@ -9,10 +9,106 @@ import (
 
 	"github.com/beckn-one/beckn-onix/pkg/plugin"
 	"github.com/beckn-one/beckn-onix/pkg/plugin/definition"
+	"github.com/beckn-one/beckn-onix/pkg/telemetry"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/embedded"
 )
 
 // noopPluginManager satisfies PluginManager with nil plugins (unused loaders are never invoked when config is omitted).
 type noopPluginManager struct{}
+
+func TestExtractBecknAction(t *testing.T) {
+	tests := []struct {
+		name     string
+		body     string
+		expected string
+	}{
+		{
+			name:     "Valid body",
+			body:     `{"context": {"action": "search"}}`,
+			expected: "search",
+		},
+		{
+			name:     "Different valid action",
+			body:     `{"context": {"action": "select"}}`,
+			expected: "select",
+		},
+		{
+			name:     "Missing context",
+			body:     `{"other": "data"}`,
+			expected: "",
+		},
+		{
+			name:     "Missing action",
+			body:     `{"context": {"other": "data"}}`,
+			expected: "",
+		},
+		{
+			name:     "Malformed JSON",
+			body:     `{"context": {"action": "search"`,
+			expected: "",
+		},
+		{
+			name:     "Empty body",
+			body:     "",
+			expected: "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := extractBecknAction([]byte(tt.body))
+			if got != tt.expected {
+				t.Errorf("extractBecknAction() = %v, want %v", got, tt.expected)
+			}
+		})
+	}
+}
+
+type mockSpan struct {
+	embedded.Span
+	attributes []attribute.KeyValue
+}
+
+func (m *mockSpan) SetAttributes(attrs ...attribute.KeyValue) {
+	m.attributes = append(m.attributes, attrs...)
+}
+func (m *mockSpan) End(options ...trace.SpanEndOption)                  {}
+func (m *mockSpan) AddEvent(name string, options ...trace.EventOption) {}
+func (m *mockSpan) AddLink(link trace.Link)                             {}
+func (m *mockSpan) IsRecording() bool                                   { return true }
+func (m *mockSpan) RecordError(err error, options ...trace.EventOption) {}
+func (m *mockSpan) SpanContext() trace.SpanContext                      { return trace.SpanContext{} }
+func (m *mockSpan) SetStatus(code codes.Code, description string)       {}
+func (m *mockSpan) SetName(name string)                                 {}
+func (m *mockSpan) TracerProvider() trace.TracerProvider                { return nil }
+
+func TestSetBecknAttr(t *testing.T) {
+	h := &stdHandler{
+		SubscriberID: "test-sub",
+		moduleName:   "test-module",
+	}
+	span := &mockSpan{}
+	req, _ := http.NewRequest("POST", "/test", nil)
+	action := "search"
+
+	setBecknAttr(span, req, h, action)
+
+	found := false
+	for _, attr := range span.attributes {
+		if attr.Key == telemetry.AttrAction {
+			found = true
+			if attr.Value.AsString() != action {
+				t.Errorf("expected action attribute %v, got %v", action, attr.Value.AsString())
+			}
+		}
+	}
+	if !found {
+		t.Error("action attribute not found in span")
+	}
+}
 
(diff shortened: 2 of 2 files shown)


nirmay:
---

## Code Review: [PR #670](https://github.com/beckn/beckn-onix/pull/670) — fix(observability): use Beckn context.action for request metric label

---

### Overview

The goal is correct — the `action` label in HTTP request metrics and tracing spans should use the Beckn `context.action` value rather than the raw URL path. However, the implementation has a design problem in how it achieves this, along with several smaller issues.

---

### Design Problem: Adding `action` to `reqpreprocessor`

The PR routes `context.action` through `reqpreprocessor`: it extracts the value from the request body, stores it in the Go request context via a new `ContextKeyAction`, and `stdHandler` reads it back out.

To evaluate whether this is right, it helps to understand what `reqpreprocessor` actually does today.

**What `reqpreprocessor` really does:**
It reads and buffers the request body, then extracts five categories of values from the Beckn `context` block:
- `bap_id` / `bpp_id` → stored as `ContextKeySubscriberID` — the adapter's own identity
- the opposite role's ID → stored as `ContextKeyRemoteID` — the remote caller's identity
- `parentID` from config → stored as `ContextKeyParentID` — for distributed trace correlation
- A configurable list of generic keys (e.g. `transaction_id`, `message_id`) → stored under their respective context keys

The fields it sets — subscriber ID, remote ID, parent ID — are not primarily observability fields. They are fundamental to the adapter's operation: routing decisions, signature validation, subscriber identity. They appear in metrics as a side-effect of being in context, but they are set because the core pipeline needs them.

**Where the PR goes wrong:**
`context.action` does not fit this pattern. It has no role in routing, identity, or signing. The PR adds it to `reqpreprocessor` purely to serve a metric label in `stdHandler`. This gives `reqpreprocessor` a responsibility it should not have.

**An important nuance — `stdHandler` already depends on `reqpreprocessor`:**
It is true that `stdHandler` already reads `ContextKeySubscriberID` and `ContextKeyRemoteID` (set by `reqpreprocessor`) when building the `senderID`/`receiverID` metric labels. So the dependency is not new in principle. However, those fields earn their place in `reqpreprocessor` because they serve the core pipeline. Adding `action` as a pure observability passenger sets a bad precedent and blurs the plugin's responsibility.

**The right approach:**
`stdHandler` should extract `context.action` directly from the request body itself. `reqpreprocessor` already reads the body and restores it (`r.Body = io.NopCloser(bytes.NewBuffer(body))`), so by the time `stdHandler.ServeHTTP` is entered, the body is a buffered reader and can be read again. A small private `extractBecknAction(body []byte) string` helper in `stdHandler.go` — mirroring the camelCase/snake_case fallback pattern already used elsewhere — is all that is needed. This makes the fix entirely self-contained: no changes to `pkg/model`, no changes to `reqpreprocessor`, no new context key.

---

### Additional Issues

#### 1. `resolve_action_test.go` is an unnecessary new file
Tests for a private helper in `stdHandler.go` belong in `stdHandler_test.go`. One source file, one test file — creating a separate file for a single helper fragments the test surface and makes it harder for a developer reading `stdHandler.go` to find the corresponding tests.

#### 2. `model.go` changes should be reverted
`ContextKeyAction` and its entry in the `contextKeys` map exist only to coordinate the value between `reqpreprocessor` and `stdHandler`. Under the correct approach (body read directly in `stdHandler`), this coordination is not needed. The `contextKeys` map entry additionally triggers log propagation, which is a separate concern and should be a deliberate, separate decision — not a side-effect of this fix.

#### 3. Verify the tracing span attribute actually works
`setBecknAttr` is called at span setup time, at the beginning of `ServeHTTP`. Whether `ContextKeyAction` is in the request context at that point depends on `reqpreprocessor` being HTTP middleware that wraps `stdHandler` (i.e. runs before `ServeHTTP` is entered). If `reqpreprocessor` is instead a step within the handler chain that runs after `ServeHTTP` starts, the span attribute will silently always fall back to the URL path. This should be verified — or a test added that confirms the span attribute receives the Beckn action under real request flow.

#### 4. Test sentinel `"MISSING"` is a code smell
In [`resolve_action_test.go`](core/module/handler/resolve_action_test.go), the string `"MISSING"` is used as a magic sentinel to mean "do not set this key in context." Use a `*string` (nil = not set) or an explicit `bool setAction` field in the test struct instead.

#### 5. Reqpreprocessor test assertion coupled to test case name
In [`reqpreprocessor_test.go`](pkg/plugin/implementation/reqpreprocessor/reqpreprocessor_test.go), the action assertion is gated on `if tt.name == "Context with action"`. This is fragile — renaming the test case silently disables the assertion. Add an `expectedAction string` field to the test struct and assert it unconditionally for all cases.

#### 6. Unrelated `go.mod`/`go.sum` churn
The diff removes a large number of transitive dependencies (`envoyproxy`, `containerd`, `cloud.google.com`, `viper`, `cobra`, `fsnotify`, and others) and adds `github.com/yuin/gopher-lua`. None of this relates to the action label fix. These changes should be explained in the PR description or moved to a separate PR to keep the diff reviewable and bisectable.

---

### Summary

| # | Severity | Issue |
|---|---|---|
| 1 | **Must fix** | `action` extraction does not belong in `reqpreprocessor` — it serves no core pipeline purpose |
| 2 | **Must fix** | `stdHandler` should read `context.action` directly from the buffered body |
| 3 | **Must fix** | `model.go` changes (`ContextKeyAction`, `contextKeys` entry) should be reverted |
| 4 | Should fix | `resolve_action_test.go` should be consolidated into `stdHandler_test.go` |
| 5 | Should fix | Verify (or test) that the span attribute in `setBecknAttr` actually receives the Beckn action |
| 6 | Should fix | Test sentinel `"MISSING"` — use typed flag or pointer instead |
| 7 | Should fix | Reqpreprocessor test assertion coupled to test case name string |
| 8 | Address | Explain or split out unrelated `go.mod`/`go.sum` dependency changes |

nirmay:
## Code Review: [PR #670](https://github.com/beckn/beckn-onix/pull/670) — fix(observability): use Beckn context.action for request metric label (updated)

---

### Overview

The architectural rework from the previous iteration is correct and the core change in `stdHandler.go` is on the right track. However, a few issues remain — some introduced by this revision, some carried over.

---

### Issue 1: Three files have whitespace-only changes — revert them entirely

These files appear in the diff with no substantive change — only blank lines added as leftover artifacts from the previous iteration:

- **[`pkg/model/model.go`](pkg/model/model.go)** — adds a blank line before the closing `)` of the const block
- **[`pkg/plugin/implementation/reqpreprocessor/reqpreprocessor.go`](pkg/plugin/implementation/reqpreprocessor/reqpreprocessor.go)** — adds a spurious double blank line between two code blocks
- **[`pkg/plugin/implementation/reqpreprocessor/reqpreprocessor_test.go`](pkg/plugin/implementation/reqpreprocessor/reqpreprocessor_test.go)** — adds a blank line before a closing `})`

All three should have zero diff against `main`. Revert them completely. Touching files with whitespace-only changes pollutes the commit history and makes `git blame` noisier for no reason.

---

### Issue 2: Setting `spanName = action` is questionable — reconsider

```go
-  spanName := r.URL.Path
+  spanName := action
   traceCtx, span := tracer.Start(traceCtx, spanName, ...)
```

This is a separate concern from the metric label fix and needs to be thought through carefully.

**The problem with using `action` as span name:** OpenTelemetry HTTP server convention expects the span name to represent the route — typically `{method} {route}` or just the route path. A span named `"search"` loses all context about which handler, role, and direction handled it. In a tracing backend (Jaeger, Tempo, etc.), all `"search"` spans across all adapters will be grouped together regardless of whether they came from a BAP, BPP, caller, or receiver path — making the service map and latency analysis misleading.

**It is also redundant:** The action is already recorded correctly as a span *attribute* via `telemetry.AttrAction.String(action)` inside `setBecknAttr`. Span attributes are the right place for Beckn-specific semantic fields. The span *name* is the structural identity of the operation in the trace.

**The original `r.URL.Path` was actually better** as a span name — `/bap/receiver/search` carries role, direction, and action in one string, and conforms to OTel conventions. The metric label fix (`RecordHTTPRequest`) does not require touching the span name at all.

**Recommendation:** Revert `spanName` back to `r.URL.Path`. The action is already properly captured as a span attribute.

---

### Issue 3: Stale duplicate comment

In [`stdHandler.go`](core/module/handler/stdHandler.go):

```go
// to start a new trace          ← leftover, wrong context
body, err := io.ReadAll(r.Body)
...
action := extractBecknAction(body)

// to start a new trace          ← correct, this is where tracing begins
propagator := otel.GetTextMapPropagator()
```

The first `// to start a new trace` is a copy of the comment from the block below — remove it or replace it with something accurate like `// Read body early to extract Beckn action for metrics and tracing.`

---

### Issue 4: Body read error silently swallowed

```go
body, err := io.ReadAll(r.Body)
if err != nil {
    body = nil
}
```

The error is discarded with no log. In production this silently causes all metric labels to fall back to `r.URL.Path` with no observable signal. Add at least a `log.Debugf` or `log.Warnf` so the condition is diagnosable.

---

### Issue 5: PR description is completely stale — must update

The summary still describes the previous, discarded approach:

> - Added ContextKeyAction to model.ContextKey...
> - Modified reqpreprocessor plugin...
> - Added test cases for action extraction in reqpreprocessor_test.go.
> - Added new unit tests for resolveMetricAction in resolve_action_test.go.

None of this reflects the current code. Rewrite the description to accurately describe what this PR actually does.

---

### Issue 6: `go.mod`/`go.sum` churn is unexplained

The diff removes a large number of transitive dependencies and adds `github.com/yuin/gopher-lua`. This is unrelated to the action label fix. Explain it in the updated PR description or split it into a separate PR.

---

### What Is Correct

| Area | Status |
|---|---|
| Fix is self-contained in `stdHandler` — no `model.go`, no `reqpreprocessor` substantive changes | ✅ |
| `extractBecknAction` reads directly from buffered body | ✅ |
| Body properly restored for downstream handlers | ✅ |
| `action` passed as parameter to `setBecknAttr` — ordering issue resolved | ✅ |
| `action` used as span attribute via `telemetry.AttrAction` | ✅ |
| Tests in `stdHandler_test.go`, no separate file | ✅ |

---

### Summary

| # | Severity | Issue |
|---|---|---|
| 1 | **Must fix** | Revert `model.go`, `reqpreprocessor.go`, `reqpreprocessor_test.go` — whitespace-only changes with no purpose |
| 2 | **Must fix** | Revert `spanName = action` back to `r.URL.Path` — non-standard, loses context, redundant with span attribute |
| 3 | Must fix | PR description still describes the old approach — needs a full rewrite |
| 4 | Should fix | Stale `// to start a new trace` comment before the body-reading block |
| 5 | Should fix | Body read error silently dropped — add a log line |
| 6 | Address | Explain or split out unrelated `go.mod`/`go.sum` dependency changes |

nirmay:
## Overview

The change reads the request body early in `ServeHTTP`, extracts `context.action` from the JSON, and uses it as the action label in both `RecordHTTPRequest` and the tracing span — falling back to `r.URL.Path` when the action is absent. The approach is correct in intent.

---

## Issues

### 1. PR description describes a different implementation (misleading)

The description says:
> - Added `ContextKeyAction` to `model.ContextKey`
> - Modified `reqpreprocessor` plugin to extract action
> - Introduced `resolveMetricAction` helper
> - Added tests in `resolve_action_test.go`

None of these exist in the actual diff. The function is named `extractBecknAction`, the tests are in `stdHandler_test.go`, and there are no changes to `model` or `reqpreprocessor`. The description appears to have been written for a different (perhaps earlier) design. Please align the description with what was actually committed.

---

### 2. Body read failure silently corrupts the request

```go
body, err := io.ReadAll(r.Body)
if err != nil {
    log.Errorf(r.Context(), err, "failed to read request body for metrics: %v", err)
    body = nil   // ← nil body
}
r.Body = io.NopCloser(bytes.NewBuffer(body))  // ← replaces with empty reader
```

If `io.ReadAll` fails (e.g., client disconnect mid-send, timeout), `body` is set to `nil`, and `r.Body` is replaced with an empty reader. Execution continues into `h.stepCtx(r, w.Header())` which reads the body for actual processing — it now sees an empty body and will fail with a confusing downstream parse/validation error, masking the real root cause.

**Fix**: Return a 500 error immediately on body read failure rather than continuing:

```go
body, err := io.ReadAll(r.Body)
if err != nil {
    log.Errorf(r.Context(), err, "failed to read request body: %v", err)
    http.Error(w, "failed to read request body", http.StatusInternalServerError)
    return
}
r.Body = io.NopCloser(bytes.NewBuffer(body))
```

---

### 3. Body is read before the trace context is established

```go
body, err := io.ReadAll(r.Body)   // body read here
...
traceCtx := propagator.Extract(r.Context(), ...)  // trace context set here
...
span = tracer.Start(traceCtx, ...)                // span started here
```

The body is consumed before the trace span exists. This means if there's ever a need to log a trace-correlated error during body reading, there's no active span. More practically, the span's `action` attribute is set correctly from the already-resolved `action` variable, so there's no functional bug. But it's worth adding a comment clarifying why body reading must precede the trace setup.

---

### 4. No integration/flow test for action resolution in `ServeHTTP`

`TestExtractBecknAction` tests the extraction helper in isolation. `TestSetBecknAttr` tests the span attribute setter. But there's no test that exercises the full `ServeHTTP` path — specifically:
- That a valid Beckn body causes the correct action to reach `RecordHTTPRequest`
- That a non-Beckn body (e.g., empty or malformed) correctly falls back to `r.URL.Path`

The fallback branch (`action = r.URL.Path`) has zero test coverage.

---

## Nit

- `extractBecknAction` handles `context.action` (snake_case). Beckn spec consistently uses snake_case for this field, so this is fine — just confirm there's no `contextAction` (camelCase) variant in real traffic.
- The log message `"failed to read request body for metrics: %v"` says "for metrics" but the body is actually needed for the full request processing too; the wording undersells the severity.

---

## What's Good

- `io.NopCloser(bytes.NewBuffer(body))` correctly restores the body for downstream reads — the right pattern.
- Fallback to `r.URL.Path` is a safe default.
- `extractBecknAction` is a clean, minimal parser with sensible nil/empty handling.
- Test coverage for `extractBecknAction` is thorough (6 cases including malformed JSON, empty body, missing fields).

---

## Summary

@SurbhiAgarwal1 
Two things to fix before merge: **align the PR description with reality** (issue 1) and **handle body read failure with an early error response** (issue 2). The rest are improvements or documentation.

nirmay:
@SurbhiAgarwal1 are you planning to update the PR description or add a comment here describing the overall change.

nirmay:
## Note on implementation approach

The PR description references changes that were not included in the final implementation. Here is what was actually done:

**What the description says vs. what was implemented:**

| Description says | Actual implementation |
|---|---|
| Added `ContextKeyAction` to `model.ContextKey` | Not added — action is resolved as a local variable in `ServeHTTP` |
| Modified `reqpreprocessor` plugin | No changes to `reqpreprocessor` |
| Introduced `resolveMetricAction` helper | Function is named `extractBecknAction`, not `resolveMetricAction` |
| Tests in `resolve_action_test.go` | Tests are in `stdHandler_test.go`, no new file created |

**What was actually implemented (in `core/module/handler/stdHandler.go`):**
- Added `extractBecknAction(body []byte) string` — parses `context.action` from the Beckn JSON body.
- In `ServeHTTP`, the request body is read early (before span creation), `context.action` is extracted, and the body is restored via `io.NopCloser` for downstream reads.
- Falls back to `r.URL.Path` when the body is empty, malformed, or missing `context.action`.
- Returns HTTP 500 immediately on body read failure rather than continuing with a silently empty body.
- `setBecknAttr` updated to accept and attach the resolved action as `telemetry.AttrAction` on the span.

**Tests added (in `core/module/handler/stdHandler_test.go`):**
- `TestExtractBecknAction`: unit tests for the extraction helper (valid body, missing context, missing action, malformed JSON, empty body).
- `TestSetBecknAttr`: verifies the action is set as a span attribute.
- `TestServeHTTP_ActionResolution`: smoke-tests `ServeHTTP` across valid Beckn body, empty body, and non-Beckn body.
- `TestServeHTTP_BodyReadError`: verifies HTTP 500 is returned when the body reader fails.

Issue #666, opened 2026-04-20
[Observability] When Observability is not configured, warning 'audit rules are not defined for this action send default' should not appear in logs.


nirmay:
[Observability] When Observability is not configured, warning 'audit rules are not defined for this action send default' should not appear in logs.

Tomeshwari-02:
Fixed in [PR #671](https://github.com/beckn/beckn-onix/pull/671). The disabled-telemetry path now returns silently instead of emitting a warning when observability is not configured.

nirmay:
### Impact

`EmitAuditLogs` is called unconditionally on **every request** in [`[core/module/handler/stdHandler.go:153](https://claude.ai/epitaxy/core/module/handler/stdHandler.go:153)`](core/module/handler/stdHandler.go:153). When observability is not configured, `LogsEnabled()` returns `false`, triggering this warning:

```
"failed to emit audit logs, logs disabled"
```

This fires **once per request**, meaning a moderately loaded adapter produces thousands of WARN entries per minute that tell the operator nothing useful — observability is intentionally optional. Operators who haven't configured telemetry see a constant stream of warnings suggesting something is broken when it isn't. Any alerting on WARN-count thresholds will fire continuously as false positives.

> Note: the issue title says "audit rules are not defined for this action send default" — that was likely the wording in a previous version before the audit pipeline rewrite (PR #662). The current warning text in the codebase is "failed to emit audit logs, logs disabled" at [`[pkg/telemetry/audit.go:26](https://claude.ai/epitaxy/pkg/telemetry/audit.go:26)`](pkg/telemetry/audit.go:26).

---

### Root Cause

```go
// audit.go:25-28
if !LogsEnabled() {
    logger.Warnf(ctx, "failed to emit audit logs, logs disabled")  ← fires per-request
    return
}
```

The warning was originally added deliberately — the comment explains the developer wanted to avoid silently dropping records into the OTel no-op provider without any signal. That's a valid concern at development time, but wrong at production level because:

1. Observability is **optional by design** — `LogsEnabled()` being false is a normal, expected state
2. The call site in `stdHandler.go` never checks `LogsEnabled()` before calling — so it relies on `EmitAuditLogs` to guard itself silently
3. The warning fires per-request, not per-startup — making it log noise proportional to traffic, not to the misconfiguration event

---

### Our Fix Plan (before looking at the PR)

The `!LogsEnabled()` early-return path needs to be quiet in production but ideally not completely invisible to developers.

[...]

PR #671, opened 2026-04-20
Quiet audit logs when telemetry is disabled

Tomeshwari-02:
## What changed
- Replaced the disabled-audit warning with a debug breadcrumb.
- Kept the disabled telemetry path quiet for normal production logging while still leaving a trace for developers with debug logging enabled.
- Added a regression test that asserts the disabled path calls the breadcrumb hook exactly once.

## Why
Observability is optional in the config, so the no-telemetry path should not produce a warning on every request. A debug-level breadcrumb preserves diagnosability without generating WARN noise.

## Validation
- `go test ./pkg/telemetry -count=1`

### DIFF
--- a/pkg/telemetry/audit.go
+++ b/pkg/telemetry/audit.go
@@ -6,7 +6,6 @@ import (
 	"encoding/hex"
 	"time"
 
-	logger "github.com/beckn-one/beckn-onix/pkg/log"
 	"github.com/beckn-one/beckn-onix/pkg/model"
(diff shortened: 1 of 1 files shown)


nirmay:
pls see my comments on #666 and update this PR accordingly @Tomeshwari-02

nirmay:
Closing this as there was no revert from author, and have opened a new PR to take this work forward. #699

Issue #673, opened 2026-04-20
Registry client ignores configured retry settings


Tomeshwari-02:
## Problem

The legacy registry client exposes retry configuration in its `Config` struct, and the docs promise that `retry_max`, `retry_wait_min`, and `retry_wait_max` can be tuned. However, the constructor never applies those values to the `retryablehttp.Client`, so every call still runs with the library defaults.

## Affected code
- `core/module/client/registery.go:17-34`
- `core/module/client/registery.go:37-87`
- `CONFIG.md:558-604`

## Why this matters
This makes the retry configuration misleading and removes an important reliability control from registry bootstrap and lookup flows. In transient failure scenarios, operators may believe they have tuned retries, but the client still uses the default retry behavior.

That is especially risky for network onboarding and registry operations, where short-lived failures are common and retries are often the difference between a successful participant setup and a failed one.

## Expected behavior
The client should read the configured retry values and assign them to the underlying `retryablehttp.Client` before sending requests.

## Suggested follow-up
- Apply `RetryMax`, `RetryWaitMin`, and `RetryWaitMax` in `NewRegisteryClient`
- Add a test that verifies the configured retry values are propagated to the client
- Consider renaming `RegisteryURL` while touching this area, but that is optional and separate from the retry bug

nirmay:
@Tomeshwari-02 thank you for filing this bug, here is my short analysis,

## Analysis for the bug report on #673
NewRegisteryClient in core/module/client/registery.go is dead code — the retry bug has no runtime impact.

The function has no production callers (confirmed by import graph search). It is only exercised by its own test file, and those tests do not assert on retry count or timing — so they pass with library defaults and would pass equally after a fix.

The live retry path runs through pkg/plugin/implementation/registry/registry.go, which correctly applies all three retry fields (RetryMax, RetryWaitMin, RetryWaitMax) in its New() constructor. Operators configuring retry via the plugin system are unaffected.

Options: (1) Fix the wiring in NewRegisteryClient as PR #674 proposes. (2) Delete core/module/client/registery.go and its test as unreferenced dead code, since the plugin is the authoritative implementation. 

Option 2 is cleaner but should be confirmed against any external documentation or migration guides that might reference this client.

Tomeshwari-02:
[...]

PR #674, opened 2026-04-20
Remove dead legacy registry client

Tomeshwari-02:
## Summary
This PR removes the legacy registry client implementation in `core/module/client`.

The maintainer confirmed that `NewRegisteryClient` has no production callers and the live registry path already goes through `pkg/plugin/implementation/registry/registry.go`, which correctly applies the configured retry settings. That means the legacy client was dead code rather than a runtime bug.

## What changed
- Deleted `core/module/client/registery.go`
- Deleted `core/module/client/registry_test.go`

## Why this matters
- Removes an unreferenced implementation that duplicated functionality already covered by the registry plugin.
- Reduces confusion around which registry client is authoritative.
- Avoids maintaining tests for code that is not part of the production path.

## Validation
- `go test ./pkg/plugin/implementation/registry ./pkg/plugin/implementation/dediregistry ./pkg/plugin/implementation/registry/cmd -count=1`

## Notes
This change intentionally removes the legacy helper instead of reworking it, since the plugin implementation is the active code path.


nirmay:
**Overall**

The approach here is correct — deleting dead code is cleaner than fixing a bug nobody hits. However, this PR mixes two unrelated concerns and should be scoped down before merging.

---

**Scope issue — please split this PR**

The registry deletion (`core/module/client/registery.go`, `registry_test.go`) and the audit log change (`pkg/telemetry/audit.go`, `audit_test.go`) address completely different issues (#673 and #666 respectively). Bundling them means:

- A reviewer cannot approve the registry cleanup without also taking a position on the audit change
- Git history conflates two independent decisions into one commit, making future bisects and reverts harder
- PR #671 is already open for the #666 fix and is now left in an ambiguous state

**Recommendation:** remove the `pkg/telemetry/` changes from this PR. Keep only the deletion of `core/module/client/registery.go` and `core/module/client/registry_test.go`. The audit fix should either land through PR #671 or a separate PR that explicitly supersedes it.

---

**Registry deletion (`core/module/client/`) — approved as-is**

`NewRegisteryClient` has no production callers (confirmed by import graph — only its own test file references it). The plugin at `pkg/plugin/implementation/registry/` is the live implementation and correctly applies retry settings. Deleting the legacy client and its tests is the right call.

---

**Audit changes (`pkg/telemetry/`) — out of scope here, but the approach is sound**

The debug-level breadcrumb and the injectable `auditDebugf` test seam are the right implementation. No objection to the approach — it just belongs in its own PR.

Tomeshwari-02:
Sorry for the extra scope in this PR.

I’ve removed the `pkg/telemetry` changes, so PR #674 now only includes the registry cleanup in:
- `core/module/client/registery.go`
- `core/module/client/registry_test.go`

Could you please take another look when you have time?

nirmay:
<html><head></head><body><p>Let me verify the "no production callers" claim before signing off on a deletion PR:</p>
<p>Clean. Here's the review:</p>
<hr>
<h2>PR #674 — Remove dead legacy registry client</h2>
<p><strong>Verdict: ✅ Approve. Safe to merge.</strong></p>
<h3>Verification</h3>
<p>All three claims in the PR description check out:</p>

Claim | Verified
-- | --
NewRegisteryClient has no production callers | ✅ Zero results across all non-test .go files
Package core/module/client is not imported anywhere | ✅ No import of "github.com/beckn-one/beckn-onix/core/module/client" exists
Active path is pkg/plugin/implementation/registry/registry.go | ✅ Confirmed — it has config validation, context-aware logging, and proper retry wiring


<p>After the deletion, <code>core/module/client/</code> will be entirely empty (these are the only two files in it). Git removes the directory automatically — no leftover empty package.</p>
<h3>Why the legacy code was worse (confirms it shouldn't be fixed, just deleted)</h3>
<ul>
<li><strong>Retry config silently ignored</strong>: <code>NewRegisteryClient</code> creates a <code>retryablehttp.Client</code> with defaults and never applies the <code>RetryMax</code>/<code>RetryWaitMin</code>/<code>RetryWaitMax</code> from <code>Config</code> — the config fields are stored but unused.</li>
<li><strong>No config validation</strong>: a nil or empty <code>RegisteryURL</code> would produce a silent bad request rather than a clear startup error.</li>
<li><strong>No context propagation</strong>: neither <code>Subscribe</code> nor <code>Lookup</code> passes <code>ctx</code> to the HTTP request.</li>
<li><strong>Typo baked into the public API</strong>: <code>NewRegisteryClient</code>, <code>RegisteryURL</code>, <code>registery.go</code> — would require a breaking rename to fix.</li>
</ul>
<p>All of these are solved in the plugin implementation. Deletion is the right call.</p>
<h3>Minor notes</h3>
<ul>
<li>The commit history is noisy (6 commits including reverted telemetry work), but the final diff is exactly right — two files deleted, nothing else. Squash on merge.</li>
<li>The PR description's validation command only runs registry plugin tests. Running <code>go build ./...</code> after the deletion would be a stronger check, but since the package has no callers, it's guaranteed to pass.</li>
</ul></body></html>

nirmay:
@ashishkgGoogle Can you pls confirm that you do not have any downstream Registry implementation that is using the code that is being cleaned up in this PR?

ashishkgGoogle:
> @ashishkgGoogle Can you pls confirm that you do not have any downstream Registry implementation that is using the code that is being cleaned up in this PR?

Confirmed.

Issue #675, opened 2026-04-21
dediregistry plugin silently drops retry config fields (retry_max, retry_wait_min, retry_wait_max)


nirmay:
## Problem

The dediregistry plugin's \`cmd/plugin.go\` parses \`timeout\` from the config map but never parses \`retry_max\`, \`retry_wait_min\`, or \`retry_wait_max\`. As a result, those fields in \`dediregistry.Config\` are always zero, so the underlying \`retryablehttp.Client\` silently falls back to library defaults regardless of what the operator configures.

## Affected code

- \`pkg/plugin/implementation/dediregistry/cmd/plugin.go:25-43\` — Config struct is populated without retry fields
- \`pkg/plugin/implementation/dediregistry/dediregistry.go:62-71\` — Retry settings are correctly wired to the client, but only if the Config fields are non-zero

## Root cause

\`cmd/plugin.go\` builds \`dediregistry.Config\` from a \`map[string]string\` and explicitly handles \`timeout\`, \`allowedNetworkIDs\`, etc., but has no equivalent parsing block for the three retry fields. The \`dediregistry.Config\` struct has supported these fields all along.

## Contrast with registry plugin

The registry plugin's \`cmd/plugin.go\` correctly parses all three retry fields using \`strconv.Atoi\` and \`time.ParseDuration\` before passing them to the client constructor. The dediregistry plugin should do the same.

## Expected behavior

When an operator sets \`retry_max\`, \`retry_wait_min\`, or \`retry_wait_max\` in the dediregistry plugin config, those values should be parsed and forwarded to \`dediregistry.Config\` so the underlying HTTP client respects them.

## Suggested fix

In \`pkg/plugin/implementation/dediregistry/cmd/plugin.go\`, add parsing for the three retry fields after the timeout block, mirroring the pattern in \`pkg/plugin/implementation/registry/cmd/plugin.go\`:

\`\`\`go
if retryMaxStr, exists := config["retry_max"]; exists && retryMaxStr != "" {
    if retryMax, err := strconv.Atoi(retryMaxStr); err == nil {
        dediConfig.RetryMax = retryMax
    } else {
        log.Warnf(ctx, "Invalid retry_max value '%s', using default", retryMaxStr)
    }
}
if retryWaitMinStr, exists := config["retry_wait_min"]; exists && retryWaitMinStr != "" {
    if d, err := time.ParseDuration(retryWaitMinStr); err == nil {
        dediConfig.RetryWaitMin = d
    } else {
[...]

PR #686, opened 2026-04-22
Parse DeDi registry retry config

Tomeshwari-02:
## Summary

This PR adds parsing for the DeDi registry plugin retry settings so operator config values are forwarded instead of being silently ignored.

## What changed

- Parse `retry_max`, `retry_wait_min`, and `retry_wait_max` in `pkg/plugin/implementation/dediregistry/cmd/plugin.go`
- Keep the existing timeout and allowed network ID handling intact
- Add tests covering config parsing and forwarding into the DeDi client constructor

## Why

The DeDi plugin already supports retry configuration in `dediregistry.Config`, but the command layer never populated those fields from the config map. That meant the HTTP client always fell back to library defaults.

## Validation

- `go test ./pkg/plugin/implementation/dediregistry/cmd -count=1`

### DIFF
--- a/pkg/plugin/implementation/dediregistry/cmd/plugin.go
+++ b/pkg/plugin/implementation/dediregistry/cmd/plugin.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/beckn-one/beckn-onix/pkg/log"
 	"github.com/beckn-one/beckn-onix/pkg/plugin/definition"
@@ -15,19 +16,16 @@ import (
 // dediRegistryProvider implements the RegistryLookupProvider interface for the DeDi registry plugin.
 type dediRegistryProvider struct{}
 
-// New creates a new DeDi registry plugin instance.
-func (d dediRegistryProvider) New(ctx context.Context, config map[string]string) (definition.RegistryLookup, func() error, error) {
-	if ctx == nil {
-		return nil, nil, errors.New("context cannot be nil")
-	}
+// newDediRegistryFunc creates a new DeDi registry instance.
+var newDediRegistryFunc = dediregistry.New
 
-	// Create dediregistry.Config directly from map - validation is handled by dediregistry.New
+func (d dediRegistryProvider) parseConfig(ctx context.Context, config map[string]string) *dediregistry.Config {
 	dediConfig := &dediregistry.Config{
 		URL:          config["url"],
 		RegistryName: config["registryName"],
 	}
 
-	// Parse timeout if provided
+	// Parse timeout if provided.
 	if timeoutStr, exists := config["timeout"]; exists && timeoutStr != "" {
 		if timeout, err := strconv.Atoi(timeoutStr); err == nil {
 			dediConfig.Timeout = timeout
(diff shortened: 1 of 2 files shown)


Tomeshwari-02:
@nirmay please take a look when you have a chance. I’ve kept this PR scoped to the DeDi retry config fix for #675 and verified it with targeted tests.

nirmay:
Thanks for picking this up — the approach is right and the test coverage for the happy path is solid. Two things to address before merging:

## 1. Error handling inconsistency

`parseConfig` currently returns only `*dediregistry.Config` and logs a `Warnf` on invalid values, silently falling back to defaults. The registry plugin (`registry/cmd/plugin.go`) sets the pattern we should follow — `parseConfig` returns `(*Config, error)` and propagates invalid config as a hard error to the caller. Silently ignoring a misconfigured retry value at startup is harder to debug than a clear error, so please align with that pattern here.

## 2. Missing negative-path tests

There are no test cases covering invalid values (e.g. `retry_max: "abc"`, `retry_wait_min: "notaduration"`). Please add those alongside the error-return fix so we have regression coverage if the behaviour changes.

Once those two are addressed this should be ready to merge. Happy to review the follow-up quickly.

Tomeshwari-02:
Thanks for the review. I have addressed both points now:

- Updated DeDi registry parseConfig to return an error for invalid retry config, matching the registry plugin pattern.
- Added negative-path tests for invalid retry_max, retry_wait_min, and retry_wait_max values.

Validation:

```
go test ./pkg/plugin/implementation/dediregistry/cmd -count=1
```

pkg/plugin/implementation/dediregistry/cmd/plugin.go
```diff
@@ -6,34 +6,82 @@ import (
...
+		retryWaitMax, err := time.ParseDuration(retryWaitMaxStr)
+		if err != nil {
+			return nil, fmt.Errorf("invalid retry_wait_max value '%s': %w", retryWaitMaxStr, err)
+		}
+		if retryWaitMax < 0 {
+			return nil, fmt.Errorf("retry_wait_max must be non-negative, got %v", retryWaitMax)
+		}
+		dediConfig.RetryWaitMax = retryWaitMax
+	}
+
```
nirmalnr, line 74: #### Missing cross-field validation: `retry_wait_min` > `retry_wait_max`

pkg/plugin/implementation/dediregistry/cmd/plugin.go validates each field independently but not their relationship. Setting retry_wait_min=5s, retry_wait_max=1s passes parseConfig and reaches the retryablehttp client with inconsistent values

pkg/plugin/implementation/dediregistry/cmd/plugin.go
```diff
@@ -6,34 +6,82 @@ import (
...
-	// Create dediregistry.Config directly from map - validation is handled by dediregistry.New
+func (d dediRegistryProvider) parseConfig(config map[string]string) (*dediregistry.Config, error) {
 	dediConfig := &dediregistry.Config{
 		URL:          config["url"],
 		RegistryName: config["registryName"],
 	}
 
-	// Parse timeout if provided
+	// Parse timeout if provided.
 	if timeoutStr, exists := config["timeout"]; exists && timeoutStr != "" {
```
nirmalnr, line 28: `timeout` lacks the non-negativity guard the retry fields have

nirmay:
I have updated the commit with some other issues fixed, and have requested for a review.

pkg/plugin/implementation/dediregistry/cmd/plugin_test.go
```diff
@@ -68,25 +254,18 @@ func TestDediRegistryProvider_New_InvalidConfig(t *testing.T) {
...
+	_, _, err := provider.New(context.Background(), map[string]string{
 		"url":          "https://test.com/dedi",
 		"registryName": "subscribers.beckn.one",
 		"timeout":      "invalid",
+	})
+	if err == nil {
+		t.Fatal("expected New() to return an error for invalid timeout")
 	}
-
-	// Invalid timeout should be ignored, not cause error
```
nirmalnr, line 80: A short comment like // invalid timeout must now be a hard error, not a warn-and-continue would preserve the intent change for future readers.

nirmalnr (commented):
@Tomeshwari-02 Could you please check these comments. The PR looks good overall.

nirmay:
Addressed nirmalnr's review comments in the latest two commits:

- Added a positive guard on `timeout` (zero and negative values now return an error, consistent with retry fields)
- Added cross-field validation: `retry_wait_min` must not exceed `retry_wait_max`
- Added test cases for zero/negative timeout and inverted wait bounds
- Added a comment to the `InvalidTimeout` test clarifying the hard-error behaviour

nirmalnr (approved):

Issue #649, opened 2026-04-08
Add manifest-backed policy support to `opapolicychecker`


nirmalnr:
## Background

`opapolicychecker` supports static policy configuration through direct `type`, `location`, and `query` settings, and planned changes are moving it toward a richer multi-policy configuration keyed by `network_id` with optional `default` fallback as per #643.

For network-managed policy distribution, OPA should be able to resolve policy source information indirectly through a verified manifest instead of requiring the policy location to be declared directly in config.

## Problem

There is currently no OPA configuration mode that:

- allows a policy entry to be backed by a manifest rather than a direct file, URL, or bundle location
- loads the manifest through the manifest loader plugin
- extracts OPA-specific policy source information from the manifest
- initializes the evaluator from the resolved policy source

## Proposed change

Build this on top of the planned multi-policy OPA config rather than as a separate configuration system.

The policy map key already represents the `network_id`, so it should not be repeated inside each entry. Add a new `type: manifest` option alongside the existing static source types such as `file`, `url`, `dir`, and `bundle`.

Example config shape:

```yaml
checkPolicy:
  id: opapolicychecker
  config:
    networkPolicyConfig: ./config/opa-network-policies.yaml
```

Example structured config:

```yaml
networkPolicies:
  retail.network/production:
    type: manifest

  retail.network/sandbox:
    type: manifest

  default:
    type: file
    location: ./policies/default.rego
    query: "data.default.policy.result"
```

Behavior for `type: manifest`:

- OPA selects the policy entry using `context.networkId`
- if the selected entry has `type: manifest`, OPA requests the manifest from the manifest loader using that network ID
- OPA parses the returned raw manifest bytes
- OPA extracts and validates the OPA-specific policy source configuration from the manifest
- OPA initializes the evaluator using the resolved file or bundle source

OPA should keep its periodic refresh behavior. On each refresh it should re-evaluate the selected policy source; for `type: manifest`, that means re-requesting the manifest from the manifest loader. Whether the manifest is actually refetched from remote sources depends on the manifest loader cache TTL / freshness policy.

## Acceptance criteria

- Existing static policy source types continue to work unchanged.
- OPA supports multi-policy configuration keyed by `network_id` with optional `default` fallback.
- OPA supports `type: manifest` as a policy source type.
- In `type: manifest` mode, OPA obtains manifest content through the manifest loader dependency.
- OPA parses raw manifest bytes and extracts OPA-specific policy source configuration from them.
[...]

Issue #648, opened 2026-04-08
Add a generic manifest loader plugin


nirmalnr:
## Background

ONIX needs a reusable way to fetch and verify a manifest, cache the verified result, and make it available to multiple consumers such as OPA and future plugins.

The existing codebase already has one special dependency-wired plugin pattern: `KeyManager` is constructed with `Cache` and `RegistryLookup`. The manifest loader should follow the same approach rather than introducing a separate dependency mechanism.

## Problem

There is currently no shared plugin that can:

- resolve manifest-related metadata
- fetch manifest artifacts
- verify signatures
- cache verified results
- return manifest content to multiple consumers

If each consumer implements this independently, the logic will be duplicated and cache reuse will be inconsistent.

The current lookup path uses `network_id` to locate manifest metadata, but the loader itself should remain generic with respect to manifest content.

## Proposed change

Add a new module-level manifest loader plugin with explicit dependency wiring similar to `KeyManager`.

The plugin should:

- support two entry paths:
  - a high-level path that accepts `network_id`
  - a lower-level path that accepts resolved manifest metadata directly
- in the `network_id` path, resolve registry metadata through a DeDi-backed metadata lookup dependency
- fetch the manifest and detached signature
- fetch the signing public key via the provided lookup URL
- validate the manifest signature
- store the verified manifest in shared cache
- return the manifest bytes and verification metadata

The plugin should also support configurable manifest freshness through cache TTL / revalidation policy. The loader should not refetch the manifest on every consumer call if a cached verified manifest is still fresh.

This plugin should stay generic. It should not impose an OPA-specific schema or a manifest-type-specific structure on the manifest content.

Suggested interface:

```go
type ManifestLoader interface {
    GetByNetworkID(ctx context.Context, networkID string) (*model.ManifestDocument, error)
    GetByMetadata(ctx context.Context, metadata model.ManifestMetadata) (*model.ManifestDocument, error)
}
```

Suggested metadata input:

```go
type ManifestMetadata struct {
    ManifestURL               string
    ManifestSignatureURL      string
    SigningPublicKeyLookupURL string
}
```

Suggested result shape:

```go
type ManifestDocument struct {
    NetworkID    string
    ContentType  string
    Content      []byte
    Digest       string
    SourceURL    string
    SignatureURL string
    Verified     bool
    FetchedAt    time.Time
}
```

## Acceptance criteria

- A caller can request a manifest by `network_id`.
[...]

Issue #647, opened 2026-04-08
Extend `dediregistry` to support registry metadata lookup


nirmalnr:
## Background

The current `dediregistry` plugin supports record-level lookup through the existing `RegistryLookup.Lookup(...)` contract. That flow is used for participant key lookup and signature validation.

For registry-level integrations, ONIX needs registry metadata instead of a specific record. In DeDi, registry metadata is exposed under `meta` and can contain arbitrary key/value pairs defined by the registry owner.

## Problem

The current `RegistryLookup.Lookup(...)` path is record-oriented:

- input expects record identity fields such as `subscriber_id` and `key_id`
- output is mapped into `model.Subscription`

That makes it the wrong contract for retrieving registry metadata.

## Proposed change

Add a separate registry-metadata lookup capability to the `dediregistry` plugin while keeping the existing `RegistryLookup` behavior unchanged.

The DeDi plugin should remain generic and should not be responsible for understanding `network_id`. Any mapping from `network_id` to registry lookup parameters should happen outside the plugin.

The DeDi plugin should also remain metadata-agnostic. It should fetch and return all metadata fields present under `meta` without hardcoding manifest-specific keys or validating which keys must exist.

Suggested interface:

```go
type RegistryMetadataLookup interface {
    LookupRegistry(ctx context.Context, nfoDomain, registryName string) (*model.RegistryMetadata, error)
}
```

Suggested result shape:

```go
type RegistryMetadata struct {
    NFODomain    string
    RegistryName string
    RawMeta      map[string]string
}
```

## Acceptance criteria

- `dediregistry` supports a registry-level lookup path without requiring a record ID.
- The returned result is typed and exposes all metadata fields as a generic key/value map.
- Existing `RegistryLookup.Lookup(...)` behavior remains backward compatible.
- Unit tests cover:
  - successful registry metadata lookup
  - malformed response body
  - missing `meta`
  - non-200 responses

## Out of scope

- downloading the manifest
- validating the manifest signature
- storing manifest data in cache
- interpreting manifest structure
- OPA integration

## Notes

- This should be implemented as a separate capability from record lookup, not by overloading `RegistryLookup.Lookup(...)`.
- Parsing `network_id` into `(nfoDomain, registryName)` belongs outside `dediregistry`.
- Manifest-specific metadata validation belongs in the manifest loader or another consumer, not in `dediregistry`.

**Please refer to #619 for more context on this issue.**

nirmalnr:
Addressed in #688

PR #688, opened 2026-04-23
Add DeDi-backed manifest loading and manifest-backed OPA network policies

nirmalnr:
## Summary

This PR adds the missing network-managed policy distribution path for ONIX:

- DeDi registry metadata lookup using `network_id`
- a generic manifest loader plugin
- manifest-backed policy resolution in `opapolicychecker`

It also includes the supporting config, docs, examples, and shared detached artifact verification needed to make that flow usable end to end.

### DeDi registry metadata lookup

- added registry-level metadata lookup support to the DeDi registry plugin
- introduced a generic `RegistryMetadata` model and `RegistryMetadataLookup` interface
- enabled resolving manifest-related metadata for a given `network_id`

### Generic manifest loader plugin

- added a new `manifestloader` plugin
- supports:
  - `GetByNetworkID(...)`
  - `GetByMetadata(...)`
- fetches:
  - manifest
  - detached signature
  - signing public key
- verifies the manifest signature
- caches verified manifests for reuse by downstream plugins
- added cache usability controls:
  - `forceRefreshOnStartup`
  - `disableCache`
- added clearer cache hit / miss / bypass logging

### Manifest-backed OPA policy loading

- extended `opapolicychecker` to support network policy maps keyed by `network_id`
- added `type: manifest` support
- wired `ManifestLoader` into policy checker creation
- for `type: manifest`, OPA now:
  - fetches the verified manifest by network ID
  - validates the manifest schema relevant to OPA
  - validates governance window and network ID match
  - resolves the manifest policy source into `file` or `bundle`
  - reuses the existing evaluator path for the resolved source
- startup logs for manifest-backed policies now show:
  - `sourceType=manifest`
  - `resolvedType=file|bundle`
  - manifest signing / verification details

### Shared detached artifact verification

- extracted shared detached artifact verification into `pkg/security/artifactverifier`
- reused it in:
  - `manifestloader`
  - `opapolicychecker` single-file verification

### OPA/plugin/config cleanup included in this branch

- moved OPA to structured `networkPolicyConfig`
- removed the old flat single-policy config mode
- simplified policy source types to:
  - `file`
  - `dir`
  - `bundle`
  - `manifest`
- made:
  - `file` support local path or remote URL
  - `bundle` support local path or remote URL
- updated refresh config to duration-based `refreshInterval`
- moved OPA verification config into a nested `verification` block
- added DeDi public key response parsing for signed artifacts

### Docs and examples

- updated OPA and manifest loader READMEs
- updated example configs to use the new OPA and manifest loader config model
- added / updated example network policy config files
- documented manifest-backed policy setup and manifest cache behavior

## Why

This completes the intended flow where an NFO can publish a signed network manifest, expose its location through registry metadata, and let ONIX resolve and enforce network policies dynamically based on `network_id` rather than hardcoding policy artifact locations directly in adapter config.

Addresses issues:
#647 Extend dediregistry to support registry metadata lookup
#648 Add a generic manifest loader plugin
#649 Add manifest-backed policy support to opapolicychecker


nirmay:
## PR #688 — DeDi-backed manifest loading and manifest-backed OPA network policies

### Overview

This PR completes the network-managed policy distribution path: NFOs can publish a signed manifest, expose it via DeDi registry metadata, and ONIX can resolve and enforce policies dynamically from it. The three main additions are:

- `dediregistry.LookupRegistry` — registry-level metadata lookup
- `manifestloader` plugin — fetch, verify, cache verified manifests
- `opapolicychecker` extension — new `type: manifest` policy source

Supporting work includes extracting `pkg/security/artifactverifier` from `opapolicychecker` and wiring `ManifestLoader` into the plugin dependency graph.

---

### Code Quality and Style

**Strengths:**
- Test coverage is excellent and deliberate — cache hit/miss, force refresh, expired manifests, network ID mismatch, missing policies section, and loader error paths are all covered
- Structured log messages are informative and consistent (`sourceType=manifest resolvedType=file|bundle manifestSigned=... manifestVerified=...`)
- The `10 MiB` cap on fetched artifacts and the `io.LimitReader` usage are good defensive choices
- Dual-key caching (network key + metadata hash key) correctly avoids re-fetching the manifest when both `GetByNetworkID` and `GetByMetadata` paths are used
- `shouldBypassCache` mutex is correctly scoped to `refreshedKeys`
- `validateNetworkManifest` is comprehensive and catches subtle cases (expired window, `effective_until` before `effective_from`, missing sub-sections when `source=bundle`)

---

### Issues

#### Security

**1. `findStringField` is an unbounded recursive JSON search** ([`[artifactverifier.go](https://claude.ai/epitaxy/pkg/security/artifactverifier/artifactverifier.go)`](pkg/security/artifactverifier/artifactverifier.go))

The function traverses the *entire* JSON tree of the public key endpoint response, looking for keys named `signing_public_key`, `public_key`, or `publicKey`. If an attacker can influence the response body (e.g., via a compromised CDN that wraps the key in a larger JSON envelope), this could extract key material from an unexpected location. For security-sensitive material like signing keys, restrict to top-level fields only, not recursive descent.

**2. Length-based Ed25519 key detection** ([`[artifactverifier.go](https://claude.ai/epitaxy/pkg/security/artifactverifier/artifactverifier.go:2993-2995)`](pkg/security/artifactverifier/artifactverifier.go:2993-2995))

```go
if len(decoded) == ed25519.PublicKeySize {
    return ed25519.PublicKey(decoded), nil
}
```

Any 32-byte value that fails `x509.ParsePKIXPublicKey` is silently accepted as an Ed25519 key. This is an overly broad fallback — a truncated ECDSA key or garbage data of the right length would pass. Consider requiring explicit Ed25519 raw-bytes only when the format is explicitly `base64` and a known length, or remove this fallback.

**3. No HTTPS enforcement for policy artifact URLs**

Manifests can reference `http://` URLs for `policies.file.url` / `policies.bundle.url`. For *signed* artifacts this is acceptable (the signature protects content), but for unsigned ones (`signed: false`) a MITM can inject arbitrary Rego. Consider logging a warning or rejecting cleartext URLs when verification is disabled.

---

#### Design / Architecture

**4. `networkManifest` YAML schema types are private to `opapolicychecker`**

The manifest format (`networkManifest`, `networkManifestPolicies`, `networkManifestBundle`, `networkManifestFile`) is defined only inside `opapolicychecker`. The `manifestloader` plugin returns raw `[]byte` content which `opapolicychecker` then re-parses. If another future consumer needs to parse a manifest, it would need to duplicate these structs. Consider promoting the manifest schema types to `pkg/model` or a `pkg/manifest` package.

**5. Manifest reload cycle is implicitly coupled to manifest cache TTL**

On each OPA policy refresh (`reloadNetworkPolicies`), `resolveManifestPolicyConfig` is called for every `type: manifest` entry, which calls `manifestLoader.GetByNetworkID`. Whether a new manifest is actually fetched depends on the manifest cache TTL vs. the OPA `refreshInterval`. An operator setting a short OPA `refreshInterval` and a long manifest `cacheTTL` will not see manifest updates until the cache expires. This is the correct behavior, but it is subtle and only partially explained in the README. A code comment in `reloadNetworkPolicies` or `resolveManifestPolicyConfig` would help.

**6. Type-assertion-at-startup is the only check that registry implements `RegistryMetadataLookup`**

In [`[stdHandler.go](https://claude.ai/epitaxy/core/module/handler/stdHandler.go:68-71)`](core/module/handler/stdHandler.go:68-71):
```go
metadataLookup, ok := registry.(definition.RegistryMetadataLookup)
if !ok {
    return nil, fmt.Errorf(...)
}
```
This is correct, but the failure only surfaces at process startup. An operator using a non-DeDi registry (e.g., a custom registry that only implements `RegistryLookup`) will get a runtime startup failure with no config-level validation. A note in the README that `manifestLoader` requires a registry plugin implementing `RegistryMetadataLookup` would help operator clarity.

---

#### Minor

**7. `_ = ctx` in `manifestloader.New`**
The context is accepted but discarded. If it's not needed, it should either be removed from the signature (but `ManifestLoaderProvider.New` requires it) or at minimum not need the blank assignment. Consider using it for a startup log line instead.

**8. `CacheKey()` with crypto code in `pkg/model`**
`ManifestMetadata.CacheKey()` imports `crypto/sha256` and `encoding/hex` into the model package. The model package currently has no other crypto dependencies. This is a minor layering concern — the cache key logic could live in `manifestloader` instead.

**9. Missing test for non-200 HTTP status in `TestLookupRegistry`**
The `LookupRegistry` implementation correctly handles non-200 responses, but there's no test for this error path in `dediregistry_test.go`. Given that the existing `TestLookup` suite covers HTTP error status, it would be consistent to add it here too.

**10. Closer panic in `ManifestLoader` plugin manager** ([`[manager.go](https://claude.ai/epitaxy/pkg/plugin/manager.go:2863)`](pkg/plugin/manager.go:2863))
```go
if err := closer(); err != nil {
    panic(err)
}
```
This is consistent with existing closer registrations in the manager, so not a regression. But it is worth noting: a manifest loader that fails on close would crash the adapter. Since the current `manifestloader.New` returns `func() error { return nil }`, this is harmless in practice today, but it is a latent risk for future implementations.

---

### Test Coverage

Coverage is strong. A few gaps worth noting:

- No test for `LookupRegistry` with a non-200 HTTP status (see point 9)
- No test for `manifestloader.GetByNetworkID` when `metadataFromRegistry` returns an error (missing `manifest_url` in registry metadata) — this path exists in the code but is only tested implicitly
- `TestEnforcer_ManifestBackedFilePolicy` constructs manifest YAML using `strings.Join` with no trailing newline after `validManifestGovernanceYAML()` removal — verify this parses correctly in CI (it looks like it does, but it is fragile)

---

### Summary

This is a well-structured and security-conscious PR. The core flow (manifest fetch → signature verify → cache → manifest-backed OPA load) is correct, and the test suite is thorough. The main issues worth addressing before merge are the recursive `findStringField` and the length-based Ed25519 heuristic, both in `artifactverifier`. The architectural concern about private manifest schema types and the reload/cache TTL coupling are worth tracking as follow-ups if not addressed now.

nirmay:
@711ayush711 @ameetdesh please ensure that downstream networks that are using OPA policy checker are updated with these changes. We are targeting this for the April 2026 release.

@nirmalnr

nirmalnr:
**1. `findStringField` is an unbounded recursive JSON search**

**Addressed.** The artifact verifier no longer performs recursive JSON field discovery for public keys. It now accepts only explicit supported response shapes (`data.details.publicKey`, `data.details.signing_public_key`, `data.details.public_key`, plus legacy top-level key fields), so key material in arbitrary nested envelopes is ignored. I also added tests for the supported `data.details.signing_public_key` path and for rejecting unexpected nested key fields.

Updated docs for `manifestloader` and `opapolicychecker` to document the accepted public-key and detached-signature payload shapes.

**2. Length-based Ed25519 key detection** 

This is left as is for now. This will not have any security implications as if the key is wrong the signature verification will fail regardless. Since there are discussions on changing the public key schema in DeDi, this part will have to updated post the DeDi updates in any case.

**3. No HTTPS enforcement for policy artifact URLs**

**Addressed.** Added a warning in `resolveManifestPolicyConfig` for both `bundle` and `file` source types. The check is placed as an `else if` on the `Signed` block so it only fires when signing is **off** and the URL scheme is `http://` — no warning for `https://` unsigned artifacts.

Went with warn-only (not hard reject) to stay consistent with the rest of the codebase, which accepts both `http://` and `https://` everywhere and only errors on unsupported schemes like `ftp://`. A hard rejection here would be a breaking change for any operator currently using unsigned HTTP-served policies.

nirmalnr:
**4. `networkManifest` YAML schema types are private to `opapolicychecker`**

Moved the network manifest schema types (`NetworkManifest`, `NetworkManifestPolicies`, `NetworkManifestBundle`, `NetworkManifestFile`, etc.) out of `opapolicychecker` into a new `pkg/manifest` package (`pkg/manifest/network_manifest.go`).

The `manifestloader` was intentionally left format-agnostic — it returns raw `[]byte` and each consumer parses according to its own schema. `pkg/manifest` is now the shared vocabulary for manifest formats: any plugin that needs to parse a network manifest imports `pkg/manifest` without depending on `manifestloader` internals or on `opapolicychecker`. Future manifest types can be added as additional files in the same package.

`pkg/model` was not used for this since it holds transport/cache types (`ManifestMetadata`, `ManifestDocument`) and pulling YAML-tagged domain structs there would mix layers and introduce a YAML dependency into a package that has none today.

The interfaces of the manifest loader and whether it needs to be redesigned can be taken up as a separate ticket if needed.

**5. Manifest reload cycle is implicitly coupled to manifest cache TTL**

Added a comment in `resolveManifestPolicyConfig` directly above the `GetByNetworkID` call clarifying that OPA's `refreshInterval` drives re-resolution of manifest-backed policies, but manifest freshness is owned entirely by `manifestloader` — a newer manifest is only fetched when the cache entry is expired, bypassed, or disabled by manifestloader configuration. Also added a note in the OPA plugin README.

**6. Type-assertion-at-startup is the only check that registry implements `RegistryMetadataLookup`**

Added a `## Requirements` section to the manifestloader README noting that it requires a registry plugin implementing `RegistryMetadataLookup`. The `dediregistry` plugin supports this; custom registry plugins that only implement `RegistryLookup` cannot be used with `manifestloader` until they add metadata lookup support. The failure surfaces at startup with a clear error message, but this makes the requirement explicit at the configuration level.

The startup error was already expl

The startup error is already explicit:
```Registry plugin does not implement RegistryMetadataLookup```

**Comment by nirmalnr** [comment]
**7. `_ = ctx` in `manifestloader.New`**
Fixed — removed the blank assignment and used `ctx` for a startup `log.Infof` line recording the effective `cacheTTL` and `fetchTimeout` after defaults are applied, consistent with the log style used throughout the rest of the file. (`d8d2846`)

**8. `CacheKey()` with crypto code in `pkg/model`**
Fixed — removed `CacheKey()` from `ManifestMetadata` and moved the logic to a private `metadataCacheKey` function in `manifestloader`, alongside the existing `networkCacheKey`. `pkg/model` no longer imports `crypto/sha256` or `encoding/hex`. (`eb5e1b7`)

**9. Missing test for non-200 HTTP status in `TestLookupRegistry`**
Fixed — added an `"http error response"` subtest to `TestLookupRegistry` matching the pattern already used in `TestLookup`. Pre-existing gap, but addressed here since it's a small fix.

**10. Closer panic in `ManifestLoader` plugin manager**
Acknowledged. The panic-on-close pattern is consistent with every other plugin registration in `manager.go` and is not a regression introduced by this PR. The current `manifestloader` closer always returns `nil` so this is harmless in practice today. Noted as a latent risk for future implementations.

**Comment by nirmay** [comment]
<html><head></head><body>
<hr>
<h2>PR #688 — Updated Review (latest push)</h2>
<h3>What Changed in This Update</h3>
<p>The update systematically addresses every issue from the first review. Here's the accounting:</p>

Prior issue | Status | How
-- | -- | --
#1 — Recursive findStringField | ✅ Fixed | Replaced with struct-typed extractPublicKeyField + extractTopLevelStringField
#2 — Length-based Ed25519 heuristic | ⚠️ Still present | See below, planned to do separately.
#3 — No HTTPS warning for unsigned policies | ✅ Fixed | log.Warnf added in resolveManifestPolicyConfig
#4 — Private manifest schema types | ✅ Fixed | pkg/manifest package introduced
#5 — Reload/cache TTL coupling undocumented | ✅ Fixed | Inline comment added in resolveManifestPolicyConfig
#6 — Registry requirement not documented | ✅ Fixed | README Requirements section added
#7 — _ = ctx discarded | ✅ Fixed | Used for startup log.Infof
#8 — CacheKey() crypto in pkg/model | ✅ Fixed | Moved to metadataCacheKey() in manifestloader
#9 — Missing non-200 test in LookupRegistry | ✅ Fixed | t.Run("http error response") added
#10 — Closer panic | Noted | Consistent with existing codebase


<hr>
<h3>New Code Assessment</h3>
<h4><code>pkg/manifest</code> — Clean and well-placed</h4>
<p>The extraction is exactly right. The package is small, purposeful, and has no business logic beyond schema definition and validation. The constants (<code>NetworkManifestType</code>, <code>PolicyTypeRego</code>, <code>PolicySourceBundle</code>, <code>PolicySourceFile</code>) remove the magic-string duplication between the old private <code>opapolicychecker</code> code. Tests live in the same package and cover all validation branches. This is the correct pattern, matching <code>pkg/security/artifactverifier</code>.</p>
<p>One small note: <code>pkg/manifest/network_manifest_test.go</code> is in <code>package manifest</code> (white-box). Since <code>ValidateNetworkManifest</code> and <code>ParseNetworkManifest</code> are both exported, a <code>package manifest_test</code> would be cleaner and enforce the public API boundary. Not blocking.</p>
<h4><code>pkg/security/artifactverifier</code> — <code>extractPublicKeyField</code> is a solid improvement</h4>
<p>The replacement of the recursive <code>findStringField</code> with struct-based JSON decoding into <code>publicKeyResponse</code> is the right fix. The logic is now explicit about which paths are accepted:</p>
<ul>
<li><code>data.details.publicKey</code> (DeDi primary)</li>
<li><code>data.details.signing_public_key</code> / <code>data.details.public_key</code></li>
<li>Top-level <code>signing_public_key</code>, <code>public_key</code>, <code>publicKey</code></li>
</ul>
<p>The test <code>TestParsePublicKeyResponse_IgnoresUnexpectedNestedKeys</code> correctly verifies the fix.</p>
<p>One remaining concern: <strong>the Ed25519 length heuristic is still present</strong>:</p>
<pre><code class="language-go">// artifactverifier.go:3144
if len(decoded) == ed25519.PublicKeySize {
    return ed25519.PublicKey(decoded), nil
}
</code></pre>
<p>This is only reached after <code>parsePublicKey(decoded)</code> fails (DER/PEM parsing), and after <code>extractPublicKeyField</code> has already constrained which JSON paths are used. In practice this is low risk — the key material comes from a trust-anchored URL, not from arbitrary attacker input. But it remains a code smell: any valid 32-byte PKIX parse failure silently becomes an Ed25519 key. If this path is intentionally needed for a known DeDi response format, add a comment saying so. If it isn't actually exercised by any test, consider removing it.</p>
<h4><code>opapolicychecker/enforcer.go</code> — Uses <code>pkg/manifest</code> correctly</h4>
<p>The import alias <code>manifestpkg</code> is necessary to avoid shadowing the local <code>manifest</code> variable. The use of package constants instead of string literals is good. The HTTP warning messages are accurate and actionable for operators.</p>
<p>One minor point: the <code>"manifest"</code> string still appears as a literal in switch cases and string comparisons in <code>enforcer.go</code> (<code>case "manifest"</code>, <code>policyConfig.Type == "manifest"</code>, <code>policy.sourceType == "manifest"</code>). Consider defining a constant in <code>pkg/manifest</code> (e.g., <code>PolicyTypeManifest = "manifest"</code>) and using it here, for consistency with how <code>PolicySourceBundle</code> / <code>PolicySourceFile</code> are now used. Not blocking.</p>
<h4><code>manifestloader.go</code> — <code>metadataCacheKey</code> correctly moved</h4>
<p>The SHA-256 cache key computation moved from <code>model.ManifestMetadata.CacheKey()</code> to a private <code>metadataCacheKey()</code> in <code>manifestloader</code>. This is the right layering. The function is now a plain package-level function rather than a method, which is cleaner given that it's an internal implementation detail.</p>
<hr>
<h3>Remaining Minor Issues</h3>
<ol>
<li>
<p><strong>Ed25519 length fallback</strong> — Should be removed or documented. Low risk given the restricted key source paths, but unexplained in the code.</p>
</li>
<li>
<p><strong><code>PolicyTypeManifest</code> constant missing</strong> — <code>"manifest"</code> appears 4 times as a string literal in <code>enforcer.go</code>. Define it in <code>pkg/manifest</code> alongside the other source type constants.</p>
</li>
<li>
<p><strong><code>ValidateNetworkManifest</code> nil manifest guard</strong> — The check <code>if manifest == nil</code> at line 411 is unreachable from <code>ParseNetworkManifest</code> (which returns a non-nil struct pointer on success). It's there as defensive programming, which is fine, but slightly misleading.</p>
</li>
</ol>
<hr>
<h3>Overall</h3>
<p>This update is thorough and directly responsive to the first review. The <code>pkg/manifest</code> introduction is the most structurally significant change and it's done right. The <code>artifactverifier</code> refactor correctly eliminates the recursive traversal. The code is in a mergeable state — the two remaining items above are minor and can be addressed in a follow-up if needed.</p></body></html>

**Comment by nirmay** [comment]
Some more thoughts on restructuring some code for better reusability:
---

Good work overall. Three things worth addressing before merge:

**1. Eliminate `pkg/manifest/` — move everything to `pkg/model/manifest.go`**

`manifestloader` is a foundational plugin in the same way `registry` is — it fetches and returns raw `ManifestDocument` bytes, and multiple consumer plugins will parse and validate the content independently. Given that, the `NetworkManifest` schema types and validation logic need a shared home that every consumer already imports without any new dependency.

`pkg/model/manifest.go` is the right place:
- Every plugin already imports `pkg/model`
- `ManifestDocument` (the envelope) is already there — `NetworkManifest` (the content schema) should sit alongside it
- `Validate()` as a method on `NetworkManifest` matches the existing pattern of `ContextKey.UnmarshalYAML` and `Role.UnmarshalYAML` in the same file
- `ParseNetworkManifest` is a constructor for `NetworkManifest` — it belongs in the same package as the type, same as `time.Parse` returning `time.Time`

The only cost is adding `gopkg.in/yaml.v3` as an explicit import to `pkg/model`. It is already in the module's `go.mod`, so this is not a new dependency — just a new import site. Add a brief comment at the top of `manifest.go` to make the intent visible.

`pkg/manifest/` can then be deleted entirely.

**2. `dediregistry.go` — extract a shared HTTP helper**

`Lookup` and `LookupRegistry` share ten identical steps before their response parsing diverges: build request, set context, execute, handle non-200, read body, unmarshal JSON, extract the `data` field. Extract a private `fetchDeDiData(ctx, url, operation string) (map[string]interface{}, error)` method. Each public method then reads as three clear stages — validate inputs, build URL, parse response — with the HTTP plumbing invisible. This also fixes a subtle ordering issue in `LookupRegistry` where `io.ReadAll` is currently called after the status check on the error path.

**3. Minor: define `PolicyTypeManifest` constant**

The string `"manifest"` appears as a literal four times in `opapolicychecker/enforcer.go` (switch cases, type comparisons). Add `PolicyTypeManifest = "manifest"` to `pkg/model/manifest.go` alongside `PolicySourceBundle` and `PolicySourceFile` for consistency.

**Comment by nirmalnr** [comment]
**1. Eliminate pkg/manifest/ — move everything to pkg/model/manifest.go**

Revised after further discussion. The initial fix moved the types to `pkg/manifest/`. On reflection, `pkg/manifest/` as a standalone package was premature — every plugin already imports `pkg/model`, and `ManifestMetadata` / `ManifestDocument` were already there as the transport layer. All manifest-related types now live in `pkg/model/manifest.go`:

- `ManifestMetadata`, `ManifestDocument` — moved from `model.go` into `manifest.go`
- `NetworkManifest` and all schema types — moved from `pkg/manifest/`
- `ParseNetworkManifest` — package-level constructor, same as `time.Parse` returning `time.Time`
- `ValidateNetworkManifest` — converted to a method `(m *NetworkManifest) Validate(...)`, consistent with `Role.UnmarshalYAML` and `ContextKey.UnmarshalYAML` in the same package

`pkg/manifest/` is deleted entirely. `gopkg.in/yaml.v3` is added as an import to `pkg/model` — it was already in `go.mod`, not a new module dependency.

**2. dediregistry.go — extract a shared HTTP helper**

**Fixed** - extracted a private `fetchDeDiData(ctx, url, operation string) (map[string]any, error)` method that handles the shared steps: create request, set context, execute, read body, check status, unmarshal JSON, and extract the `data` field. Each public method now reads as three clear stages — validate inputs, build URL, parse response.

This also fixed a subtle bug in both methods where `io.ReadAll` was called in two separate code paths — once inside the non-200 block with the error discarded (`body, _ :=`), and again on the success path. The refactored method always reads the body first before checking the status, so `io.ReadAll` is called exactly once, its error is always propagated, and the body is available for error logging regardless of status code.

**3. Minor: define PolicyTypeManifest constant**

The review suggested placing `PolicyTypeManifest = "manifest"` in `pkg/model/manifest.go` alongside `PolicySourceBundle` and `PolicySourceFile`. The location is wrong — `PolicySourceBundle` and `PolicySourceFile` are values for fields *inside the network manifest YAML schema* (`policies.source`), whereas `PolicyConfig.Type` is the OPA plugin's own operator config. Different domains.

All four OPA policy type strings are internal to the `opapolicychecker` package, so the constants are defined there as private package-level constants in `enforcer.go`:

```go
const (
    policyTypeFile     = "file"
    policyTypeDir      = "dir"
    policyTypeBundle   = "bundle"
    policyTypeManifest = "manifest"
)
```

All literal occurrences across the switch cases and equality checks in `enforcer.go` are replaced. Error message display strings (`"type=manifest"` in `fmt.Errorf`) were also updated to use constants. Even though this was an existing pattern in this plugin, it was updated.

nirmalnr:
As discussed, will also be adding a switch to enable/disable signature verification at the `manifestloader` plugin so that testing manifests without signing is easier.

nirmalnr:
A new config `skipSignatureVerification` has been added to `manifestloader` so that signature verification can be skipped.

nirmay:
## PR #688 — Updated Review

### What's been addressed since last review

Every point from the previous review rounds has been acted on:

- ✅ `pkg/manifest/` eliminated — `pkg/model/manifest.go` created with types, constants, `ParseNetworkManifest`, and `Validate()` as a method. The yaml import comment is present as recommended.
- ✅ `dediregistry.go` refactored — `fetchDeDiData` helper extracted cleanly. `map[string]interface{}` modernised to `map[string]any` throughout.
- ✅ `artifactverifier` security fix — recursive `findStringField` traversal replaced with typed struct unmarshaling (`extractPublicKeyField`) and `extractTopLevelStringField` restricted to top-level keys only. This directly closes the concern raised in round one.
- ✅ Cleartext HTTP warning added in `opapolicychecker` for unsigned file and bundle policies sourced from `http://` URLs.
- ✅ `ManifestMetadata.CacheKey()` removed from model — cache key logic is now a private function in `manifestloader` where it belongs.
- ✅ `SkipSignatureVerification` is well-designed: warns loudly at init, sets `Verified: false` on stored documents, conditional in `loadFromCache`, passes `skipSig` through to `metadataFromRegistry` and `validateMetadata`, tested in `manifestloader_test.go`, documented in README.
- ✅ `opapolicychecker` correctly uses `model.ParseNetworkManifest`, `manifest.Validate()`, and `model.PolicySourceBundle`/`model.PolicySourceFile` constants in `resolveManifestPolicyConfig`.

---

### Remaining issues

**1. Ed25519 length heuristic in `artifactverifier` — still present**

```go
if len(decoded) == ed25519.PublicKeySize {
    return ed25519.PublicKey(decoded), nil
}
```

Raised in round one, not yet addressed. Any 32-byte value that fails `x509.ParsePKIXPublicKey` is accepted as an Ed25519 key purely by length. Real-world risk is low since the key endpoint is NFO-controlled, but it's still a loose heuristic. Consider gating this on an explicit `keyType: "ed25519"` signal in the response, or at minimum adding a comment documenting why the heuristic is acceptable here.

It is agreed that this will be fixed subsequently.

---

### Minor observations

**`RegistryMetadata` placement** — it's in `model.go` rather than `manifest.go`. This is correct: `RegistryMetadata` is a general registry concept used in `definition/registry.go` for the `RegistryMetadataLookup` interface, not specific to the manifest pipeline.

**`loadFromCache` + `SkipSignatureVerification` interaction** — if a process runs with `skipSignatureVerification: true`, stores an unverified document in Redis, then restarts with `skipSignatureVerification: false`, the cached document will be rejected and a fresh verified fetch will occur. This is the right behaviour and the README documents it.

**`GetByNetworkID` double-store on partial failure** — if `getByMetadata` stores under the metadata hash key successfully but the subsequent `store` under the network key fails, the next `GetByNetworkID` call will miss the network cache, re-fetch registry metadata, hit the metadata cache, and retry the network key store. No correctness issue, just an extra registry round-trip. Not worth changing — the failure mode is safe.

---

### Summary

The architecture is clean, the three structural recommendations from earlier reviews have all been implemented correctly, the `SkipSignatureVerification` addition is well-executed, and the `artifactverifier` security fix is solid. The remaining items above are all minor — the Ed25519 heuristic is the only one with any security framing, and even that is low risk given the trust model. This PR is in good shape to merge once those are addressed.

nirmay (approved):
Approved.

PR #691, opened 2026-04-27
 fix(keymanager): synchronize config keyset storage with public API identifiers

### LINKED ISSUE #672: simplekeymanager stores config keysets under a different key than its public API
## Problem

When `simplekeymanager` loads a keyset from configuration, it stores the keyset under `networkParticipant`, but the public `Keyset(ctx, keyID)` API, `InsertKeyset`, and `DeleteKeyset` all operate on `keyID`. That means config-loaded key material is indexed differently from the rest of the key manager API.

## Affected code
- `pkg/plugin/implementation/simplekeymanager/simplekeymanager.go:219-241`
- `pkg/plugin/implementation/simplekeymanager/simplekeymanager.go:322-338`
- `pkg/plugin/implementation/simplekeymanager/simplekeymanager_test.go:392-418`

## Why this matters
This creates an inconsistent storage model inside the key manager:
- config-loaded keysets are stored under the participant ID
- API callers retrieve and delete by key ID

That inconsistency makes the in-memory keyset map harder to reason about and risks lookup failures or duplicate entries if a deployment mixes config-loaded and programmatic keysets.

It also makes the behavior easy to misuse in future maintenance, because the storage key does not match the public API contract.

## Expected behavior
Config-loaded keysets should be stored under the same key that the public API uses for retrieval and deletion, or the API should explicitly document that the participant ID is the map key and keep that behavior consistent everywhere.

## Suggested follow-up
- Store the loaded keyset under `keyId` instead of `networkParticipant`
- Update the log message to reflect the actual storage key
- Add a regression test that verifies `Keyset(ctx, keyId)` can retrieve a config-loaded keyset

SurbhiAgarwal1:
This PR fixes an inconsistency in how keysets are stored when they are loaded from configuration versus how they are accessed through the public API.

**What was happening**

In `loadKeysFromConfig`, keysets were being stored using `networkParticipant` as the key. However, the public APIs like `Keyset`, `InsertKeyset`, and `DeleteKeyset` all work with `keyId`.

Because of this mismatch:
- Keysets loaded from config could not be reliably retrieved using the public API
- It introduced confusion around which key is actually used internally
- There was a risk of duplicate or unreachable entries when mixing config-based and programmatic usage

**What’s changed** 

- Updated the storage logic to use `keyId` when loading keysets from configuration
- Adjusted log messages to reflect the correct key being used
- Kept the behavior consistent with the public API

 **Testing** 

- Added a regression test to confirm that config-loaded keysets can be retrieved using `Keyset(ctx, keyId)`
- Verified that existing tests continue to pass

 **Result**

This keeps the internal storage aligned with the public API, making the behavior more predictable and easier to maintain.

fixed #672

### DIFF
--- a/pkg/plugin/implementation/simplekeymanager/simplekeymanager.go
+++ b/pkg/plugin/implementation/simplekeymanager/simplekeymanager.go
@@ -333,9 +333,14 @@ func (skm *SimpleKeyMgr) loadKeysFromConfig(ctx context.Context, cfg *Config) er
 			EncrPublic:     encodeBase64(encrPublic),
 		}
 
-		// Store the keyset using the keyID
+		// Store the keyset using both the networkParticipant (SubscriberID)
+		// and the KeyID (UniqueKeyID) to ensure it can be retrieved by either
+		// through the public API or internal signing steps.
 		skm.keysets[networkParticipant] = keyset
-		log.Infof(ctx, "Successfully loaded keyset from configuration with keyID: %s", keyId)
+		if keyId != networkParticipant {
+			skm.keysets[keyId] = keyset
+		}
+		log.Infof(ctx, "Successfully loaded keyset from configuration for SubscriberID: %s and KeyID: %s", networkParticipant, keyId)
 	} else {
 		log.Debug(ctx, "No keys found in configuration, keyset storage will be empty initially")
 	}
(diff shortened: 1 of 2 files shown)


nirmay:
Closing this PR as per the comments on #672 @SurbhiAgarwal1

Issue #696, opened 2026-04-28
Security: Resolve Critical/High Security Vulnerabilities in Adapter Image Dependencies


ashishkgGoogle:
Hello team,

Security analysis of a recently built Docker image for the `adapter` component revealed multiple vulnerabilities stemming from outdated dependencies, including **1 CRITICAL, 12 HIGH, and 7 MEDIUM** severity issues. These vulnerabilities may expose users to security risks.

We request that the dependencies for the `adapter` component be updated to resolve these issues. Based on the analysis, this requires updating the Go version, Go module dependencies, and the base Docker image.

---

### Recommended Actions

Fixes are available for all identified vulnerabilities. The following changes are recommended to resolve them:

#### 1. Upgrade Go Version to 1.26.1 or later

This is required to fix vulnerabilities in `stdlib` and `toolchain` (e.g., CVE-2026-27143, CVE-2026-32289, CVE-2025-4674).

*   **In `go.mod`:** Update the Go version directive:
    ```diff
    - go 1.24
    + go 1.26.1
    ```
*   **In `Dockerfile`:** Update the Go builder image:
    ```diff
    - FROM golang:1.24-bullseye AS builder
    + FROM golang:1.26.1-bookworm AS builder
    ```

#### 2. Update Go Module Dependencies

This is required to fix vulnerabilities in OpenTelemetry modules (CVE-2026-29181, CVE-2026-39883).

*   **In the `adapter` source directory:** Run the following commands to update all modules to their latest compatible versions and clean `go.mod`:
    ```bash
    go get -u ./...
    go mod tidy
    ```
*   After running these commands, verify that `go.mod` includes `go.opentelemetry.io/otel` v1.41.0 or later, and `go.opentelemetry.io/otel/sdk` v1.43.0 or later.

#### 3. Update Docker Base Image

This is required to fix vulnerabilities in OS-level packages such as `glibc` (CVE-2026-4437, CVE-2026-0915), `zlib` (CVE-2026-27171), and `busybox` (CVE-2024-58251).

*   **In `Dockerfile`:** Update the runtime base image to use the `:latest` tag to ensure the most recent patched version is used:
    ```diff
    - FROM cgr.dev/chainguard/wolfi-base
    + FROM cgr.dev/chainguard/wolfi-base:latest
    ```

---

### Vulnerability Details

| CVE | Severity | Affected Package | Details |
|---|---|---|---|
[...]

PR #700, opened 2026-04-30
fix(schemav2validator): read native Const field for OpenAPI 3.1 action extraction

### LINKED ISSUE #701: Schema-validator gives errors after openkin 1.37 update
These errors are for most actions.. and give `BAD Request: unsupported action: on_status"` etc.

**Root cause by Claude**:
kin-openapi v0.137 adds a first-class Const any field on Schema (it didn't exist in v0.133). Old code reaches into
  Extensions["const"], which only worked because v0.133 dumped unrecognized 3.1 keywords there. v0.137 now parses const into the typed field, so the Extensions lookup returns nil.

**Fix**: 
Patch getActionValue in beckn-onix's schemav2validator to read the native .Const field (with the Extensions fallback for safety):

Tackled in https://github.com/beckn/beckn-onix/pull/700

ameetdesh:
Closes #701 

## Summary
- After #696 bumped kin-openapi v0.133 → v0.137, OpenAPI 3.1 `const` is now parsed into the typed `Schema.Const` field instead of landing in `Extensions`. The action extractor in `schemav2validator` only read `Extensions["const"]`, so every action declared with `const:` (select, init, confirm, status, on_status, …) silently failed to index. Live BAP/BPP traffic NACKed with `BAD Request: unsupported action: <action>`. Only `enum:`-declared actions still worked.
- Read `Schema.Const` first; keep `Extensions["const"]` as a fallback so older kin-openapi versions still work.

## Repro
The existing `TestValidate_ActionExtraction/valid_search_action` test (which uses `const: search` against an `openapi: 3.1.0` doc) reproduces the regression on `main`:

```
$ go test ./pkg/plugin/implementation/schemav2validator/ -run TestValidate_ActionExtraction/valid_search_action -v
--- FAIL: TestValidate_ActionExtraction/valid_search_action
    schemav2validator_test.go:144: Validate() error = unsupported action: search, wantErr false
```

With this change the test passes.

End-to-end verified by rebuilding the DEG demand-flex devkit's onix-adapter image against this branch and running its Arazzo workflows — `select` / `init` / `confirm` / `update` / `status` / `on_status` all index and ACK; container logs show `25 actions indexed` (was `3`).

## Test plan
- [x] `go test ./pkg/plugin/implementation/schemav2validator/...` passes
- [x] Confirmed the existing test reproduces the regression on `main` and is green on this branch
- [x] DEG demand-flex Arazzo flows return ACK end-to-end against an image built from this branch

### DIFF
--- a/pkg/plugin/implementation/schemav2validator/schemav2validator.go
+++ b/pkg/plugin/implementation/schemav2validator/schemav2validator.go
@@ -384,7 +384,11 @@ func (v *schemav2Validator) extractActionFromSchema(schema *openapi3.Schema) str
 // getActionValue extracts action value from context schema.
 func (v *schemav2Validator) getActionValue(contextSchema *openapi3.Schema) string {
 	if actionProp := contextSchema.Properties["action"]; actionProp != nil && actionProp.Value != nil {
-		// Check const field
+		// Native OpenAPI 3.1 const (kin-openapi >= v0.137 parses this into the typed field).
+		if action, ok := actionProp.Value.Const.(string); ok && action != "" {
+			return action
+		}
+		// Fallback: older kin-openapi versions surfaced const via Extensions.
 		if constVal, ok := actionProp.Value.Extensions["const"]; ok {
 			if action, ok := constVal.(string); ok {
 				return action


nirmay:
## PR #700 Review — `fix(schemav2validator): read native Const field for OpenAPI 3.1 action extraction`

### Overview

After PR #696 bumped `kin-openapi` from v0.133 → v0.137, the library changed how it surfaces OpenAPI 3.1 `const` values: they now land in the typed `Schema.Const` field instead of `Extensions["const"]`. The action extractor in `schemav2validator` only read the `Extensions` path, so any schema using `const:` to declare an action silently failed to index — causing live traffic to NACK with `unsupported action: <action>`. This PR fixes that by reading `Schema.Const` first, with `Extensions["const"]` as a fallback.

---

### Code Quality & Correctness

**The fix is correct and minimal (+5/-1 lines).**

```go
// Native OpenAPI 3.1 const (kin-openapi >= v0.137 parses this into the typed field).
if action, ok := actionProp.Value.Const.(string); ok && action != "" {
    return action
}
// Fallback: older kin-openapi versions surfaced const via Extensions.
if constVal, ok := actionProp.Value.Extensions["const"]; ok {
```

- The type assertion `.(string)` with `ok` and the `action != ""` guard is the right pattern here — same defensive style used in the existing `enum` path.
- Priority order is correct: `Schema.Const` (the canonical path in v0.137+) before the legacy `Extensions` fallback.
- The comments are justified — they document a non-obvious kin-openapi version-dependent behavior difference, which is exactly the kind of comment worth keeping.

---

### Suggestions / Minor Issues

**1. The `Extensions["const"]` fallback is currently dead code (minor)**
`go.mod` pins `kin-openapi v0.137.0`. No reachable code path will populate `Extensions["const"]` for `const` fields under v0.137+. The fallback is still a reasonable safety net, but you could add a note like `// dead under v0.137+ but retained for accidental downgrades` to make it clear it's intentional, not forgotten code. Alternatively, if you don't want dead code, you could remove it — but the current approach of keeping it is the safer, more conservative choice.

---

### Test Coverage

The existing `TestValidate_ActionExtraction/valid_search_action` test:
- Uses `const: search` against an `openapi: 3.1.0` spec — directly exercises the regression.
- The PR author confirms it fails on `main` and passes on this branch.

The `valid_select_action` test exercises the `enum` path via `allOf`, which is unaffected.

No new tests are required — the existing regression test is sufficient and was already in place.

---

### Verdict

**Approve.** This is a clean, well-scoped regression fix with a good PR description, existing test coverage, and end-to-end verification. The only note worth acting on is clarifying the intent of the `Extensions` fallback (dead code or not), and that's a bikeshed-level concern — not a blocker.

PR #703, opened 2026-05-04
fix(config): update registry URL and key manager configuration for example configs

### LINKED ISSUE #702: Registry URLs in example config does not have nfh domain
The registry URLs in the example config currently have beckn domain. The current registry URL is https://fabric.nfh.global/registry/dedi and this needs to be updated in the example configs.

nirmalnr:
**Registry URL update**
- Example config YAMLs contain `https://api.dev.beckn.io/registry/dedi` as the registry URL.
- This has been updated to use `https://fabric.nfh.global/registry/dedi` instead.

**Key Manager Config update**
- Example config YAMLs contain `ev-charging.sandbox1.com` and `ev-charging.sandbox2.com` as the network participants. 
- This has been updated to use the [generic example BAP and BPP](https://explore.dedi.global/records/76EU7wu5EJGPXGeM4QxyWf8YEU9N1wYnNmGrEUZMK33PJT3uGRQwK4/example-NPs) (`bap.example.com` and `bpp.example.com`) instead.

Fixes #702

### DIFF
--- a/config/local-beckn-one-bap.yaml
+++ b/config/local-beckn-one-bap.yaml
@@ -51,7 +51,7 @@ modules:
         registry:
           id: dediregistry
           config:
-            url: https://api.dev.beckn.io/registry/dedi # This is the testnet URL. The production URL is https://api.beckn.io/registry/dedi
+            url: https://fabric.nfh.global/registry/dedi # This is the registry URL.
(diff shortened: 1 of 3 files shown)


nirmay (approved):
approved.

PR #704, opened 2026-05-05
Feat/local schema support

### LINKED ISSUE #692: Support canonical $ref resolution with local filesystem mapping in ONIX
Context

We are standardising on canonical schema URIs, e.g.:

$ref: "https://schema.beckn.io/hiring-common/JobLocation/v2.1/attributes.yaml#/components/schemas/JobLocation"

This keeps schemas portable, reusable, and registry-ready.

⸻

Problem

During local dev / CI / high-volume deployments:
	•	Schemas are available locally, not always hosted
	•	Current approach forces:
	•	brittle relative paths (../../../...) ❌
	•	or HTTP fetches ❌ (slow, unnecessary)

⸻

Proposal

Add a URI resolver layer in the Go SDK:
	•	Map canonical URLs → local filesystem (when configured)
	•	Fallback to HTTP if not available locally

Example:

https://schema.beckn.io/... → /local/schema-repo/...

Config:

BECKN_SCHEMA_REMOTE_BASE=https://schema.beckn.io/
BECKN_SCHEMA_LOCAL_BASE=/path/to/schema-repo/

⸻

Why this matters
	•	Cleaner schemas → no relative paths
	•	Future-ready → works with/without hosted registry
	•	Better DX → consistent refs across repos
	•	Performance (important):
	•	High-volume NPs can avoid repeated HTTP fetches
	•	Load schemas from local disk → lower latency, higher throughput
	•	Enables offline / controlled deployments

⸻

Expected Behavior

Environment	Resolution
Local / CI	URL → local filesystem
Production	URL → HTTP (or cached locally)

abhey-WIL:
Local Schema Support for Extended Schema Validation Summary

- Added PreloadSchemasToCache — when devTest: true, schemas are walked from config/schema/ at startup and stored in rawSchemas as TypeName/attributes.yaml keys.
- @type-first local lookup — in devTest mode, schema resolution tries TypeName/attributes.yaml directly against rawSchemas before touching the network. @context URL is used only as fallback for LRU cache and network fetch.
- Fixed namespaced $ref path resolution — rewrote extractRelativeSchemaPath to scan all path segments and take the last non-version, non-file segment as the type name. This fixes schemas with paths like common/CodedValue/v2.1/attributes.yaml that previously extracted common instead of CodedValue.
- $ref resolution in memory — when loading a schema from rawSchemas, the loader's ReadFromURIFunc is overridden to resolve $ref references from rawSchemas first, falling back to network only if not found locally.

**Lookup chain (devTest=true)**
@type → rawSchemas["TypeName/attributes.yaml"]
      → LRU cache (via @context URL)
      → Network fetch (via @context URL)
**Lookup chain (devTest=false)**
@context URL → LRU cache → Network fetch

### DIFF
--- a/pkg/plugin/implementation/schemav2validator/README.md
+++ b/pkg/plugin/implementation/schemav2validator/README.md
@@ -40,8 +40,22 @@ schemaValidator:
 | `extendedSchema_maxCacheSize` | string | No | "100" | Maximum number of cached domain schemas |
 | `extendedSchema_downloadTimeout` | string | No | "30" | Timeout for downloading domain schemas |
 | `extendedSchema_allowedDomains` | string | No | "" | Comma-separated domain whitelist (empty = all allowed) |
+| `extendedSchema_localSchemaPath` | string | No | "" | Path to local schema directory; schemas preloaded at startup, network used as fallback |
 
+### Local Schema Directory Layout
 
+When `extendedSchema_localSchemaPath` is set, schemas are preloaded from that directory at startup. Each schema must follow this structure:
+
+```
+config/
+ schema/
+  <TypeName>/
+    attributes.yaml
+  <AnotherType>/
+    attributes.yaml
+```
+
+A startup warning is logged if no schemas are found, which usually means the directory path is wrong or the layout doesn't match the expected structure.
 
 ## How It Works
 
--- a/pkg/plugin/implementation/schemav2validator/cmd/plugin.go
+++ b/pkg/plugin/implementation/schemav2validator/cmd/plugin.go
@@ -66,6 +66,9 @@ func (vp schemav2ValidatorProvider) New(ctx context.Context, config map[string]s
 		if v, ok := config["extendedSchema_allowedDomains"]; ok && v != "" {
 			cfg.ExtendedSchemaConfig.AllowedDomains = strings.Split(v, ",")
 		}
+		if v, ok := config["extendedSchema_devTest"]; ok {
+			cfg.ExtendedSchemaConfig.DevTest = v == "true"
+		}
 
 	}
 
--- a/pkg/plugin/implementation/schemav2validator/cmd/plugin.go
+++ b/pkg/plugin/implementation/schemav2validator/cmd/plugin.go
@@ -66,8 +66,8 @@ func (vp schemav2ValidatorProvider) New(ctx context.Context, config map[string]s
 		if v, ok := config["extendedSchema_allowedDomains"]; ok && v != "" {
 			cfg.ExtendedSchemaConfig.AllowedDomains = strings.Split(v, ",")
 		}
-		if v, ok := config["extendedSchema_devTest"]; ok {
-			cfg.ExtendedSchemaConfig.DevTest = v == "true"
+		if v, ok := config["extendedSchema_localSchemaPath"]; ok && v != "" {
+			cfg.ExtendedSchemaConfig.LocalSchemaPath = v
 		}
 
 	}
--- a/pkg/plugin/implementation/schemav2validator/extended_schema.go
+++ b/pkg/plugin/implementation/schemav2validator/extended_schema.go
@@ -18,9 +18,6 @@ import (
 	"github.com/getkin/kin-openapi/openapi3"
 )
 
-// localSchemaBasePath is the base directory for local schema files in dev/test mode.
-const localSchemaBasePath = "config/schema"
-
 func extractRelativeSchemaPath(u *url.URL) string {
 	const marker = "/schema/"
 	var pathPart string
@@ -36,7 +33,7 @@ func extractRelativeSchemaPath(u *url.URL) string {
 			continue
 		}
 		ext := filepath.Ext(seg)
-		if ext == ".yaml" || ext == ".yml" || ext == ".json" {
+		if ext == ".yaml" || ext == ".yml" || ext == ".json" || ext == ".jsonld" {
 			continue
 		}
 		if isSchemaVersionSegment(seg) {
@@ -73,7 +70,7 @@ type ExtendedSchemaConfig struct {
 	MaxCacheSize    int      // default 100
 	DownloadTimeout int      // seconds, default 30
 	AllowedDomains  []string // whitelist (empty = all allowed)
-	DevTest         bool     // when true, preload schemas from disk into memory at startup and serve from memory
+	LocalSchemaPath string   // when non-empty, preload schemas from this directory; network is fallback
 }
 
 // referencedObject represents a domain-specific object with @context.
@@ -114,7 +111,7 @@ func rawSchemaKey(rel string) string {
 	return filepath.ToSlash(rel)
 }
 
-func PreloadSchemasToCache(ctx context.Context, c *schemaCache, baseDir string) error {
+func preloadSchemasToCache(ctx context.Context, c *schemaCache, baseDir string) error {
 	count := 0
 	err := filepath.WalkDir(baseDir, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
@@ -132,7 +129,9 @@ func PreloadSchemasToCache(ctx context.Context, c *schemaCache, baseDir string) 
 			return fmt.Errorf("failed to compute relative path for %s: %w", path, relErr)
 		}
 		key := rawSchemaKey(rel)
+		c.mu.Lock()
 		c.rawSchemas[key] = data
+		c.mu.Unlock()
 		count++
 		log.Debugf(ctx, "Preloaded schema to memory: %s", key)
 		return nil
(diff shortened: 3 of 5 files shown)


nirmay (commented):
## Code Review

The three-tier lookup design (local memory → LRU cache → network) is the right approach and should be kept as-is. Issues below range from blocking to minor.

---

### [Blocking] Design: `DevTest bool` → `LocalSchemaPath string`

Every other `extendedSchema_*` field is a resource config (TTL, timeout, cache size, domains). A boolean mode flag is structurally inconsistent, and the name `devTest` signals "not for production" when the feature — preloading schemas locally to avoid per-request network fetches — has clear production value for high-volume deployments.

The fix is to replace the bool with a string path. Presence of the value implies local-first mode; empty string means network-only (current default).

```go
// ExtendedSchemaConfig
- DevTest         bool
+ LocalSchemaPath string // when non-empty, preload schemas from this directory; network is fallback
```

```go
// schemav2validator.go
- if config.ExtendedSchemaConfig.DevTest {
-     log.Infof(ctx, "DevTest mode: preloading schemas from %s to memory", localSchemaBasePath)
-     if err := PreloadSchemasToCache(ctx, v.schemaCache, localSchemaBasePath); err != nil {
+ if p := config.ExtendedSchemaConfig.LocalSchemaPath; p != "" {
+     log.Infof(ctx, "Local schema mode: preloading schemas from %s", p)
+     if err := preloadSchemasToCache(ctx, v.schemaCache, p); err != nil {
```

```go
// cmd/plugin.go
- if v, ok := config["extendedSchema_devTest"]; ok {
-     cfg.ExtendedSchemaConfig.DevTest = v == "true"
- }
+ if v, ok := config["extendedSchema_localSchemaPath"]; ok && v != "" {
+     cfg.ExtendedSchemaConfig.LocalSchemaPath = v
+ }
```

Drop the `localSchemaBasePath` constant entirely — replaced by the config value.

---

### [Blocking] Bug: `rawSchemaKey` produces wrong keys for domain-prefixed directory structures

[`extended_schema.go:109`]

```go
func rawSchemaKey(rel string) string {
    parts := strings.SplitN(filepath.ToSlash(rel), "/", 3)
    if len(parts) == 3 {
        return parts[0] + "/" + parts[2]  // skips parts[1]
    }
    return filepath.ToSlash(rel)
}
```

For a file at `hiring-common/JobLocation/v2.1/attributes.yaml`:
- `SplitN` (limit 3) → `["hiring-common", "JobLocation", "v2.1/attributes.yaml"]`
- Returns `"hiring-common/v2.1/attributes.yaml"` ❌

But `extractRelativeSchemaPath` on `https://schema.beckn.io/hiring-common/JobLocation/v2.1/attributes.yaml` returns `"JobLocation/attributes.yaml"`. The keys never match — all local lookups miss silently.

`rawSchemaKey` works correctly only when the local directory is structured as `TypeName/version/attributes.yaml` (type at top level, no domain prefix). This assumption needs to be either fixed or explicitly documented and enforced by a test that encodes the expected directory structure.

---

### [Blocking] No tests for any new logic

The test changes are purely mechanical signature updates. There are no tests for:

- `preloadSchemasToCache` — correct directory walk and key generation
- `extractRelativeSchemaPath` — canonical URLs with domain prefix, versioned paths, missing `/schema/` marker
- `isSchemaVersionSegment` — boundary cases: `v`, `1.0`, `v1`, `v1.2.3`, non-version strings
- `rawSchemaKey` — 2-part and 3-part split cases, and crucially whether its output matches `extractRelativeSchemaPath` for the same schema
- Local-first lookup chain end-to-end: local hit, local miss → LRU hit, local miss → network

The path extraction functions are the most fragile part of this PR and are exactly what needs table-driven unit tests.

---

### [Medium] Domain whitelist bypass must be scoped to local path only

[`extended_schema.go:482`]

```go
if !devTest && !isAllowedDomain(obj.Context, allowedDomains) {
```

When `localSchemaPath` is set and the schema is found locally, the domain check is irrelevant — no network request is made. But when the schema is **not** found locally and the code falls back to network, the `@context` URL is used to make a real outbound request. The current code bypasses the whitelist for that network fallback too, which is incorrect.

Move the check to guard the network fallback specifically:

```go
var doc *openapi3.T

if localSchemaPath != "" {
    doc, _ = c.loadSchemaFromPath(...)  // local — domain check not relevant
}

if doc == nil {
    // about to hit the network — whitelist applies
    if !isAllowedDomain(obj.Context, allowedDomains) {
        return fmt.Errorf("domain not allowed: %s", obj.Context)
    }
    schemaPath := transformContextToSchemaURL(obj.Context)
    doc, err = c.loadSchemaFromPath(ctx, schemaPath, ttl, timeout, false)
}
```

---

### [Medium] `obj.Type` used as a filesystem path without validation

[`extended_schema.go:492`]

```go
doc, _ = c.loadSchemaFromPath(ctx, obj.Type+"/attributes.yaml", ttl, timeout, devTest)
```

`obj.Type` comes directly from the JSON body's `@type` field. If it is empty, contains slashes, or is a full URL, the constructed path is malformed or potentially unsafe. Add a guard before using it as a path component:

```go
if obj.Type != "" && !strings.ContainsAny(obj.Type, "/\\:") {
    doc, _ = c.loadSchemaFromPath(...)
}
```

---

### [Medium] Silent error discard on local lookup

Same line as above — the error is discarded with `_`. If the local lookup fails for a reason other than a simple miss (e.g. a corrupt YAML file was preloaded), the failure is invisible and execution silently falls through to the network. Log at debug level:

```go
if localDoc, err := c.loadSchemaFromPath(...); err != nil {
    log.Debugf(ctx, "local @type lookup failed for %s: %v", obj.Type, err)
} else {
    doc = localDoc
}
```

---

### [Low] `PreloadSchemasToCache` exported with an unexported type

```go
func PreloadSchemasToCache(ctx context.Context, c *schemaCache, baseDir string) error {
```

`*schemaCache` is unexported, so this function cannot be called from outside the package — the export is meaningless. Unexport it: `preloadSchemasToCache`.

---

### [Low] `rawSchemas` written without holding the mutex

`preloadSchemasToCache` writes to `c.rawSchemas[key]` without acquiring `c.mu`. This is safe today only because preload runs at startup before the cache is shared — an implicit ordering constraint with no enforcement. Either acquire the write lock during the walk, or add a comment explicitly stating this must be called before the cache is in use.

---

### Minor

- Log at `Warn` (not `Info`) when `LocalSchemaPath` is configured, so it is visible in environments where `Info` is filtered.
- Add a short block comment above the three steps in `loadSchemaFromPath` — the lookup chain is well-described in the PR description but will be lost without it.

abhey-WIL:
- DevTest bool → LocalSchemaPath string — replaced bool flag with path string; empty = network-only
- The schema is needed in "JobLocation/attributes.yaml" format only as for local schema we are picking the schema names from @type, reason for picking the @type for schema names is for covering the scenario when the user does not have hosted schema so they can use the schema by storing in their local.
- Added unit tests for isSchemaVersionSegment, extractRelativeSchemaPath, rawSchemaKey, preloadSchemasToCache, and the full lookup chain
- Domain whitelist check moved inside the network fallback block — no longer bypassed when local schema is found
- obj.Type path guard added — strips JSON-LD prefix, blocks /, \ to prevent path traversal
- Silent _ error on local lookup replaced with log.Debugf so corrupt preloaded schemas are visible
- PreloadSchemasToCache → preloadSchemasToCache — unexported to match its unexported receiver type
- c.mu.Lock() added per entry in preloadSchemasToCache — guards against concurrent reads during startup preload

nirmay:
<html><head></head><body>
<hr>
<h3>Resolved ✅</h3>

Item | How
-- | --
DevTest bool → LocalSchemaPath string | Done correctly, including config key rename
PreloadSchemasToCache unexported | → preloadSchemasToCache
Mutex during preload writes | c.mu.Lock()/Unlock() added per-write in the walk
Domain whitelist scoped to network fallback | Whitelist check now inside if doc == nil block
obj.Type used as path without validation | Namespace prefix stripped (LastIndex(":")), path chars guarded
Silent error discard | Now logged at Debugf
Log level for local schema mode | Warnf ✓
Three-tier block comment | Added above the steps in loadSchemaFromPath
Tests | isSchemaVersionSegment, extractRelativeSchemaPath, rawSchemaKey, consistency test, preloadSchemasToCache, full lookup chain, validateReferencedObject local hit and fallback
.jsonld skipped in extractRelativeSchemaPath | Added to extension skip list


<hr>
<h3>Still open</h3>
<p><strong>1. <code>rawSchemas</code> reads happen without holding the lock</strong></p>
<p>Writes during preload now acquire <code>c.mu</code>, but reads in <code>loadSchemaFromPath</code> and the <code>ReadFromURIFunc</code> closure do not:</p>
<pre><code class="language-go">// no lock held here
if schemaBytes, ok := c.rawSchemas[relPath]; ok {
</code></pre>
<p>In practice it's safe because <code>rawSchemas</code> is write-once at startup before any requests are served. But since writes now hold the lock, reads should too — otherwise a race detector will flag it. Either also lock on reads, or drop the write lock (since it's single-threaded at startup) and add a comment explaining the write-once invariant.</p>
<p><strong>2. <code>baseURL</code> for in-memory loading is hardcoded to <code>schema.beckn.io</code></strong></p>
<pre><code class="language-go">baseURL := &amp;url.URL{Scheme: "https", Host: "schema.beckn.io", Path: "/" + relPath}
doc, err = loader.LoadFromDataWithPath(schemaBytes, baseURL)
</code></pre>
<p>When the loader resolves relative <code>$ref</code> paths, it does so against this base URL. That works correctly for schemas originally hosted at <code>schema.beckn.io</code>, but for <code>@context</code> values pointing to other domains (which the whitelist allows), relative <code>$ref</code> resolution will use the wrong host. For absolute <code>$ref</code> URLs, <code>extractRelativeSchemaPath</code> strips the host so it still works — but relative <code>$ref</code>s in schemas from non-beckn domains will silently resolve against the wrong base. Worth either making this configurable or documenting the assumption.</p>
<p><strong>3. <code>rawSchemaKey</code> directory structure assumption undocumented for operators</strong></p>
<p>The tests correctly encode the expectation: local schemas must be structured as <code>TypeName/[version/]attributes.yaml</code> (TypeName at top level, no domain prefix). That's the right design. But there's no operator-facing documentation — no README update, no startup log showing the expected layout, and no warning if the directory contains files that don't produce any valid keys. A silent misconfiguration (wrong directory structure) would result in all local lookups missing with no indication why.</p>
<hr>
<h3>Summary</h3>
<p>The code is in good shape — all the behavioural and design issues are resolved. The three remaining items are: a read/write lock consistency issue (low risk, but will trip the race detector), a <code>baseURL</code> assumption for multi-domain <code>$ref</code> resolution (minor correctness edge case), and missing operator documentation for the expected directory layout (usability gap).</p></body></html>

abhey-WIL:
1. Added c.mu.RLock()/RUnlock() around both rawSchemas reads in loadSchemaFromPath to match the existing write lock in preloadSchemasToCache.

2. Replaced LoadFromDataWithPath with LoadFromData, removing the hardcoded schema.beckn.io entirely — kin-openapi passes relative $ref paths as-is to ReadFromURIFunc when no base URL is set, which extractRelativeSchemaPath already handles correctly.

3. Added a Warnf when no schemas are loaded stating the expected TypeName/attributes.yaml layout, and updated the README with a Local Schema Directory Layout section.

nirmay (approved):
Approved.

Issue #690, opened 2026-04-24
Tests(opapolicychecker plugin): add network_id snake_case variant to enforcer tests and clarify get("action", "action") comment


nirmalnr:
**Test gap:** `TestEnforcer_NetworkPolicySelection` and `TestEnforcer_NetworkPolicyDefaultFallback` only construct requests with `"networkId"` (camelCase). While `parseRequestContext` handles the snake_case variant and is tested in isolation via `TestParseRequestContext_NetworkIDVariants`, there is no end-to-end enforcer test where the request body uses `"network_id"`. Adding a one-liner variant to `TestEnforcer_NetworkPolicySelection` would close this gap.

**Comment clarity:** In `parseRequestContext`, the call `get("action", "action")` passes identical strings for both the snake_case and camelCase keys, making the second lookup a no-op. This reads confusingly. A short comment — e.g. `// "action" has no camelCase variant` — would make the intent clear.

SurbhiAgarwal1:
I've looked into this. The plan is to:
1. Add a test case to `TestEnforcer_NetworkPolicySelection` (in `opapolicychecker_test.go`) that uses `"network_id"` in the request body to ensure end-to-end coverage for the snake_case variant.
2. 2. Add a clarifying comment in `parseRequestContext` for the `get("action", "action")` line.
Raising a PR for this now.

Issue #689, opened 2026-04-24
Refactor(opapolicychecker): remove redundant networkID return value from selectedPolicy


nirmalnr:
In `enforcer.go`, `selectedPolicy` currently returns `(*loadedPolicy, string)` where the second return value is `networkID`. After the recent refactor, `networkID` is extracted before the call and is already available at the call site (via `reqCtx.NetworkID`). Returning it back serves no purpose — the only downstream use is log formatting, which can use `reqCtx.NetworkID` directly.

**Suggested change:** Update the signature to return only `*loadedPolicy` and replace any log references to the returned string with `reqCtx.NetworkID`.

nirmay:
#708 is merged, closing.

PR #708, opened 2026-05-06
refactor(opapolicychecker): drop redundant networkID return from selectedPolicy, add snake_case enforcer test, clarify no-camelCase comments

nirmalnr:
## Summary

- **Refactor**: `selectedPolicy` no longer returns `networkID` as a second value — it was identical to the input and already available at the call site via `reqCtx.NetworkID`. All downstream log references updated accordingly.
- **Test**: `TestEnforcer_NetworkPolicySelection` now includes a variant with `"network_id"` (snake_case) in the request body, closing the end-to-end gap that existed despite `parseRequestContext` handling it in isolation.
- **Clarity**: Added `// "action" has no camelCase variant` and `// "timestamp" has no camelCase variant` comments in `parseRequestContext` to make the intentional duplicate-key `get()` calls self-explanatory.

Addresses issues #689 and #690

### DIFF
--- a/pkg/plugin/implementation/opapolicychecker/enforcer.go
+++ b/pkg/plugin/implementation/opapolicychecker/enforcer.go
@@ -652,19 +652,19 @@ func (e *PolicyEnforcer) reloadNetworkPolicies(ctx context.Context) {
 	log.Infof(ctx, "OPAPolicyChecker: network policies reloaded in %s (policies=%d, hasDefault=%t)", time.Since(start), len(policies), defaultPolicy != nil)
 }
 
-func (e *PolicyEnforcer) selectedPolicy(networkID string) (*loadedPolicy, string) {
+func (e *PolicyEnforcer) selectedPolicy(networkID string) *loadedPolicy {
 	e.evaluatorMu.RLock()
 	defer e.evaluatorMu.RUnlock()
 
 	if networkID != "" {
 		if policy, ok := e.policies[networkID]; ok {
-			return policy, networkID
+			return policy
 		}
 	}
 	if e.defaultPolicy != nil {
-		return e.defaultPolicy, networkID
+		return e.defaultPolicy
 	}
-	return nil, networkID
+	return nil
 }
 
 // CheckPolicy evaluates the message body against loaded OPA policies.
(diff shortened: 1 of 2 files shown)


nirmay:
### Overview

A focused cleanup PR that does three independent things:
1. Removes a redundant second return value from `selectedPolicy` (it always echoed the input)
2. Adds a missing end-to-end test for snake_case `network_id` in request context parsing
3. Adds clarifying comments for duplicate-key `get()` calls in `parseRequestContext`

All changes are confined to [`[enforcer.go](https://claude.ai/epitaxy/pkg/plugin/implementation/opapolicychecker/enforcer.go)`](pkg/plugin/implementation/opapolicychecker/enforcer.go) and its test file.

---

### Analysis

**Refactor — Drop `networkID` second return value**

- Correct and clean. The original signature `selectedPolicy(networkID string) (*loadedPolicy, string)` always returned the same `networkID` that was passed in — it added zero information. Using `reqCtx.NetworkID` directly at the call site is clearer and avoids the misleading implication that the function might transform or derive a different ID.
- All four log statements updated consistently. No behavior change.
- `selectedPolicy` is unexported, so no external callers to worry about.

**Test — snake_case `network_id` variant**

- Good gap closure. The `get("network_id", "networkId")` logic in `parseRequestContext` was tested in isolation but lacked an end-to-end path through `CheckPolicy`. The new case exercises that full path.
- The test case is appended inline to `TestEnforcer_NetworkPolicySelection` rather than as a separate table-driven entry. Minor style inconsistency but not a problem.
- One subtle thing worth verifying: the test body `{"context":{"action":"confirm","network_id":"retail.network/production"}}` — ensure that `parseRequestContext` actually extracts `NetworkID` from the `context` object correctly. If `get()` does a top-level JSON key lookup and `network_id` is nested under `context`, this test might be passing vacuously (matching the default policy, not the retail network policy). This should be confirmed.

**Comments in `parseRequestContext`**

- Rationale is sound — without the comments, `get("action", "action")` looks like a mistake or copy-paste error. The comments prevent future "fix" PRs that break things.
- Minor style note: Go convention favors comments on their own line rather than trailing end-of-line comments, especially on already-long struct literal lines. Not a blocker, but worth considering for consistency with the rest of the file.

---

### Potential Issues

- **Risk level: low.** This is a net-negative diff (more deletions than additions) with no logic changes.
- **One thing to verify:** The snake_case test case — confirm that `parseRequestContext` actually resolves `network_id` from within the `context` sub-object, not just from a top-level key. If the JSON traversal is top-level only, the test may not exercise what the PR claims. @nirmalnr can you pls verify this?

---

### Summary

**Approve with one verification request.** The refactor is correct and the intent is clear. The main thing to double-check before merging is whether the new test case in `TestEnforcer_NetworkPolicySelection` is actually routing through the retail network policy (not just hitting the default policy silently). If the `get()` function does nested JSON path resolution, it's fine — if it's top-level only, the test assertion may need adjustment.

nirmalnr:
Confirmed. `parseRequestContext` unmarshals into a struct with `Context map[string]interface{} json:"context"` and the `get()` helper only reads from that nested map — there's no top-level lookup. So `{"context":{"network_id":"retail.network/production"}}` correctly resolves `NetworkID`, which `selectedPolicy` uses to hit the retail policy (not the default), confirmed by the `"retail policy violation"` assertion. Looks good to me.

nirmay (approved):
Approved.

Issue #707, opened 2026-05-06
feat: PayloadStore — stateful payload history infrastructure plugin


nirmay:
# PayloadStore Plugin — Design Document

## Overview

`PayloadStore` is an infrastructure plugin for ONIX that records payloads as they pass through the gateway, indexed by transaction and message identifiers, with a configurable TTL-based expiration policy.

ONIX has been stateless since inception. This plugin is the foundation for all stateful use cases — it makes transaction history available to other plugins without requiring any of them to manage storage themselves.

Like `ManifestLoader`, `PayloadStore` is an **infra plugin**: it depends on `Cache`, and gets injected into business-logic plugins that need historical context. It does not implement business logic itself.

---

## Positioning in the Pipeline

`PayloadStore` is **not a step**. Cache is not a step. ManifestLoader is not a step. Infrastructure plugins are injected into the components that use them — they are not pipeline stages. Making `storePayload` a configurable step would add ordering constraints operators must get right, a configuration surface where omitting it silently breaks all history-dependent plugins, and a conceptual inconsistency: storage is infrastructure, not a validation or transformation stage.

Instead, the handler stores the payload automatically at fixed points in its own flow — once when the pipeline succeeds (ACK path) and once when it fails (NACK path). The operator only decides *whether* to use PayloadStore via `PluginCfg`, not *where* it fires.

```
HTTP Request
    │
    ├─► [dedup check]  ── Exists() ──► PayloadStore ──► definition.Cache
    │         │                        (Redis / ElastiCache /
    │     duplicate?                    Memcached / any impl)
    │         │ yes
    │         └──► NACK ──────────────────────────────► Store(Outcome: NACK)
    │
    ▼
[validateSign]
    │
[validateSchema]
    │
[checkPolicy]  ──── GetByTransactionID() ──► PayloadStore
    │                GetByMessageID()
[addRoute]
    │
    ├── success ──► forward ──► ACK ──────────────────► Store(Outcome: ACK)
    │
    └── failure ──────────────► NACK ─────────────────► Store(Outcome: NACK)
```

**Key consequence:** when `checkPolicy` queries PayloadStore for transaction history, the *current* request is not yet in the store. This is intentional and correct — policy checks look at what arrived *before* the current message. The current message is stored only after the pipeline resolves.

[...]

Issue #712, opened 2026-05-12
fix(router): TestValidateRulesFailure broken by Go stdlib URL error message change


nirmalnr:
## Description

`TestValidateRulesFailure/Invalid_URL_for_BAP_targetType` in `pkg/plugin/implementation/router/router_test.go` fails because the test asserts a hardcoded error substring that no longer matches the error produced by the current Go version.

## Root Cause

The test expects the error message to contain:
```
invalid character " " in host name
```

But Go 1.26.1's `url.Parse` now returns:
```
invalid IP-literal
```
for the URL `http:// [invalid].com`. The underlying stdlib error message changed between Go versions, causing the substring assertion to fail.

## Fix

Update the expected error substring in the test to match the current Go stdlib message, or use a more version-stable assertion (e.g. check that an error is returned at all, or assert on a higher-level message the router itself produces rather than the raw `url.Parse` error).

**Affected file:** `pkg/plugin/implementation/router/router_test.go`

Issue #711, opened 2026-05-12
fix(cache): TestCache_Get fails due to OTel context wrapping


nirmalnr:
## Description

`TestCache_Get` in `pkg/plugin/implementation/cache/cache_test.go` panics with an unexpected mock call because the test sets up the mock expecting a bare `context.Background()`, but the cache implementation wraps the context with an OpenTelemetry tracing span before passing it to Redis.

## Root Cause

`cache.go:120` enriches the context with an OTel trace before calling the Redis client. The `testify/mock` strict argument comparison sees `*context.valueCtx` (with trace metadata) instead of the plain `context.backgroundCtx` the mock was set up to expect, causing a panic.

```
Diff: 0: FAIL: (*context.valueCtx=context.Background.WithValue(trace.traceContextKeyType, ...))
           != (context.backgroundCtx=context.Background)
```

## Fix

In the test, replace the exact context argument matcher with `mock.Anything` (or `mock.MatchedBy` to check only non-trace fields) so the expectation is not broken by the OTel context enrichment.

**Affected file:** `pkg/plugin/implementation/cache/cache_test.go`

PR #716, opened 2026-05-14
feat(payloadstore): add stateful payload recording plugin

nirmalnr:
## Summary

- Adds the `payloadstore` plugin - beckn ONIX's first stateful infrastructure primitive. Records every request and response payload passing through beckn ONIX, indexed by `message_id` and `transaction_id`, with TTL-based expiration via the cache backend.
- Wires PayloadStore into the handler at two fixed points: a dedup pre-check before the step pipeline (~~NACK~~ logs a warning on duplicate `message_id`) and an outcome store after the pipeline (ACK or NACK with response body and error reason).
- Cache keys are namespaced by beckn ONIX module name (`payload:{moduleName}:msg:{id}`) to prevent collisions between handlers sharing the same Redis backend.
- Serialized entries use a format prefix (`j:` for plain JSON, `c:` for gzip+base64) so the compression setting can be changed between restarts without breaking existing entries.
- Duplicate detections do ~~not~~ overwrite the original entry — the ~~legitimate~~ latest message is preserved.

### New plugin
- `pkg/plugin/definition/payloadstore.go` — `PayloadStore` interface, `PayloadEntry` struct, `Outcome` type, `PayloadStoreProvider` interface
- `pkg/plugin/implementation/payloadstore/` — full plugin implementation: config parsing, cache key helpers, serializer, store logic, plugin entrypoint, tests, and README
- `install/build-plugins.sh` — added `payloadstore` to the build list

### Handler wiring
- `core/module/handler/config.go` — added `PayloadStore` to `PluginCfg` and `PluginManager` interface
- `core/module/handler/http_metric.go` — extended `responseRecorder` with a `bytes.Buffer` to capture the response body for storage
- `core/module/handler/stdHandler.go` — added dedup pre-check, outcome store, `loadPayloadStore` helper, `buildPayloadEntry` helper

### Plugin manager
- `pkg/plugin/manager.go` — added `PayloadStore` method following the `ManifestLoader` pattern

### Example configs
- `config/local-beckn-one-bap.yaml` — added `payloadStore` to both BAP modules; corrected `uuidKeys` → `contextKeys` in `reqpreprocessor`
- `config/local-beckn-one-bpp.yaml` — added `payloadStore` to both BPP modules; added missing `reqpreprocessor` middleware to `bppTxnReceiver`

## Test plan

- [x] `go test ./pkg/plugin/implementation/payloadstore/...` — unit tests for config, serializer, and store logic
- [x] `go test ./core/module/handler/...` — handler tests including updated mocks
- [x] `go test ./...` — full suite (two pre-existing failures unrelated to this PR: cache OTel context mismatch #711, router test broken by Go 1.26 url.Parse change #712)
- [x] Send a request through a handler with `payloadStore` configured → verify `payload:onix:msg:{id}` and `payload:onix:txn:{txnId}:index` appear in Redis
- [x] ~~Send the same `message_id` again → verify NACK with `"duplicate message_id"` and original entry is unchanged in Redis~~
- [x] Toggle `compress` between restarts → verify entries written under either setting still deserialize correctly


config/local-beckn-one-bpp.yaml
```diff
@@ -90,6 +99,11 @@ modules:
           config:
             networkPolicyConfig: ./config/opa-network-policies.yaml
             refreshInterval: "5m"
+        middleware:
+          - id: reqpreprocessor
```
nirmay, line 103: @nirmalnr Is this change really needed? pls remove if not.

nirmalnr: I was making the plugin not require this middleware but I think it is still used by otel plugin etc. Could not understand why this is not present only in the example BPP config? It is present already in the BAP config.

nirmalnr:
Namespacing based on handler name has been removed as discussed.

nirmay (approved):
Approved.

nirmalnr:
Fixes #707

PR #721, opened 2026-05-19
fix(tests): resolve flaky/failing unit tests across cache, router, and otelsetup packages

### LINKED ISSUE #711: fix(cache): TestCache_Get fails due to OTel context wrapping

### LINKED ISSUE #712: fix(router): TestValidateRulesFailure broken by Go stdlib URL error message change

nirmalnr:
### Fixes

- Closes #711
- Closes #712

#### `pkg/plugin/implementation/cache/cache_test.go` (fixes #711)

`TestCache_Get`, `TestCache_Set`, `TestCache_Delete`, and `TestCache_Clear` all panicked
because the cache implementation wraps the context with an OpenTelemetry span before
passing it to Redis. The strict `testify/mock` argument matcher saw `*context.valueCtx`
instead of the expected `context.backgroundCtx`, causing a panic.

**Fix:** replaced exact `ctx` argument with `mock.Anything` in all four mock registrations.

#### `pkg/plugin/implementation/router/router_test.go` (fixes #712)

`TestValidateRulesFailure/Invalid_URL_for_BAP_targetType` (and the equivalent BPP subtest)
asserted on a raw `url.Parse` error substring (`invalid character " " in host name`) that
changed in Go 1.26 (`invalid IP-literal`).

**Fix:** trimmed both `wantErr` strings to the router-produced prefix only, which is
stable across Go versions. The test loop uses `strings.Contains`, so the shorter prefix
is sufficient.

#### `pkg/plugin/implementation/otelsetup/cmd/plugin_test.go`

Multiple tests set `enableMetrics: "true"` without an `otlpEndpoint`, causing the
5-second shutdown context in the cleanup closure to time out while trying to flush metrics
to an unreachable gRPC endpoint.

**Fix:** set `enableMetrics: "false"` in test cases whose purpose is to verify
initialisation or config parsing (not OTLP export), and dropped the `assert.NoError` on
cleanup in `TestMetricsProviderNew_BooleanParsing` (which tests `ParseBool` behaviour,
not flush behaviour).

### Test plan

- [x] `go test ./pkg/plugin/implementation/cache/... -v` — all cache tests pass
- [x] `go test ./pkg/plugin/implementation/router/... -run TestValidateRulesFailure -v` — all subtests pass
- [x] `go test ./pkg/plugin/implementation/otelsetup/... -v` — all 14 tests across both sub-packages pass


nirmay:
TestMetricsProviderNew_ConfigConversion — Consider adding a separate sub-case with enableMetrics: "true" that skips the cleanup or uses a mock OTLP endpoint, so the true bool conversion path remains covered.

@nirmalnr

nirmalnr:
True bool conversion path is covered already in `TestMetricsProviderNew_BooleanParsing`. This PR was raised to fix some already broken tests which were surfaced in an unrelated PR. Intention of this PR was to fix the issues within the existing scope. I suggest if test coverage is an issue we can create a separate ticket addressing the same. Let me know what you think.

nirmay:
Approving — the fix is scoped correctly and the `BooleanParsing` test does cover the `"true"→true` path.

Two minor items worth tracking as follow-ups:

1. **`TestMetricsProviderNew_BooleanParsing` is still slow** — the 4 cases that default to `enableMetrics: true` (`"true"`, `"TRUE"`, `"invalid"`, `""`) each block ~5s waiting for the OTLP flush to time out. That's ~20s added to the test suite on every run. Silencing the error was the right call to unblock CI, but a follow-up should either point these at a dummy endpoint or wrap cleanup in a short-deadline context.

2. **`TestMetricsProviderNew_ConfigConversion` "All fields provided" no longer exercises `enableMetrics: "true"` through the full `provider.New()` round-trip** — minor, since `BooleanParsing` covers the parse, but the config-conversion path for an explicit `"true"` input is untested at the integration level.

Suggest one ticket to address both.

nirmay (approved):
Approved.

PR #723, opened 2026-05-19
fix(cache): cache.Get returns ("", nil) on miss instead of leaking redis.Nil

### LINKED ISSUE #717: fix(cache): cache.Get should return ("", nil) on key not found instead of propagating redis.Nil
## Problem

The Redis cache plugin's `Get` implementation currently leaks `redis.Nil` — go-redis's sentinel error for key not found — as a non-nil error to callers. This violates the implicit contract of `definition.Cache.Get`, which should return `("", nil)` on a miss and only return an error for genuine failures (connection down, timeout, auth failure).

## Impact

PayloadStore's `Exists` method is forced to treat all cache errors as a miss (fail-open), because it cannot distinguish `redis.Nil` (normal miss) from a real Redis failure without importing go-redis — which would break the cache backend independence of the plugin. This means `Exists` will never surface a real cache error to the caller.

Without this fix, every first-time message lookup logs a spurious warning:
```
payloadStore.Exists: cache error — duplicate detection skipped: redis: nil
```

## Fix

In the Redis cache plugin's `Get`:

```go
val, err := c.client.Get(ctx, key).Result()
if errors.Is(err, redis.Nil) {
    return "", nil // key not found — normal miss, not an error
}
if err != nil {
    return "", err // genuine Redis failure
}
return val, nil
```

## Follow-up

Impact analysis across call sites that use the cache plugin.

nirmalnr:
## Summary

Closes #717

- `cache.Get` now absorbs `redis.Nil` (go-redis key-not-found sentinel) and returns `("", nil)`, honouring the implicit contract of `definition.Cache.Get`
- `payloadstore.Exists` drops the fail-open workaround and propagates real cache errors now that misses are clean
- `TestCache_Get` expanded to a table-driven test covering hit, miss, and genuine error cases
- Fixed pre-existing broken mock expectations in `TestCache_Set`, `TestCache_Delete`, `TestCache_Clear` (bare `ctx` used instead of `mock.Anything` for the OTel span-wrapped context)

## Impact on other call sites

All other `cache.Get` callers use `if err != nil || raw == ""` — the `raw == ""` branch still catches misses after this change, so no further edits were needed.

## Test plan

- [x] `go test ./pkg/plugin/implementation/cache/...` — all three subtests (hit / miss / error) pass
- [x] `go test ./pkg/plugin/implementation/payloadstore/...` — `TestExists_*` pass without modification

### DIFF
--- a/pkg/plugin/implementation/cache/cache.go
+++ b/pkg/plugin/implementation/cache/cache.go
@@ -123,7 +123,7 @@ func (c *Cache) Get(ctx context.Context, key string) (string, error) {
 			telemetry.AttrOperation.String("get"),
 		}
 		switch {
-		case err == redis.Nil:
+		case errors.Is(err, redis.Nil):
 			c.metrics.CacheMissesTotal.Add(spanCtx, 1, metric.WithAttributes(attrs...))
 			c.metrics.CacheOperationsTotal.Add(spanCtx, 1,
 				metric.WithAttributes(append(attrs, telemetry.AttrStatus.String("miss"))...))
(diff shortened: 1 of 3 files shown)


nirmay (approved):
Approved.

PR #725, opened 2026-05-19
refactor(dediregistry): remove `registryName` config and hardcode wildcard constant

### LINKED ISSUE #719: Remove `registryName` config from DediRegistry Plugin
## Summary

The `registryName` configuration field in the DediRegistry plugin should be removed and hardcoded internally. It is effectively a constant — always `subscribers.beckn.one` — and exposing it as a user-configurable option is misleading and error-prone.

## Background

The DediRegistry plugin constructs lookup URLs of the form:

```
{url}/lookup/{subscriber_id}/{registryName}/{key_id}
```

The value `subscribers.beckn.one` is a wildcard constant defined by the beckn registry service that enables searches across all cached registries. It is not a registry name in the traditional sense and must not be changed for lookups to work correctly.

This is already acknowledged in the existing config documentation and example configs:

```yaml
registryName: subscribers.beckn.one # This is the wildcard string used to lookup across registries in Beckn One. Do not change this.
```

```
"Keep this value as `subscribers.beckn.one` so lookups include all cached registries"
```

The fact that users are warned never to change it is a strong signal it should not be a config option at all.

## Problem

- Users who are supposed to filter by network should use `allowedNetworkIDs`, not `registryName`. The current design implies `registryName` is the correct knob to control scoping, which is wrong.
- Having `registryName` as a required field adds unnecessary cognitive overhead during setup.
- Any misconfiguration (e.g., setting it to a specific registry name instead of the wildcard) silently breaks cross-registry lookups in non-obvious ways.

## Proposed Change

- Remove `registryName` from the `Config` struct and all config files.
- Hardcode the value `subscribers.beckn.one` directly in the lookup URL construction within the plugin.
- Update `CONFIG.md` and `README.md` for the plugin to remove references to this field.
- Remove validation that currently enforces `registryName` is non-empty.
- Update tests accordingly.

## Impact

- **Config files:** `config/local-beckn-one-bap.yaml`, `config/local-beckn-one-bpp.yaml` (and any others referencing `registryName` under the dediregistry plugin config)
- **Plugin code:** `pkg/plugin/implementation/dediregistry/dediregistry.go`, `pkg/plugin/implementation/dediregistry/cmd/plugin.go`
- **Docs:** `CONFIG.md`, `pkg/plugin/implementation/dediregistry/README.md`

nirmalnr:
## Summary
- Removes the `registryName` field from `Config` struct and all config files — it was always `subscribers.beckn.one` and docs warned users to not change it
- Hardcodes the value as a package-level constant `dediAllRegistriesWildcard` in the lookup URL construction, making the intent clear
- Removes the now-redundant validation check for a non-empty `registryName`

## Why
`subscribers.beckn.one` is a wildcard used in the registry service to search across all cached registries. Exposing it as a required user-configurable field was misleading (implying it's tunable), added setup friction, and risked silent misconfiguration if a user set it to a specific registry name instead of the wildcard.

## Files Changed
- `pkg/plugin/implementation/dediregistry/dediregistry.go` — added constant, removed `Config.RegistryName`, updated URL construction and validation
- `pkg/plugin/implementation/dediregistry/cmd/plugin.go` — removed `RegistryName` from `parseConfig`
- `config/local-beckn-one-bap.yaml`, `config/local-beckn-one-bpp.yaml` — removed `registryName` lines
- `CONFIG.md`, `pkg/plugin/implementation/dediregistry/README.md` — updated docs
- `dediregistry_test.go`, `cmd/plugin_test.go` — removed `RegistryName` from all Config literals, removed "missing registryName" test cases

## Test Plan
- [x] `go test ./pkg/plugin/implementation/dediregistry/...` passes (verified locally)
- [x] Existing lookup URL path assertions still pass (URL still contains `subscribers.beckn.one`, now hardcoded)
- [x] No `registryName` key in any config file or `Config` struct

Closes #719

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -586,7 +586,6 @@ registry:
   id: dediregistry
   config:
     url: "https://fabric.nfh.global/registry/dedi"
-    registryName: "subscribers.beckn.one"
     allowedNetworkIDs: "commerce-network.org/prod,local-commerce.org/production"
     timeout: 30
     retry_max: 3
@@ -596,7 +595,6 @@ registry:
 
 **Parameters**:
 - `url`: Beckn registry base URL including the `/dedi` path (Required)
-- `registryName`: Registry name used in the lookup path. Keep this value as `subscribers.beckn.one` so lookups include all cached registries. Use `allowedNetworkIDs` if you want lookup to be restricted within one or more networks. (Required)
 - `allowedNetworkIDs`: Comma-separated list of allowed network IDs used to restrict lookup results to specific networks. See [Beckn network setup documentation](https://docs.beckn.io/creating-an-open-network/setting-up-the-network-environment#registering-the-network-on-beckn-fabric) for more on network IDs. (Optional)
 - `timeout`: Request timeout in seconds (Optional, default: client default)
 - `retry_max`: Maximum number of retry attempts (Optional, default: 4)
--- a/config/local-beckn-one-bap.yaml
+++ b/config/local-beckn-one-bap.yaml
@@ -52,7 +52,6 @@ modules:
           id: dediregistry
           config:
             url: https://fabric.nfh.global/registry/dedi # This is the registry URL.
-            registryName: subscribers.beckn.one # This is the wildcard string used to lookup across registries in Beckn One. Do not change this.
             timeout: 10
             retry_max: 3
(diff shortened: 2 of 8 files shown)


nirmay (approved):
Approved.

PR #726, opened 2026-05-19
feat(proto-v2): HTTP 409 AckNoCallback response type (#680)

### LINKED ISSUE #680: [Proto-v2.0] HTTP 409 AckNoCallback response type
## Summary

Beckn v2.0 LTS introduces a new response type — `AckNoCallback` (HTTP `409 Conflict`) — where the receiver acknowledges receipt of the request but explicitly signals that no async callback will follow. ONIX currently has no mechanism to produce this response.

Spec reference: https://github.com/beckn/protocol-specifications-v2/blob/core-v2.0.0-lts/api/v2.0.0/beckn.yaml — `409` response definition

## Response Shape

```json
HTTP 409 Conflict

{
  "context": { "action": "...", "messageId": "...", "version": "2.0.0", ... },
  "message": {
    "ack": { "status": "ACK" }
  }
}
```

Note: status is `"ACK"` (not `"NACK"`), and there is **no `error` field**. The HTTP 409 status code itself signals "received but no callback coming."

## Use Case

A BPP may use this when it receives a valid `search` or `select` but has no matching results and does not intend to send `on_search` / `on_select`. Rather than leaving the BAP waiting for a callback that will never arrive, it responds with 409.

## Files Affected

- `pkg/model/error.go` — add `AckNoCallbackErr` error type with `BecknError()` returning nil (no error body)
- `pkg/response/response.go` — `SendNack()` handles `AckNoCallbackErr` → HTTP 409 with ACK status body
- Only active when `BecknVersion == "2.0.0"`

## Acceptance Criteria

- [ ] A plugin returning `AckNoCallbackErr` causes ONIX to send HTTP `409`
- [ ] Response body contains `message.ack.status: "ACK"` with **no** `error` field
- [ ] `context` is echoed in the response body (consistent with LTS response envelope)
- [ ] `version != "2.0.0"` does not expose this response type
- [ ] Unit tests for the new error type and response shape

## Dependencies

- #676 — `BecknVersion` in `StepContext`
- #678 — Error response structure (response envelope must be settled first)

nirmay:
## Summary

- Introduces `AckNoCallbackErr` — a new error type pipeline steps can return when the receiver authenticated and accepted a request but will not send an async callback (no matching catalog, inventory unavailable, provider closed)
- ONIX maps this to HTTP 409 Conflict with the v2 LTS flat response shape (`message.status`, `message.messageId`, `message.error`)
- `ackSigner` runs automatically on the 409 path — no new wiring needed
- Pre-v2 requests that encounter `AckNoCallbackErr` fall through to 500 (no-op for legacy)

## What changed

- `pkg/model/error.go` — `AckNoCallbackErr{Status, Err}` with `NewAckNoCallbackErr` constructor; `Status` carries `ACK` (accepted, no callback) or `NACK` (rejected outright) per spec
- `core/module/handler/responsestep.go` — `nackBecknError` returns a third `model.Status` (body-level status); `nackBodyBytes`, `nackBytes`, `nack`, `sendNack` updated to thread it through
- `core/module/handler/responsestep_test.go` — v2 ACK/NACK body status cases added to `TestSendNack`; `TestSendNack_AckNoCallback_PreV2_FallsThrough` added

## Path 1 vs Path 2

**Path 1 (upstream app returns 409):** already worked before this PR. `ackSignerStep.RunOnResponse` signs and forwards transparently — `AckNoCallbackErr` is not involved.

**Path 2 (ONIX pipeline step decides):** what `AckNoCallbackErr` enables. No existing step currently returns it — this PR wires the infrastructure. A follow-up is needed to identify which steps should use it (e.g. a routing step that finds zero matching providers).

## Test plan

- [x] `go test ./core/module/handler/... ./pkg/model/...` passes
- [x] v2 request with step returning `AckNoCallbackErr(StatusACK, ...)` → 409, body `{"message":{"status":"ACK","messageId":"...","error":{...}}}`, `Signature` header set
- [x] v2 request with step returning `AckNoCallbackErr(StatusNACK, ...)` → 409, body `{"message":{"status":"NACK",...}}`
- [x] Pre-v2 request with step returning `AckNoCallbackErr` → 500 (fallthrough)
- [x] Upstream app returning 409 on URL-routing path → forwarded unchanged (Path 1 regression)

Closes #680

### DIFF
--- a/pkg/model/error.go
+++ b/pkg/model/error.go
@@ -111,3 +111,32 @@ func (e *NotFoundErr) BecknError() *Error {
 		Message: "Endpoint not found: " + e.Error(),
 	}
 }
+
+// AckNoCallbackErr is returned by a step when the receiver has authenticated and
+// accepted the request but will not send an async callback — for example, no
+// matching catalog, inventory unavailable, or provider closed. ONIX maps this to
+// HTTP 409 Conflict using the v2 flat response shape. For protocol versions prior
+// to 2.0.0 this error falls through to a 500 Internal Server Error.
+type AckNoCallbackErr struct {
+	// Status is ACK when the request was accepted but no callback will follow,
+	// or NACK when the request was outright rejected.
+	Status Status
+	// Err explains why no callback will be sent. Required by the spec.
+	Err *Error
+}
+
+// NewAckNoCallbackErr constructs an AckNoCallbackErr.
+// Use StatusACK for "accepted but no callback" and StatusNACK for outright rejection.
+func NewAckNoCallbackErr(status Status, err *Error) *AckNoCallbackErr {
+	return &AckNoCallbackErr{Status: status, Err: err}
+}
+
+// Error implements the error interface.
+func (e *AckNoCallbackErr) Error() string {
+	return fmt.Sprintf("AckNoCallback(status=%s): %s", e.Status, e.Err.Error())
+}
+
+// BecknError returns the wrapped *Error payload.
+func (e *AckNoCallbackErr) BecknError() *Error {
+	return e.Err
+}
--- a/pkg/model/error.go
+++ b/pkg/model/error.go
@@ -127,7 +127,11 @@ type AckNoCallbackErr struct {
 
 // NewAckNoCallbackErr constructs an AckNoCallbackErr.
 // Use StatusACK for "accepted but no callback" and StatusNACK for outright rejection.
+// Panics if err is nil — the spec requires an error explanation on every 409 response.
 func NewAckNoCallbackErr(status Status, err *Error) *AckNoCallbackErr {
+	if err == nil {
+		panic("AckNoCallbackErr: Err is required")
+	}
 	return &AckNoCallbackErr{Status: status, Err: err}
 }
 
--- a/core/module/handler/responsestep.go
+++ b/core/module/handler/responsestep.go
@@ -70,36 +70,45 @@ func sendAck(ctx context.Context, w http.ResponseWriter) {
 	}
 }
 
-// nackBecknError maps an error to its Beckn model.Error representation and HTTP
-// status code.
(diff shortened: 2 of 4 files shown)


pkg/model/error.go
```diff
@@ -111,3 +111,32 @@ func (e *NotFoundErr) BecknError() *Error {
...
+	Status Status
+	// Err explains why no callback will be sent. Required by the spec.
+	Err *Error
+}
+
+// NewAckNoCallbackErr constructs an AckNoCallbackErr.
+// Use StatusACK for "accepted but no callback" and StatusNACK for outright rejection.
+func NewAckNoCallbackErr(status Status, err *Error) *AckNoCallbackErr {
+	return &AckNoCallbackErr{Status: status, Err: err}
+}
```
nirmalnr, line 136: Should we add a nil check for `err` here? Currently nothing prevents a caller from doing:

```go
model.NewAckNoCallbackErr(model.StatusACK, nil)
```

If `nil` is passed, two things break downstream:
1. `Error()` panics — it calls `e.Err.Error()` on a nil pointer
2. `BecknError()` returns `nil`, which serialises as `"error": null` in the response body

Since no production step calls this yet, it's safe now — but once follow-up PRs start wiring steps to return `AckNoCallbackErr`, a future contributor could easily pass `nil` without realising it causes a panic.

A simple guard in the constructor would make the contract explicit:
```go
func NewAckNoCallbackErr(status Status, err *Error) *AckNoCallbackErr {
    if err == nil {
        panic("AckNoCallbackErr: Err is required")
    }
    return &AckNoCallbackErr{Status: status, Err: err}
}
```

Or alternatively, handle `nil` gracefully in `Error()` and `BecknError()` if there are valid cases where no error detail is needed.

nirmalnr (approved):

PR #731, opened 2026-05-20
feat: made reqMapper a dedicated step instead of a plugin

### LINKED ISSUE #722: Make request transformation run as a handler step instead of middleware so signature and schema validation happen before payload mutation
Right now reqmapper runs as middleware, which means it transforms the payload before the handler steps run. This causes a problem on receiver flows because network payloads are signed in their original format, but we are mutating them before validateSign runs. As a result, signature verification can fail even when the sender keys are correct.

We should move transformation out of middleware and make it a proper handler step. That will let us place it differently for caller and receiver flows.

**Expected behavior:**
On caller flows, transform first, then sign and send.
On receiver flows, validate signature and schema on the original network payload first, then transform it before forwarding internally.
This change will make signature verification reliable while still allowing conversion in the right place.

abhey-WIL:
This PR fixes reqmapper so request/response transformation can run as an explicit handler step instead of only as middleware, which is required for correct receiver-side signature and schema validation.

The main change is a new step plugin, reqmapperstep, backed by the existing reqmapper mapping logic. The handler now supports a dedicated plugins.reqMapper config slot and a reqMapper step in the pipeline, similar to checkPolicy. That lets receiver flows validate the original inbound network payload first, then transform it afterward, while caller flows can transform before signing and forwarding.

The PR also fixes two runtime issues discovered during the refactor:

forwarded requests now resync body metadata like Content-Length after transformation
instrumented handler steps now propagate mutated StepContext fields back to the parent context, so downstream steps see the transformed body and any other state changes
Docs were updated to reflect the final design:

reqmapper remains the middleware plugin
reqmapperstep is the step plugin used by reqMapper
recommended step ordering is documented for caller and receiver flows


nirmay:
<html><head></head><body><hr>
<h2>PR #731 — Code Review</h2>
<h3>What this PR does</h3>
<p>Adds <code>reqmapperstep</code> as a handler-pipeline <code>Step</code> so receiver flows can control <em>where</em> in the pipeline payload transformation runs — specifically, after signature and schema validation, not before. It also fixes <code>InstrumentedStep</code> to propagate mutated <code>Body</code>/<code>Request</code> back to the caller context (without which any body-mutating step inside a telemetry wrapper would silently lose its changes), and introduces <code>syncRequestBody</code> to update <code>Content-Length</code>, <code>GetBody</code>, and <code>TransferEncoding</code> after transformation. The motivation is correct; the implementation has several issues.</p>
<hr>
<h3>Architecture</h3>
<p><strong>The <code>reqmapper</code> middleware plugin must be removed, not kept.</strong><br>
The PR adds <code>reqmapperstep</code> but keeps the original <code>reqmapper</code> middleware, leaving two mechanisms for the same transformation. Placing <code>transformPayload</code> (the step) first in the pipeline achieves exactly the same effect as middleware, plus gives positional control. The middleware form runs before sign/schema validation by construction — it is the wrong default for receiver flows and a footgun for anyone reading the config. There are no users of this codebase, so there is no backward compatibility cost. Remove <code>reqmapper</code> middleware, its <code>cmd/plugin.go</code>, its entry from <code>build-plugins.sh</code>, and its <code>middleware:</code> example from CONFIG.md in this same PR. Keeping it creates ongoing confusion with no benefit.</p>
<p><strong>The global singleton <code>MappingEngine</code> is a latent bug made worse by this PR.</strong><br>
<code>initMappingEngine</code> uses package-level <code>engineOnce</code>/<code>engineInstance</code>. Both <code>NewReqMapper</code> (middleware) and <code>NewReqMapperStep</code> (step) call it; the first caller wins and all subsequent calls silently return the same engine regardless of their config. If the middleware is removed this becomes less acute, but the issue survives: a process that loads two handlers with different <code>mappingsFile</code> paths (BAP and BPP, for example) will silently lock in the first config for both. The fix is straightforward — remove the package-level singleton and construct a <code>MappingEngine</code> per caller:</p>
<pre><code class="language-go">func initMappingEngine(cfg *Config) (*MappingEngine, error) {
    engine := &amp;MappingEngine{config: cfg, ...}
    // no Once, no global
    ...
    return engine, nil
}
</code></pre>
<hr>
<h3>Handler integration (<code>stdHandler.go</code>, <code>step.go</code>)</h3>
<p><strong><code>newReqMapperHandlerStep</code> adds no value and should be removed.</strong></p>
<pre><code class="language-go">func newReqMapperHandlerStep(step definition.Step) (definition.Step, error) {
    if step == nil {
        return nil, fmt.Errorf("invalid config: ReqMapper plugin not configured")
    }
    return step, nil
}
</code></pre>
<p>This wraps a <code>Step</code> in a <code>Step</code> with a nil-guard as the only logic. It differs from <code>checkPolicyStep</code>, which wraps a domain-specific <code>PolicyChecker</code> interface into the <code>Step</code> contract — that is a meaningful adapter. Here there is nothing to adapt. Inline the nil-check directly into <code>initSteps</code>:</p>
<pre><code class="language-go">case "transformPayload":
    if h.payloadTransformer == nil {
        return fmt.Errorf("invalid config: PayloadTransformer step not configured")
    }
    s = h.payloadTransformer
</code></pre>
<p><strong><code>syncRequestBody</code> header manipulation is redundant and has a logic defect.</strong><br>
Go's HTTP stack uses <code>r.ContentLength</code> for outbound requests — setting or deleting the <code>Content-Length</code> header on an <code>*http.Request</code> has no effect. The header manipulation is dead code. Remove it. The remaining body of the function is correct:</p>
<pre><code class="language-go">func syncRequestBody(r *http.Request, body []byte) {
    if r == nil {
        return
    }
    r.Body = io.NopCloser(bytes.NewReader(body))
    r.ContentLength = int64(len(body))
    r.TransferEncoding = nil
    r.GetBody = func() (io.ReadCloser, error) {
        return io.NopCloser(bytes.NewReader(body)), nil
    }
}
</code></pre>
<p>The current code also has a consistency defect: for an empty body it sets <code>r.ContentLength = 0</code> and then deletes the header, while for a non-empty body it sets both. The field and header are out of sync in the empty case, which would surprise a reader even if Go ignores the header.</p>
<hr>
<h3><code>reqmapper.go</code> — step implementation</h3>
<p><strong><code>parseRequestBody</code> swallows the JSON decode error.</strong></p>
<pre><code class="language-go">if err := json.Unmarshal(body, &amp;req); err != nil {
    return nil, fmt.Errorf("Failed to decode request body")  // err discarded
}
</code></pre>
<p>The original error is thrown away, making malformed-payload diagnosis in production impossible. Use <code>%w</code>:</p>
<pre><code class="language-go">return nil, fmt.Errorf("failed to decode request body: %w", err)
</code></pre>
<p>Also: Go error strings must be lowercase — <code>"Failed to..."</code> violates the convention and <code>staticcheck</code> will flag it. The other two error strings in <code>parseRequestBody</code> are lowercase; this one is not.</p>
<p><strong><code>reqMapperStep.Run</code> does not update <code>ctx.Request.GetBody</code>.</strong></p>
<pre><code class="language-go">ctx.Request.Body = io.NopCloser(bytes.NewReader(mappedBody))
ctx.Request.ContentLength = int64(len(mappedBody))
// GetBody is not updated
</code></pre>
<p><code>syncRequestBody</code> sets <code>GetBody</code> but the step does not. If any downstream path calls <code>r.GetBody()</code> (HTTP/1.1 redirects, retry middleware), it returns the pre-transformation body. Add:</p>
<pre><code class="language-go">ctx.Request.GetBody = func() (io.ReadCloser, error) {
    return io.NopCloser(bytes.NewReader(mappedBody)), nil
}
</code></pre>
<p><strong>The nil <code>StepContext</code> guard is wrong-typed.</strong></p>
<pre><code class="language-go">if ctx == nil {
    return model.NewBadReqErr(errors.New("step context cannot be nil"))
}
</code></pre>
<p>A nil <code>StepContext</code> is a programmer error at the call site, not a bad client request. It should either panic or (preferably) have no guard at all — callers never pass nil. Using <code>BadReqErr</code> here means a bug in the adapter would produce an HTTP 400, not a 500 or a panic you can catch in tests.</p>
<hr>
<h3><code>reqmapperstep/cmd/plugin.go</code> — plugin entry point</h3>
<p><strong>Config parsing is duplicated from <code>reqmapper/cmd/plugin.go</code>.</strong><br>
The <code>buildConfig</code> refactor in <code>reqmapper/cmd/plugin.go</code> was done to enable sharing, but <code>buildConfig</code> is unexported so <code>reqmapperstep/cmd/plugin.go</code> re-inlines the same 5 lines. Export it as <code>reqmapper.BuildConfig</code> and call it from both entry points:</p>
<pre><code class="language-go">// reqmapper package
func BuildConfig(c map[string]string) *Config { ... }

// reqmapperstep/cmd/plugin.go
step, err := reqmapper.NewReqMapperStep(reqmapper.BuildConfig(c))
</code></pre>
<p><strong>No test file.</strong><br>
Every other plugin <code>cmd/</code> directory has a <code>plugin_test.go</code>. <code>reqmapperstep/cmd/</code> has none.</p>
<hr>
<h3><code>step_instrumentor.go</code></h3>
<p><strong>The write-back change is correct but needs a comment.</strong></p>
<pre><code class="language-go">ctx.Body = stepCtx.Body
ctx.Request = stepCtx.Request
</code></pre>
<p><code>Body</code> is <code>[]byte</code> — a value type. Without this write-back, mutations made by the inner step (<code>stepCtx.Body = mappedBody</code>) are invisible to the pipeline after <code>InstrumentedStep.Run</code> returns. The fix is right. <code>ctx.Request</code> is a pointer, so mutations made through it (<code>.Body</code>, <code>.ContentLength</code>) are already visible via the original pointer — that line is defensive and harmless, but future-proof if a step ever replaces the pointer entirely.</p>
<p>What is confusing is the asymmetry with <code>Route</code>:</p>
<pre><code class="language-go">if stepCtx.Route != nil {    // conditional
    ctx.Route = stepCtx.Route
}
ctx.Body = stepCtx.Body      // unconditional — why?
</code></pre>
<p>The reason: <code>Route == nil</code> means "not yet determined" and an unconditional write-back would erase a previously-set route. <code>Body == nil</code> is not semantically meaningful (the pipeline always has a body); unconditional is correct. Add a comment so the next reader does not make the write-back conditional:</p>
<pre><code class="language-go">// Body is unconditional — nil body is valid and the nil-guard used for Route
// would silently discard a step that intentionally clears the body.
ctx.Body = stepCtx.Body
</code></pre>
<hr>
<h3>Naming</h3>
<p><code>reqmapperstep</code> and <code>reqMapper</code> (the config step key) are both poor names. The existing step keys are action-verb-oriented: <code>validateSign</code>, <code>validateSchema</code>, <code>addRoute</code>, <code>checkPolicy</code>. Suggested names:</p>

Artifact | Current | Suggested
-- | -- | --
YAML step key | reqMapper | transformPayload
Plugin ID | reqmapperstep | payloadtransformer
Go struct | reqMapperStep | payloadTransformerStep
stdHandler field | reqMapper | payloadTransformer
PluginCfg field / YAML tag | ReqMapper / reqMapper | PayloadTransformer / payloadTransformer


<p>The existing <code>step_instrumentor_test.go</code> tests <code>Run</code> but does not verify that the <code>Body</code> write-back actually propagates. Add a test with a stub step that sets <code>ctx.Body = []byte("mutated")</code> and assert the outer <code>StepContext.Body</code> reflects it after <code>InstrumentedStep.Run</code>.</p>
<hr>
<h3>Config files</h3>
<p><code>reqmapperstep</code> is added to all three <code>plugin.yaml</code> files unconditionally. If a handler has no <code>plugins.payloadTransformer</code> config, the <code>.so</code> loads and does nothing — wasted startup. Only include it in configs that actually configure the step, or note that the plugin list is purely advisory (if the manager handles unrecognised entries gracefully). Also, all three <code>plugin.yaml</code> files are missing a trailing newline — add it.</p>
<hr>
<h3>Summary</h3>
<p>The pipeline step approach is the right design. The <code>InstrumentedStep</code> body write-back fix is necessary and correct. The blockers before merge are:</p>
<ol>
<li>Remove <code>reqmapper</code> middleware entirely</li>
<li>Fix the global singleton <code>MappingEngine</code> — construct per caller</li>
<li>Fix <code>parseRequestBody</code> to wrap the json error with <code>%w</code> and lowercase the string</li>
<li>Fix <code>reqMapperStep.Run</code> to update <code>ctx.Request.GetBody</code></li>
<li>Clean up <code>syncRequestBody</code> — remove the header manipulation, keep field updates</li>
<li>Remove <code>newReqMapperHandlerStep</code> — inline the nil-check into <code>initSteps</code></li>
<li>Export <code>BuildConfig</code> and remove the duplication in <code>reqmapperstep/cmd</code></li>
<li>Rename <code>reqmapperstep</code> → <code>payloadtransformer</code>, step key <code>reqMapper</code> → <code>transformPayload</code></li>
<li>Add tests for the new code paths above</li>
</ol></body></html>

nirmay:
---

## PR #731 — Code Review

### Overview

The update lands the core design correctly. The `reqmapper` middleware is gone; `reqmapper` is now a step plugin directly. The global singleton engine is replaced with per-instance construction. `BuildConfig` is exported and the duplicate config parsing is eliminated. `syncRequestBody` is cleaned up. `parseRequestBody` wraps the JSON error. `reqMapperStep.Run` sets `GetBody` and clears `TransferEncoding`. The `newReqMapperHandlerStep` wrapper is gone and the nil-check is inlined. `reqmapperstep` as a separate plugin is eliminated. Naming is consistent throughout — `payloadTransformer` in the config struct, `transformPayload` as the step key. Several issues remain.

---

### `step_instrumentor.go`

**`ctx.Body` is written back twice.**

```go
ctx.Body = stepCtx.Body     // first
ctx.Request = stepCtx.Request
ctx.Body = stepCtx.Body     // duplicate — remove one
```

**`Request` and `RespHeader` write-backs are unnecessary and should be removed.**

Both `*http.Request` and `http.Header` are reference types. When `InstrumentedStep` does `stepCtx := *ctx`, the pointer/map value is copied — not the underlying object. Any mutation a step makes through the pointer (`ctx.Request.Body = ...`, `ctx.Request.Header.Set(...)`, writing into `ctx.RespHeader`) is already visible through the original pointers without any write-back. The write-back only matters if a step replaces the pointer or map entirely — no real step in the pipeline does this. Remove both lines:

```go
ctx.Request = stepCtx.Request   // remove
ctx.RespHeader = stepCtx.RespHeader   // remove
```

The correct write-back set is only the fields where a value copy genuinely loses mutations:

```go
// Value types — copy mutations are not visible in the original.
ctx.Body  = stepCtx.Body
ctx.SubID = stepCtx.SubID
ctx.Role  = stepCtx.Role
// Pointer used as a set/not-set signal — unconditional write-back intentional.
ctx.Route = stepCtx.Route
```

**`net/http` and `net/http/httptest` imports in `step_instrumentor_test.go` follow from the same problem.**

`mutatingStep` uses `http.Header` and `httptest.NewRequest` specifically to exercise the `Request` and `RespHeader` write-backs. Once those write-backs are removed from the instrumentor, that part of the test goes away with them and so do the imports. The test should cover the value-type propagation — `Body`, `SubID`, `Role`, `Route` — which needs no `net/http` dependency at all.

**Unconditional `Route` write-back is a behavioral change worth documenting.**

The original code guarded the Route write-back:
```go
if stepCtx.Route != nil {
    ctx.Route = stepCtx.Route
}
```

Making it unconditional means a step that explicitly sets `ctx.Route = nil` will now clear a route set by an earlier step. No current step does this, but the guard was the policy-level protection against it. Add a comment:

```go
// Route is written back unconditionally. Setting ctx.Route = nil inside a
// step is a valid signal that no route was determined and will clear any
// route set by a prior step.
ctx.Route = stepCtx.Route
```

---

### `reqmapper_test.go` — test regression

The entire test file (240 lines) is deleted with no replacement for the logic it covered. What is now completely untested:

- `NewReqMapperStep` — nil config, invalid role, valid config
- `reqMapperStep.Run` — successful transformation, fallback when the engine errors, empty body
- `parseRequestBody` — malformed JSON, missing `context`, missing `action`, empty `action`
- `MappingEngine.Transform` — unknown action fallback, unknown role fallback
- `MappingEngine.ReloadMappings`

The only new test is `TestProviderNew_ReturnsStep` in `cmd/plugin_test.go`, which confirms `provider.New` returns a non-nil step without calling `Run` at all. The transformation logic — the entire purpose of this plugin — has zero test coverage. These tests must be rewritten targeting `NewReqMapperStep` and `reqMapperStep.Run`.

The old error-path tests in `cmd/plugin_test.go` (`TestProviderNewMissingRole`, `TestProviderNewInvalidRole`) are also gone with no replacement. Add them back:

```go
func TestProviderNew_MissingRole(t *testing.T) {
    _, _, err := (provider{}).New(context.Background(), map[string]string{})
    require.Error(t, err)
}
func TestProviderNew_InvalidRole(t *testing.T) {
    _, _, err := (provider{}).New(context.Background(), map[string]string{"role": "invalid"})
    require.Error(t, err)
}
```

---

### `config/mappings.yaml` and reference configs

`config/mappings.yaml` maps every action to `$` — the JSONata identity expression that returns the input unchanged. This file is wired into all four handlers in both `local-beckn-one-bap.yaml` and `local-beckn-one-bpp.yaml`, so every request in the reference deployment now parses JSON, evaluates a JSONata expression, and re-serialises for zero effect.

`transformPayload` is an opt-in step for flows that actually need transformation. The reference configs should not opt everyone into an identity transform. Either remove `payloadTransformer` config and `transformPayload` from the step lists in these files, or replace `config/mappings.yaml` with a real example transformation so the file demonstrates actual usage rather than a no-op placeholder.

---

### `reqMapperStep.Run` — nil context guard

```go
if ctx == nil {
    return fmt.Errorf("step context cannot be nil")
}
```

The pipeline never passes a nil `StepContext`. This is dead code, and returning a plain error here would surface as an opaque pipeline failure rather than identifying the programming mistake. Remove the guard.

---

### Summary

Fix before merge:

1. Remove the duplicate `ctx.Body` write-back in `step_instrumentor.go`
2. Remove the `ctx.Request` and `ctx.RespHeader` write-backs — both are reference types that do not need them; also simplifies the test and removes the `net/http`/`net/http/httptest` imports
3. Add a comment on the unconditional `ctx.Route` write-back explaining the behavioral intent
4. Rewrite `reqmapper_test.go` covering `NewReqMapperStep`, `reqMapperStep.Run`, and `parseRequestBody` — the transformation logic must have test coverage
5. Restore the missing error-path tests in `cmd/plugin_test.go`
6. Remove `transformPayload` from the reference configs, or replace `config/mappings.yaml` with a real transformation example. I would prefer the first. also, include a sample config in the CONFIG.md file.
7. Remove the nil `StepContext` guard from `reqMapperStep.Run`

nirmay:
## PR #731 — Code Review

### What's resolved

The previous feedback is fully addressed. The duplicate `ctx.Body` write-back is gone. `ctx.Request` and `ctx.RespHeader` are no longer written back in `InstrumentedStep` — correctly recognising that both are reference types whose field-level mutations are already visible through the original pointer. `step_instrumentor_test.go` now has no `net/http` dependency. The reqmapper test file is completely rewritten to cover the new step interface with good breadth: constructor validation, successful transformation with full `GetBody`/`ContentLength`/`TransferEncoding` verification, engine-error fallback, empty body error path, `parseRequestBody` edge cases, engine fallback for unknown action and role, and `ReloadMappings`. Error-path tests are restored in `cmd/plugin_test.go`.

---

### Merge blockers

**The PR has a merge conflict with main and cannot be merged as-is.** `mergeStateStatus: DIRTY`. Rebase onto main before requesting a final merge review.

**The `otelsetup` test changes do not belong in this PR.**  
The diff silences cleanup errors in `otelsetup/cmd/plugin_test.go` with `_ = cleanup()`. Main still has `assert.NoError(t, err, "cleanup() should not return error")` for all of those call sites — this change is not in main. It is a real fix (cleanup legitimately fails without an OTLP backend in unit tests), but it is entirely unrelated to reqmapper. Move it to a separate PR. If it was needed to make CI green on this branch, the right fix is still the separate PR — merge that first, then rebase.

**The `router` test changes do not belong in this PR.**  
The diff trims two expected error strings in `router_test.go`, removing the Go `url.Parse` detail from the tail:
```diff
-wantErr: `...target type bpp: parse "htp:// invalid-url.com": invalid character " " in host name`,
+wantErr: `...target type bpp:`,
```
Main still has the full error strings. This looks like a Go version difference causing the test to fail on the PR branch. It is a real fix, but it belongs in its own PR. File a ticket and keep it out of here.

---

### Test coverage

Coverage is now in good shape. A few observations:

**`TestReqMapperStepRun_Success` relies on testdata having a non-trivial mapping.** The assertion `require.NotEqual(t, string(body), string(ctx.Body))` only holds if the testdata `mappings.yaml` actually transforms the search payload. The subsequent check for `message.filters` makes this concrete, so it is fine as long as testdata is not changed to identity mappings. The `GetBody` round-trip verification is a nice touch.

**`TestMappingEngineReloadMappings` accesses `bapMaps` directly without locks.** The old version used `engine.mutex.RLock()`. The new version reads and deletes from the map in a single goroutine without any concurrent access, so the race detector will not flag it. Acceptable for a unit test, but the prior approach was more consistent with how the production code treats the map.

**`syncRequestBody` has no unit test.** It is a small function and is exercised indirectly by handler integration tests, so this is not a blocker, but a direct test for the nil-guard and field-assignment behaviour would be cheap to add.

**`initSteps "transformPayload"` case has no unit test.** The nil-plugin error path (`payloadTransformer not configured`) is not explicitly covered. Also minor, but worth adding alongside the existing `initSteps` tests.

---

### `config/mappings.yaml`

The file is still present — 52 lines of identity mappings (`$` for every action and role). It is no longer wired into the `local-beckn-one-*.yaml` configs, so it does not cause unnecessary JSONata overhead at runtime. As a standalone reference template this is acceptable, but it needs a comment at the top:

```yaml
# Template mappings file. Replace '$' (JSONata identity) with actual
# transformation expressions for your deployment.
```

Without that, someone copying it will not realise all expressions are no-ops.

---

### Step reordering in local configs

Both `local-beckn-one-bap.yaml` and `local-beckn-one-bpp.yaml` now have `validateSchema` before `addRoute` on the receiver pipeline:

```diff
-        - addRoute
         - validateSchema
+        - addRoute
```

This is correct — schema validation should not depend on having a route. The change is unrelated to reqmapper but touches the same files and is safe to carry here.

---

### Summary

Two things block merge:

1. **Rebase onto main** — conflict must be resolved
2. **Extract `otelsetup` and `router` test fixes into separate PRs** (or merge them into main first and rebase) — they are real fixes but they do not belong here

Everything else is in good shape. Once the branch is clean and the unrelated changes are out, this is ready.

nirmay:
Good — the PR is now `MERGEABLE` and `CLEAN`. 

## PR #731 — Final Review

### State

Rebased on main, conflict-free, `mergeStateStatus: CLEAN`. The otelsetup cleanup-error silencing is removed. The router error-message truncation is gone; only a cosmetic struct-field alignment fix remains in `router_test.go`, which is harmless.

---

### One blocker

**`TestInstrumentedStep_PropagatesMutations` was dropped — `mutatingStep` is now dead code.**

The diff adds the `mutatingStep` struct to `step_instrumentor_test.go`:

```go
type mutatingStep struct{}

func (mutatingStep) Run(ctx *model.StepContext) error {
    ctx.Body  = []byte(`{"mutated":true}`)
    ctx.Route = nil
    ctx.SubID = "sub-updated"
    ctx.Role  = model.RoleBPP
    return nil
}
```

But the test function that exercises it (`TestInstrumentedStep_PropagatesMutations`) is nowhere in the diff. Main does not have it either — confirmed. The struct was left behind when the test was cleaned up. The result: the behavioral change in `step_instrumentor.go` (writing back `Body`, `Route`, `SubID`, `Role` unconditionally) has zero test coverage. Add the test back — it was present in a prior iteration of this PR and is the right thing to have.

---

### minors

**`syncRequestBody` is called unconditionally, outside the success guard.**

```go
// after the step loop
syncRequestBody(r, stepCtx.Body)      // ← called even when pipelineErr != nil

if pipelineErr == nil {
    // forward / publish
}
```

On the error path the NACK is already written inside the loop; `r` is not used afterward. The call is harmless but is dead work on every failed request. The original code had `r.Body = ...` inside `if pipelineErr == nil`. Move `syncRequestBody` inside the same guard:

```go
if pipelineErr == nil {
    syncRequestBody(r, stepCtx.Body)
    ...
}
```

---

nirmay (approved):
Approved.

PR #732, opened 2026-05-21
fix(manifestloader): resolve cache-miss log noise and Redis fault tolerance

### LINKED ISSUE #724: fix(manifestloader): loadFromCache logs "unexpected end of JSON input" on cache miss after #717
## Problem

After #717 fixed `cache.Get` to return `("", nil)` on a Redis key-not-found (miss) instead of propagating `redis.Nil`, `manifestloader.loadFromCache` now hits a spurious JSON parse error on every cache miss.

**Before #717:**
`cache.Get` → `("", redis.Nil)` → `if err != nil` guard in `loadFromCache` returns early → caller logs a clean cache miss.

**After #717:**
`cache.Get` → `("", nil)` → `loadFromCache` skips the error guard → `json.Unmarshal([]byte(""), &doc)` → returns `(nil, "unexpected end of JSON input")` → caller logs:
```
ManifestLoader: cache miss for networkID="...": unexpected end of JSON input
```

## Impact

- Every cold-start cache miss (first load, or after TTL expiry) logs what appears to be a data-corruption error, misleading operators and polluting traces.
- Functionally a no-op — the fallback to the remote registry still fires correctly.
- Latent risk: if `loadFromCache` is ever changed to return `(nil, nil)` as the clean-miss signal (the natural fix), callers at lines 96 and 141 will nil-pointer-panic on `doc.FetchedAt.Format(...)`.

## Fix

Add an explicit empty-string guard in `loadFromCache` before the unmarshal:

```go
func (l *Loader) loadFromCache(ctx context.Context, key string) (*model.ManifestDocument, error) {
    raw, err := l.cache.Get(ctx, key)
    if err != nil {
        return nil, err
    }
    if raw == "" {
        return nil, fmt.Errorf("cache miss") // key not found — normal, not an error
    }
    var doc model.ManifestDocument
    if err := json.Unmarshal([]byte(raw), &doc); err != nil {
        return nil, err
    }
    ...
}
```

## Related

- Introduced by #717
- Affects `pkg/plugin/implementation/manifestloader/manifestloader.go` lines 214–227

nirmalnr:
## Summary

Fixes issue #724, a regression introduced by #717 which changed `cache.Get` to return `("", nil)` on a Redis key-not-found instead of propagating `redis.Nil`. Three related problems are fixed together, along with test coverage improvements and a build-tag fix.

### Problem 1 — Spurious `"unexpected end of JSON input"` on every cache miss
`loadFromCache` had no guard for an empty raw string before calling `json.Unmarshal`. After #717, every cold-start miss hit `json.Unmarshal([]byte(""), ...)` and logged a misleading data-corruption error. Functionally harmless, but noisy and alarming for operators.

**Fix:** return `(nil, nil)` on empty raw; callers updated to check `doc != nil` before treating the result as a hit.

### Problem 2 — Redis write failure caused the entire load to fail
`store()` propagated `cache.Set` errors back to the caller. A Redis outage would cause `GetByNetworkID` / `getByMetadata` to return an error even after successfully fetching and verifying the manifest from the remote registry.

**Fix:** `store()` now logs a `Warn` and returns `nil` on a write failure — the next request gets a cache miss and re-fetches, which is acceptable degraded behaviour.

### Test coverage improvements
- **`manifestloader`**: Updated `mockCache.Get` to return `("", nil)` on miss, matching the real post-#717 cache contract so this class of regression is caught by tests going forward. Added `TestLoadFromCacheOnCacheMissReturnsNilNil` to directly exercise the fixed path.
- **`payloadstore`**: Added `errGetCache` stub for cache error injection. Added `TestExists_CacheErrorPropagated` (verifies `Exists` surfaces real errors to callers) and `TestStore_ContinuesAfterExistsCacheError` (verifies `Store` still persists the entry when duplicate-detection fails due to a cache error).

### Additional: `benchmarks/tools` build failure under `go test ./...`
`parse_results.go` and `generate_report.go` each declare `func main()`, causing a `main redeclared` build error. Added `//go:build ignore` to both — `go test ./...` skips them, `go run file.go` still works unchanged.

## Test plan
- [x] `go test ./pkg/plugin/implementation/manifestloader/...` — all pass, including new `TestLoadFromCacheOnCacheMissReturnsNilNil`
- [x] `go test ./pkg/plugin/implementation/cache/...` — all pass
- [x] `go test ./pkg/plugin/implementation/payloadstore/...` — all pass, including new `TestExists_CacheErrorPropagated` and `TestStore_ContinuesAfterExistsCacheError`
- [x] `go test ./benchmarks/tools/...` — no longer fails with build error
- [x] `go test ./...` — no longer has any failures
- [x] `go run benchmarks/tools/parse_results.go --help` and `go run benchmarks/tools/generate_report.go --help` — tools still run normally

Fixes #724.

### DIFF
--- a/benchmarks/tools/generate_report.go
+++ b/benchmarks/tools/generate_report.go
@@ -1,3 +1,5 @@
+//go:build ignore
+
 // generate_report.go — Fills REPORT_TEMPLATE.md with data from a completed
 // benchmark run and writes BENCHMARK_REPORT.md to the results directory.
 //
@@ -51,8 +53,6 @@
 //	__CACHE_DELTA__       formatted warm-vs-cold delta string
 //	__THROUGHPUT_TABLE__  generated markdown table from throughput_report.csv
 //	__BENCHSTAT_SUMMARY__ raw contents of benchstat_summary.txt
-
-//go:build ignore
 
 package main
 
--- a/benchmarks/tools/generate_report.go
+++ b/benchmarks/tools/generate_report.go
@@ -51,6 +51,9 @@
 //	__CACHE_DELTA__       formatted warm-vs-cold delta string
 //	__THROUGHPUT_TABLE__  generated markdown table from throughput_report.csv
 //	__BENCHSTAT_SUMMARY__ raw contents of benchstat_summary.txt
+
+//go:build ignore
+
 package main
 
 import (
--- a/benchmarks/tools/parse_results.go
+++ b/benchmarks/tools/parse_results.go
@@ -1,3 +1,5 @@
+//go:build ignore
+
 // parse_results.go — Parses raw go test -bench output from the benchmark results
 // directory and produces two CSV files for analysis and reporting.
 //
@@ -11,8 +13,6 @@
 //
 //	latency_report.csv    — per-benchmark mean, p50, p95, p99 latency, allocs
 //	throughput_report.csv — RPS and mean latency at each GOMAXPROCS level from the parallel sweep
-
-//go:build ignore
 
 package main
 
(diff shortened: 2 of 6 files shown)


nirmay (approved):
Approved.

PR #747, opened 2026-05-28
fix: strip module base path in step layer before plugin URL lookup

### LINKED ISSUE #741: bug(router): compound action verbs and bodyless requests break routing for protocol-spec v2 endpoints
## Summary

The router plugin has two failure modes that prevent it from handling the compound/nested action verbs and bodyless (GET/DELETE) endpoints introduced in protocol-specifications-v2 (`core-v2.0.0-lts-rc3`).

---

## Background

Protocol-spec v2 introduces a new family of endpoints beyond the traditional single-word verbs (`search`, `select`, `init`, etc.). The complete set of compound verbs in `core-v2.0.0-lts-rc3` is:

| Path | HTTP Method | Body |
|---|---|---|
| `catalog/publish` | POST | Yes |
| `catalog/on_publish` | POST | Yes |
| `catalog/push` | POST | Yes |
| `catalog/subscription` | POST | Yes |
| `catalog/subscription` | GET | No |
| `catalog/subscription` | DELETE | No |
| `catalog/pull` | POST | Yes |
| `catalog/on_pull` | POST | Yes |
| `catalog/search` | POST | Yes |

---

## Failure Mode 1 — Compound paths are silently truncated

**Location:** [`pkg/plugin/implementation/router/router.go:243`](https://github.com/beckn/beckn-onix/blob/main/pkg/plugin/implementation/router/router.go#L243)

```go
endpoint := path.Base(url.Path)
```

`path.Base` returns only the last segment of the URL path. For any compound verb, the prefix is discarded:

- `path.Base("/catalog/subscription")` → `"subscription"`
- `path.Base("/catalog/search")` → `"search"`

The rules map (populated at config load time, keyed by the full endpoint string) will never match. Every compound verb results in:

```
endpoint 'subscription' is not supported for version 2.x.x in routing config
```

This is unconditional — there is no fallback.

---

## Failure Mode 2 — Body is parsed unconditionally; bodyless requests fail immediately

**Location:** [`pkg/plugin/implementation/router/router.go:224`](https://github.com/beckn/beckn-onix/blob/main/pkg/plugin/implementation/router/router.go#L224)

```go
if err := json.Unmarshal(body, &requestBody); err != nil {
    return nil, fmt.Errorf("error parsing request body: %w", err)
}
```

For GET and DELETE requests, `body` is `nil` or empty. `json.Unmarshal` returns `unexpected end of JSON input` and the router returns an error before reaching any routing logic.

This affects `GET /catalog/subscription` and `DELETE /catalog/subscription`.

---

## Observed vs Expected Behavior

| Request | Expected | Actual |
|---|---|---|
| `POST /catalog/publish` with body | Routed successfully | Error: endpoint `publish` not supported |
| `POST /catalog/search` with body | Routed successfully | Error: endpoint `search` not supported (ambiguous with core `/search`) |
| `GET /catalog/subscription` | Routed successfully | Error: parsing request body |
| `DELETE /catalog/subscription` | Routed successfully | Error: parsing request body |
| `POST /search` with body | Routed successfully | Works (unaffected) |

---

## Related

- `schemav2validator` handles both cases correctly: it uses `strings.TrimPrefix` (not `path.Base`) for path matching and has an explicit bodyless branch before any JSON parsing. See [`schemav2validator.go:122-134`](https://github.com/beckn/beckn-onix/blob/main/pkg/plugin/implementation/schemav2validator/schemav2validator.go#L122-L134).
- Issue [#739](https://github.com/beckn/beckn-onix/issues/739) added bodyless support to `schemav2validator` — the router did not receive a corresponding update.

### LINKED ISSUE #742: fix: strip module base path before plugin URL lookup — schemav2validator bodyless and router compound actions
## Summary

The module base path (e.g. `/bap/receiver/`) was never stripped from the incoming URL before it was passed to plugins. This caused two related failures:

1. **`schemav2validator` bodyless lookup always fails** — `strings.TrimPrefix(path, "/")` produced `bap/receiver/catalog/subscription` instead of `catalog/subscription`, which never matched the `bodylessActions` index built from the OpenAPI spec.
2. **`router` compound action routing always fails** — `path.Base()` fallback returned only the last segment (`subscription`), not the full compound key (`catalog/subscription`), so routing rules were never matched.

## Root cause

Both plugins extracted the action from the raw URL path, which still carried the module mount prefix. Neither had a reliable way to strip it — the `basePath` workaround added in #743 was never wired through `cmd/plugin.go` and therefore dead in production.

## Fix

Move base-path stripping out of plugins and into the **step layer**, which already owns the request-processing pipeline and knows the module path:

- `handler.Config` gets a `BasePath` field (set from `module.Config.Path`, not from YAML)
- `stdHandler` stores it and passes it to `validateSchemaStep` and `addRouteStep` constructors
- A `stripBasePath(u *url.URL, basePath string) *url.URL` helper in `step.go` strips the prefix and re-adds the leading slash before each plugin call
- Plugins receive a clean URL (`/catalog/subscription`) and need no awareness of mount paths

**Router cleanup:** `router.Config.BasePath` and `r.basePath` (the #743 workaround) are removed. `extractEndpoint` is simplified to `strings.TrimPrefix(urlPath, "/")` — correct for all paths now that the step always pre-strips.

## Affected plugins

| Plugin | Issue |
|---|---|
| `schemav2validator` | Bodyless action lookup now receives `/catalog/subscription` → correct key |
| `router` | Compound action routing now receives `/catalog/subscription` → correct endpoint match |

## Not in scope

OPA `extractActionFromPath` has a related compound-path bug (hardcoded 3-segment structure assumption) — tracked separately.

## References

- Fixes #742
- Fixes #741 (router compound action routing)
- Supersedes per-plugin `basePath` config workaround from #743

nirmay:
## Summary

- **Root cause:** The module mount prefix (e.g. \`/bap/receiver/\`) was never stripped from the incoming URL before plugins saw it. Both \`schemav2validator\` and \`router\` extracted actions from the raw path, so compound paths like \`catalog/subscription\` never matched their internal indexes.
- **Fix:** Move base-path stripping into the step layer (\`validateSchemaStep\` and \`addRouteStep\`). \`stripBasePath()\` returns a shallow \`*url.URL\` copy whose \`.Path\` is the clean endpoint action (e.g. \`catalog/subscription\`) with no leading slash — plugins use \`.Path\` directly, no further \`TrimPrefix\` needed.
- **Router cleanup:** Removes the \`Config.BasePath\` / \`Router.basePath\` workaround added in #743 that was never wired through \`cmd/plugin.go\` and was dead in production.

## Changes

| File | Change |
|---|---|
| \`core/module/handler/config.go\` | Add \`BasePath string\` (set from module path) |
| \`core/module/module.go\` | Thread \`module.Config.Path\` → \`handler.Config.BasePath\` |
| \`core/module/handler/stdHandler.go\` | Store \`basePath\`, pass to step constructors |
| \`core/module/handler/step.go\` | \`stripBasePath()\` returns \`*url.URL\` with clean \`.Path\`; both steps pass it to plugins |
| \`pkg/plugin/implementation/router/router.go\` | Remove \`Config.BasePath\`; \`Route\` uses \`reqURL.Path\` directly |
| \`pkg/plugin/implementation/schemav2validator/schemav2validator.go\` | \`Validate\` uses \`reqURL.Path\` for bodyless lookup; nil URL guard |
| \`pkg/plugin/implementation/schemavalidator/schemavalidator.go\` | \`Validate\` uses \`reqURL.Path\` directly |
| \`core/module/handler/step_test.go\` | \`TestStripBasePath_*\` unit tests; wiring tests assert \`.Path\` on recorded \`*url.URL\` |
| \`pkg/plugin/implementation/router/router_test.go\` | \`Route\` calls pass \`&url.URL{Path: action}\` |
| \`pkg/plugin/implementation/schemav2validator/schemav2validator_test.go\` | Bodyless tests pass \`&url.URL{Path: action}\` or \`nil\` |
| \`pkg/plugin/implementation/schemavalidator/schemavalidator_test.go\` | \`Validate\` calls pass \`&url.URL{Path: action}\` |

## Test plan

- [x] \`go test ./core/...\` — all pass
- [x] \`go test ./pkg/plugin/implementation/router/...\` — all pass including \`TestRouteCompoundEndpointSuccess\` and \`TestRouteBodyless\`
- [x] \`go test ./pkg/plugin/implementation/schemav2validator/...\` — all pass
- [x] \`go test ./pkg/plugin/implementation/schemavalidator/...\` — all pass
- [x] Full suite \`go test ./...\` (excl. plugin cmd) — all pass

## Notes

- **Plugin interfaces unchanged:** \`SchemaValidator.Validate\` and \`Router.Route\` signatures remain \`(ctx, reqURL *url.URL, ...)\`. The contract clarification is that \`reqURL.Path\` now carries the clean endpoint action rather than the raw mount-prefixed path.
- OPA \`extractActionFromPath\` has a related compound-path bug — not in scope here, tracked separately.

Fixes #742, fixes #741

### DIFF
--- a/core/module/handler/config.go
+++ b/core/module/handler/config.go
@@ -121,4 +121,8 @@ type Config struct {
 	Role             model.Role
 	SubscriberID     string           `yaml:"subscriberId"`
 	HttpClientConfig HttpClientConfig `yaml:"httpClientConfig"`
+	// BasePath is the HTTP path prefix at which this module is mounted (e.g.
+	// "/bap/receiver/"). Set by the module layer from module.Config.Path; not
+	// read from YAML. Steps use it to strip the prefix before calling plugins.
+	BasePath string `yaml:"-"`
 }
--- a/core/module/handler/stdHandler.go
+++ b/core/module/handler/stdHandler.go
@@ -49,6 +49,7 @@ type stdHandler struct {
 	ackSigner    *ackSignerStep
 	SubscriberID string
 	role         model.Role
+	basePath     string
 	httpClient   *http.Client
 	moduleName   string
 }
@@ -88,6 +89,7 @@ func NewStdHandler(ctx context.Context, mgr PluginManager, cfg *Config, moduleNa
 		responseSteps: []definition.ResponseStep{},
 		SubscriberID:  cfg.SubscriberID,
 		role:          cfg.Role,
+		basePath:      cfg.BasePath,
 		moduleName:    moduleName,
 	}
 	// Initialize plugins.
@@ -588,9 +590,9 @@ func (h *stdHandler) initSteps(ctx context.Context, mgr PluginManager, cfg *Conf
 		case "validateSign":
 			s, err = newValidateSignStep(h.signValidator, h.km, h.payloadStore)
 		case "validateSchema":
-			s, err = newValidateSchemaStep(h.schemaValidator)
+			s, err = newValidateSchemaStep(h.schemaValidator, h.basePath)
 		case "addRoute":
-			s, err = newAddRouteStep(h.router)
+			s, err = newAddRouteStep(h.router, h.basePath)
 		case "checkPolicy":
 			s, err = newCheckPolicyStep(h.policyChecker)
 		case "transformPayload":
--- a/core/module/handler/step.go
+++ b/core/module/handler/step.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"strings"
 	"time"
 
@@ -359,14 +360,31 @@ func parseHeader(header string) (*authHeader, error) {
 	}, nil
 }
 
+// stripBasePath returns a shallow copy of u with basePath stripped from the
+// front of its Path, then re-adding a leading slash so the URL remains valid.
+// When basePath is empty or u is nil the original pointer is returned unchanged.
+func stripBasePath(u *url.URL, basePath string) *url.URL {
+	if basePath == "" || u == nil {
+		return u
+	}
+	stripped := *u
+	p := strings.TrimPrefix(u.Path, basePath)
+	if !strings.HasPrefix(p, "/") {
+		p = "/" + p
+	}
+	stripped.Path = p
+	return &stripped
+}
+
 // validateSchemaStep represents the schema validation step.
 type validateSchemaStep struct {
 	validator definition.SchemaValidator
+	basePath  string
 	metrics   *HandlerMetrics
 }
 
 // newValidateSchemaStep creates and returns the validateSchema step after validation.
-func newValidateSchemaStep(schemaValidator definition.SchemaValidator) (definition.Step, error) {
+func newValidateSchemaStep(schemaValidator definition.SchemaValidator, basePath string) (definition.Step, error) {
 	if schemaValidator == nil {
 		return nil, fmt.Errorf("invalid config: SchemaValidator plugin not configured")
 	}
@@ -374,13 +392,14 @@ func newValidateSchemaStep(schemaValidator definition.SchemaValidator) (definiti
 	metrics, _ := GetHandlerMetrics(context.Background())
 	return &validateSchemaStep{
 		validator: schemaValidator,
+		basePath:  basePath,
 		metrics:   metrics,
 	}, nil
 }
 
 // Run executes the schema validation step.
 func (s *validateSchemaStep) Run(ctx *model.StepContext) error {
-	err := s.validator.Validate(ctx, ctx.Request.URL, ctx.Body)
+	err := s.validator.Validate(ctx, stripBasePath(ctx.Request.URL, s.basePath), ctx.Body)
 	if err != nil {
 		err = fmt.Errorf("schema validation failed: %w", err)
 	}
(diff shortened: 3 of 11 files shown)


nirmay:
## Code Review: PR #747 — fix: strip module base path in step layer before plugin URL lookup

### Summary

Moves base-path stripping from inside individual plugins (where it was either broken or a dead wiring) into the step layer via a new `stripBasePath()` helper, then propagates a clean endpoint action string to plugins. The router cleanup (removing the `#743` `BasePath`/`extractEndpoint` dead code) is the right call. Architecture is sound and the test coverage is solid.

---

### Critical Issues

None.

---


### What Looks Good

- **Clean separation**: step layer owns mount-prefix stripping; plugins receive a clean action string and are path-agnostic.
- **`stripBasePath` is pure and non-mutating**: the shallow copy + `DoesNotMutateURL` test confirm the original URL is never touched.
- **Dead code removal**: `Config.BasePath`, `Router.basePath`, `extractEndpoint`, and `setupRouterWithBasePath` are all cleanly gone. The router simplified from `~30 lines → ~10 lines` in `Route()`.
- **Interface contract clarified, not changed**: plugin signatures are identical; the semantic change (`reqURL.Path` = action, not raw path) is documented in each plugin's `Validate`/`Route` docstring.
- **Test depth**: 8 `stripBasePath` unit tests + wiring tests for both steps cover nil, empty basePath, trailing-slash, no-trailing-slash, compound, query-string, and mutation scenarios.
- **`TestRouteCompoundEndpointSuccess`** no longer needs `setupRouterWithBasePath` — the router fixture is simplified to the standard `setupRouter`, which is correct.
- **Commit history** is clean (fix → tests → refactor), reads well as a stand-alone narrative.


---

### Verdict: **Approve**

The fix is correct, the dead code removal is clean, and the test coverage is thorough.

PR #750, opened 2026-05-28
fix(schemav2validator): reject empty-body POST before schema validation

### LINKED ISSUE #740: refactor(schemavalidator): pass *model.StepContext to SchemaValidator.Validate instead of (url, payload)
## Background

PR #739 (fixes #735, #737) implements bodyless GET/DELETE handling in `schemav2validator` using Option A — a path-only map inside the plugin. A known limitation was documented in that PR: `bodylessActions` is keyed by path only, not by HTTP method, because the current `SchemaValidator` interface carries no method information.

This ticket tracks the long-term fix: change the interface to pass `*model.StepContext`, which carries the full `*http.Request` (and therefore the HTTP method), `Body`, `MessageID`, `ProtocolVersion`, and other per-request data.

## Current interface

```go
// pkg/plugin/definition/schemaValidator.go
type SchemaValidator interface {
    Validate(ctx context.Context, url *url.URL, payload []byte) error
}
```

Call site (`core/module/handler/step.go` line 383):
```go
err := s.validator.Validate(ctx, ctx.Request.URL, ctx.Body)
```

## Proposed interface

```go
type SchemaValidator interface {
    Validate(sc *model.StepContext) error
}
```

Call site becomes:
```go
err := s.validator.Validate(ctx)
```

`model.StepContext` already carries everything the current signature provides (`Request.URL` → replaces `url`, `Body` → replaces `payload`) plus what's missing today (`Request.Method`, `MessageID`, `ProtocolVersion`, `Role`).

## Scope

This is a **breaking interface change**. Every `SchemaValidator` implementor must be updated.

| File | Change | Notes |
|---|---|---|
| `pkg/plugin/definition/schemaValidator.go` | Signature change | Breaking — new import of `pkg/model` |
| `core/module/handler/step.go` | Call site update | `Validate(ctx, ctx.Request.URL, ctx.Body)` → `Validate(ctx)` |
| `pkg/plugin/implementation/schemav2validator/schemav2validator.go` | Primary beneficiary | Use `sc.Request.Method` for method-aware bodyless check; replace `reqURL`/`data` with `sc.Request.URL`/`sc.Body` |
| `pkg/plugin/implementation/schemav2validator/cmd/plugin.go` | Mechanical | Follow-through on impl signature |
| `pkg/plugin/implementation/schemavalidator/schemavalidator.go` | Mechanical — no behavior change | Update signature; replace `url`→`sc.Request.URL`, `data`→`sc.Body`; v1 has no use for method |
| `pkg/plugin/implementation/schemavalidator/cmd/plugin.go` | Mechanical | Follow-through on impl signature |

**Not in scope:** `SchemaValidatorProvider` interface, `pkg/plugin/manager.go`, `stdHandler.go`, all other plugins.

## Architectural note

This change introduces a dependency from `pkg/plugin/definition` → `pkg/model`. Today `definition` only imports stdlib (`context`, `net/url`). After this change, external plugin authors building against the interface must also take `pkg/model` as a dependency. This is intentional — `model.StepContext` is already the canonical per-request envelope for the entire ONIX pipeline, and aligning the plugin interface with it is the right direction.

## Acceptance criteria

- [ ] `SchemaValidator.Validate` accepts `*model.StepContext`
- [ ] `schemav2validator` uses `sc.Request.Method` to distinguish GET/DELETE from POST in the bodyless check — the known-limitation test case in `TestValidate_BodylessRequest` (`wantErr: false` for empty-body POST) flips to `wantErr: true`
- [ ] `schemavalidator` v1 compiles and passes all existing tests with the new signature (mechanical change, no behavior delta)
- [ ] `step.go` call site updated
- [ ] PR description notes this as a breaking change for external plugin authors

## References

- PR #739 (introduces the known limitation this fixes)
- Fixes known limitation documented in `schemav2validator.go` bodyless branch comment

nirmay:
## Summary

- Adds a method guard in `validateSchemaStep.Run`: a POST request with no body is rejected with a 400 `BadReqErr` before `Validate` is ever called
- Closes the known limitation in `schemav2validator` where an empty-body POST to a bodyless-indexed path (e.g. `/catalog/subscription`) bypassed body schema validation
- All `SchemaValidator` plugin implementations benefit automatically — no interface change, no context injection, no model change

## Approach

The HTTP method is available on `ctx.Request` in the step layer. Rather than threading it into the plugin via interface changes or `context.WithValue`, the step enforces the HTTP invariant ("POST requires a body") upstream. The plugin remains method-agnostic by design.

Checked against the Beckn 2.0.0 spec (`api/v2.0.0/beckn.yaml`): `/catalog/subscription` is the only path with both a bodyless GET/DELETE and a body-bearing POST — confirming the fix targets a real spec case.

## Test plan

- [ ] `TestValidateSchemaStep_Run_EmptyBodyPost_ReturnsError` — empty-body POST returns error; validator mock not called
- [ ] Existing `TestValidateSchemaStep_Run_ExtractsAction` and `TestValidateSchemaStep_Run_NoBasePath_UsesRawAction` still pass
- [ ] `TestValidate_BodylessRequest` in `schemav2validator` — all cases pass; known-limitation label removed

Closes #740

### DIFF
--- a/core/module/handler/step.go
+++ b/core/module/handler/step.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"net/url"
 	"strings"
 	"time"
@@ -397,9 +398,14 @@ func newValidateSchemaStep(schemaValidator definition.SchemaValidator, basePath 
 
 // Run executes the schema validation step.
 func (s *validateSchemaStep) Run(ctx *model.StepContext) error {
-	err := s.validator.Validate(ctx, stripBasePath(ctx.Request.URL, s.basePath), ctx.Body)
-	if err != nil {
-		err = fmt.Errorf("schema validation failed: %w", err)
+	var err error
+	if len(ctx.Body) == 0 && ctx.Request.Method == http.MethodPost {
+		err = fmt.Errorf("schema validation failed: %w", model.NewBadReqErr(fmt.Errorf("POST request requires a body")))
+	} else {
+		err = s.validator.Validate(ctx, stripBasePath(ctx.Request.URL, s.basePath), ctx.Body)
+		if err != nil {
+			err = fmt.Errorf("schema validation failed: %w", err)
+		}
 	}
 	s.recordMetrics(ctx, err)
 	return err
--- a/core/module/handler/step_test.go
+++ b/core/module/handler/step_test.go
@@ -595,6 +595,24 @@ func TestValidateSchemaStep_Run_NoBasePath_UsesRawAction(t *testing.T) {
 	}
 }
 
+func TestValidateSchemaStep_Run_EmptyBodyPost_ReturnsError(t *testing.T) {
+	mv := &mockRecordingValidator{}
+	step, _ := newValidateSchemaStep(mv, "")
+	req, _ := http.NewRequest(http.MethodPost, "http://localhost/catalog/subscription", nil)
+	ctx := &model.StepContext{
+		Context:    context.Background(),
+		Request:    req,
+		Body:       []byte{},
+		RespHeader: http.Header{},
+	}
+	if err := step.Run(ctx); err == nil {
+		t.Fatal("expected error for empty-body POST, got nil")
+	}
+	if mv.gotURL != nil {
+		t.Error("Validate should not have been called for empty-body POST")
+	}
+}
+
 // ---------------------------------------------------------------------------
 // addRouteStep constructor + stripBasePath wiring tests
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


nirmay:
### Summary
The test regression from the first review is resolved correctly. errors.As properly unwraps through the "schema validation failed: %w" wrapper to assert the underlying *model.BadReqErr, so a future type change from 400 to 500 would be caught. All three assertions in the test now cover what matters: an error is returned, it carries the right type, and the plugin was never invoked.

### Issues
None.

### What Looks Good
errors.As(err, &badReq) is the right call here — not a string match, not a type cast, but a proper chain traversal that survives the wrapping layer.
Three orthogonal assertions in the test: non-nil error, correct error type, validator not called. Collectively they pin down the full behaviour of the guard.
schemav2validator.go comment accurately captures the new invariant without overstating it.
Plugin test renamed to document intent rather than limitation — that's the right framing now that the upstream fix is in place.
### Verdict
Approve — ready to merge.

PR #754, opened 2026-05-28
fix(opapolicychecker): remove actions allowlist and extractActionFromPath

### LINKED ISSUE #748: fix(opapolicychecker): extractActionFromPath hardcodes 3-segment path, returns "" for compound actions
## Summary

`extractActionFromPath` in `pkg/plugin/implementation/opapolicychecker/enforcer.go:779` only recognises a strict 3-segment URL shape (`/<role>/<direction>/<action>`).  For compound actions such as `catalog/subscription` the path arrives as `/<role>/<direction>/catalog/subscription` (4 segments), the guard `len(parts) == 3` fails, and the function returns `""`.  This silently drops the action field from the OPA input document, making any policy that keys on `action` evaluate against an empty string.

## Current code

```go
func extractActionFromPath(urlPath string) string {
    parts := strings.FieldsFunc(strings.Trim(urlPath, "/"), func(r rune) bool { return r == '/' })
    if len(parts) == 3 && isBecknDirection(parts[1]) && parts[2] != "" {
        return parts[2]
    }
    return ""
}
```

Path `/bpp/caller/catalog/subscription` → `parts` has 4 elements → returns `""`.

## Root cause

The function was written assuming actions are always a single path segment.  Beckn 2.0 introduced compound action paths (e.g. `catalog/subscription`, `catalog/fulfillments`), which break that assumption.

## Why PR #747 changes the fix approach

PR #747 ([c940874](https://github.com/beckn/beckn-onix/commit/c940874)) moved base-path stripping into the step layer (`core/module/handler/step.go`).  By the time a plugin receives `req.URL`, the module mount prefix (`/<role>/<direction>/`) has already been stripped.  The URL the OPA enforcer sees is now a clean path — `catalog/subscription`, `search`, etc. — with no leading slash.

This means `extractActionFromPath` no longer needs to peel off the role/direction prefix at all.  The entire heuristic is now unnecessary complexity.

## Proposed fix

Replace `extractActionFromPath` with a direct read of the already-cleaned `req.URL.Path`, mirroring what `router` and `schemav2validator` do after #747:

```go
// enforcer.go — inside parsedRequestContext population
action := ctx.Request.URL.Path   // already stripped to "search" or "catalog/subscription" by step layer
```

Remove `extractActionFromPath` entirely (or keep it only for unit-test backwards compat and mark it deprecated).  Update the call site at line 687 and the tests at lines 778 and 1803.

## Acceptance criteria

- [ ] OPA input document contains `action = "catalog/subscription"` when the request path is `catalog/subscription`
- [ ] Existing single-segment actions (`search`, `confirm`, etc.) continue to work
- [ ] `extractActionFromPath` is removed or clearly scoped to tests only
- [ ] Unit tests cover compound-action paths

## Related

- Fixes the OPA half of the same class of bug addressed by #747 (fixes #741, #742)
- See also issue #658 (action metric label uses URL path instead of Beckn context.action) — both issues stem from the same incorrect path-parsing pattern

nirmalnr:
Closes #748

### Problem

`extractActionFromPath` only handled 3-segment URL paths (`/<role>/<direction>/<action>`), returning `""` for compound actions like `catalog/subscription`. This caused the `IsActionEnabled` check to fall back to the body action — but more fundamentally, the `actions` allowlist itself is a security hole: an NP can omit any action from their local config to bypass NFO-mandated policy enforcement for that action entirely.

### Fix

Remove the `actions` config and the code that depends on it. OPA policy evaluation now runs for every action. Action-specific filtering belongs in Rego rules via `input.context.action` guards, not in the plugin config.

- Removed `Actions` field from `PolicyConfig` and its config parsing
- Removed `IsActionEnabled` method
- Removed `extractActionFromPath` and `isBecknDirection` (only existed to serve the allowlist)
- `CheckPolicy` no longer reads the URL path; action for debug logging comes from `reqCtx.Action` (request body)
- Updated log messages to drop `actions=` field
- Removed `actions:` from all test configs; removed tests for deleted behaviour
- Added `TestEnforcer_CompoundActionNotFiltered` to verify compound actions reach OPA evaluation
- Updated README: removed `actions` from config table, flowchart, recipe section, and troubleshooting entry

### DIFF
--- a/pkg/plugin/implementation/opapolicychecker/README.md
+++ b/pkg/plugin/implementation/opapolicychecker/README.md
@@ -59,14 +59,11 @@ Both teams share a single configuration contract: the **network policy config fi
          │     │    → skip evaluation (allow)          │    │
          │     └──────────────────────────────────────┘    │
          │           │                                      │
-         │  ③ Check action filter (if actions: set)        │
-         │     not in list? → skip evaluation (allow)      │
-         │           │                                      │
-         │  ④ Evaluate OPA query                           │
+         │  ③ Evaluate OPA query                           │
          │     input        = full Beckn message body       │
          │     data.config  = adapter plugin config keys    │
          │           │                                      │
-         │  ⑤ Interpret result                             │
+         │  ④ Interpret result                             │
          │     {"valid": true,  "violations": []}  → allow │
          │     {"valid": false, "violations": […]} → 400   │
          │     empty / undefined result            → 400   │
@@ -97,7 +94,7 @@ If the OPA query returns an undefined or empty result, the plugin treats it as a
 
 `opapolicychecker` is a module-level plugin compiled as a `.so` and loaded by the plugin manager. It runs as one step in a module's `steps` list alongside the schema validator, signer, and router. The plugin ID (`opapolicychecker`, used in `id:`) and the step name (`checkPolicy`, used in `steps:` and as the YAML config key) are distinct — both must appear in the module config.
 
-> **Migrating from an older config format.** Previous versions accepted top-level keys such as `type`, `location`, `query`, `actions`, and `refreshIntervalSeconds` directly in the plugin config. These keys are no longer used as plugin parameters — unrecognised keys are forwarded to Rego as `data.config.<key>` rather than silently dropped, so leftover keys from the old format will appear as Rego data values instead of configuring the plugin.
+> **Migrating from an older config format.** Previous versions accepted top-level keys such as `type`, `location`, `query`, and `refreshIntervalSeconds` directly in the plugin config. These keys are no longer used as plugin parameters — unrecognised keys are forwarded to Rego as `data.config.<key>` rather than silently dropped, so leftover keys from the old format will appear as Rego data values instead of configuring the plugin.
 
 ```yaml
 manifestLoader:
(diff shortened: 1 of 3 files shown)


nirmay:
## Code Review

The security motivation is correct and well-explained — removing the `actions` allowlist closes a real bypass vector where an NP could opt out of NFO-mandated policy enforcement by simply omitting actions from their config. The deletion is clean, `reqCtx.Action` is strictly better than URL-path parsing, and test coverage is complete.

Two things worth addressing before merge:

### Breaking change needs a louder callout

Any live deployment that used `actions:` to limit evaluation scope will now have every action hit OPA. A policy with no action guard will evaluate (and potentially block) traffic it never saw before. This should be annotated as a breaking change — either `BREAKING CHANGE:` in the commit/PR body footer or a `!` in the conventional-commit title (`fix(opapolicychecker)!: ...`) so tooling and changelogs flag it correctly.

### Stale `actions:` keys silently become Rego data

The README notes that unrecognised keys are forwarded as `data.config.<key>`, but doesn't spell out the consequence clearly: a leftover `actions: confirm,init` becomes `data.config.actions = "confirm,init"` inside OPA — a string, not a filter. Operators who don't read the migration note closely will get confusing behaviour. Worth adding one sentence: *"A leftover `actions:` key will appear as `data.config.actions` in Rego instead of restricting evaluation — remove it to avoid confusing policy authors."*

---

Otherwise this is good to go. The core change is sound and the new `TestEnforcer_CompoundActionNotFiltered` test is exactly the right regression coverage.

Issue #749, opened 2026-05-28
[Proto-v2.0] fix: digest prefix in signing string should be `BLAKE2b-512=` not `BLAKE-512=`


nirmalnr:
### Summary

The signing string digest line uses the label `BLAKE-512=` but the NFH-007 specification requires `BLAKE2b-512=`. This causes a wire-incompatibility with any external peer that implements the spec correctly.

---

### What the spec says

**NFH-007 §3.1 — Body Digest:**
> `digest = "BLAKE2b-512=" + Base64( BLAKE2b-512( UTF-8(body) ) )`

The prefix `BLAKE2b-512=` is a literal string prepended to the Base64-encoded hash. The conformance requirement is stated in **CON-004-04**:
> The body digest MUST be computed using BLAKE2b-512 and encoded as `BLAKE2b-512={Base64}`.

---

### What the code does

The hash algorithm itself is correct — both files import `golang.org/x/crypto/blake2b` and call `blake2b.New512(nil)`. The bug is only in the label string written into the signing string.

**`pkg/plugin/implementation/signer/signer.go:41`**
```go
return fmt.Sprintf("(created): %d\n(expires): %d\ndigest: BLAKE-512=%s", createdAt, expiresAt, digestB64), nil
```

**`pkg/plugin/implementation/signvalidator/signvalidator.go:146`**
```go
return fmt.Sprintf("(created): %d\n(expires): %d\ndigest: BLAKE-512=%s", createdTimestamp, expiredTimestamp, digestB64)
```

---

### Impact

Any peer that implements NFH-007 correctly will produce a signing string with `digest: BLAKE2b-512=<hash>`. Onix produces `digest: BLAKE-512=<hash>`. Because the signing string is what is signed and verified, the two strings will never match, causing all cross-implementation signature verifications to fail.

---

### Backwards Compatibility Concern

All currently deployed ONIX nodes use `BLAKE-512=`. A node that upgrades and switches to `BLAKE2b-512=` will be unable to communicate with any unupgraded node. Three approaches are possible:

**Option 1 — Protocol-version gate**
[...]

PR #756, opened 2026-05-28
fix(handler): use 4-line signing string for solicited callbacks (#752)

### LINKED ISSUE #752: [Proto-v2.0] fix: solicited callback signing and verification use 3-line signing string instead of 4-line callback signing string
### Summary

NFH-007 requires solicited callbacks to be signed using a 4-line signing string that cryptographically binds the PN's callback to the CN's original request via a `request-signature` line. The current implementation signs solicited callbacks with the standard 3-line string and bolts `request-signature` on as a plain header attribute — meaning the chain is never cryptographically proven. The two sides are internally consistent with each other today, but both deviate from the spec and either side will fail when talking to a compliant third-party peer.

---

### What the spec says

**NFH-007 §3.3 — Callback Signing String** (solicited callbacks only):
```
(created): {created_unix_ts}
(expires): {expires_unix_ts}
digest: BLAKE2b-512={callback_body_digest}
request-signature: {cn_original_signature_value}
```
The `request-signature` line is the 4th line of the **signing string** — the data that is actually signed. It is not a header attribute.

**NFH-007 §4 — Authorization header format:**
```
Authorization: Signature keyId="...",algorithm="ed25519",created="...",expires="...",headers="(created) (expires) digest request-signature",signature="..."
```
Six attributes only. No `request-signature` attribute is defined.

**NFH-007 §7 Step 5 & Step 7 — CN verification of solicited callbacks:**
> Reconstruct the callback signing string (§3.3) using the `created`, `expires`, body digest, **and the CN's own stored outbound signature** as the `request-signature` value. Verify the Ed25519 signature against that 4-line string. If the signature verifies, the chain is proven — the PN demonstrably included the correct CN signature in the signed data.

---

### Comparison across all cases

| Case | Signing string | Spec | Implementation |
|---|---|---|---|
| CN request / Provider-initiated notification | 3-line | `(created)` `(expires)` `digest` | ✓ Correct |
| **Solicited callback (PN → CN)** | **4-line** | `(created)` `(expires)` `digest` `request-signature` | ✗ Signs 3-line; `request-signature` added as plain header attribute |
| Synchronous ACK response | 4-line | `(created)` `(expires)` `digest` `request-signature` | ✓ Correct structure via `SignAck` / `ValidateAck` |

---

### What the code does

**PN side — signing (`core/module/handler/step.go:66, 77`)**

`signStep.Run` calls `s.signer.Sign` for all outgoing messages including solicited callbacks. `Sign` always produces a 3-line signing string. The CN's original signature is looked up from the payload store and appended as a **7th header attribute** — it is never included in the data that is actually signed:

```go
// step.go — same Sign call for both regular requests and solicited callbacks
sign, err := s.signer.Sign(signerCtx, ctx.Body, keySet.SigningPrivate, createdAt, validTill)

// generateAuthHeader — request-signature added as a header attribute, not in the signing string
",headers=\"(created) (expires) digest request-signature\",signature=\"%s\",request-signature=\"%s\""
```

**CN side — verification (`core/module/handler/step.go:175–285`)**

`validateHeaders` calls `validator.Validate`, which reconstructs a **3-line** signing string and verifies against that. A separate `validateRequestSignatureChain` then does a plain string comparison of the `request-signature` header attribute against the stored outbound signature:

```go
// validateHeaders — always verifies against 3-line string
s.validator.Validate(ctx, ctx.Body, value, signingPublicKey)

// validateRequestSignatureChain — separate string comparison, not cryptographic
reqSig := extractRequestSignatureValue(authHeader)  // reads from header attribute
if reqSig != storedEntry.Signature {                 // plain string comparison
    return model.NewSignValidationErr(...)
}
```

---

### Why this works today but breaks with compliant peers

The two deviations cancel each other out when both sides are Onix:

- Onix PN signs 3-line → Onix CN verifies 3-line → ✓ passes
- Onix PN puts `request-signature` in header attribute → Onix CN reads it from there → ✓ chain check passes

It breaks in both directions with a compliant peer:

- **Compliant PN → Onix CN:** Compliant PN signs 4-line string. Onix CN reconstructs 3-line string for verification → **signature mismatch → 401**
- **Onix PN → Compliant CN:** Onix PN emits `request-signature` as a header attribute. Compliant CN uses its stored outbound signature to reconstruct the 4-line string and verify. Because Onix signed only 3 lines, the 4-line verification **fails → 401**

---

### Fix

**PN side:**
- In `signStep.Run`, detect solicited callbacks (action starts with `on_`) and call `signer.SignAck` instead of `signer.Sign`, passing the stored CN signature. This includes `request-signature` in the actual signed data.
- Remove `request-signature="..."` as a header attribute from `generateAuthHeader`.

**CN side:**
- In `validateHeaders`, detect solicited callbacks (check if `headers` attribute contains `request-signature`) and use a 4-line verification path instead of `validator.Validate`, passing the stored outbound signature from the payload store as the `request-signature` value. This reconstructs the correct 4-line string for cryptographic verification, making the chain check implicit in a successful signature verification.
- Remove the separate `validateRequestSignatureChain` step.
- This requires the public key lookup and the payload store lookup to happen together, merging what are currently two separate steps.

**Note on `SignAck` / `ValidateAck`:**
`SignAck` and `ValidateAck` implement the right 4-line structure and could be reused here — the callback signing string (§3.3) and the ACK response signing string (§3.4) are structurally identical, differing only in which body and which signature value is used. However two things should be noted before reusing them directly:
1. Both methods currently have the `BLAKE-512=` digest prefix bug (tracked separately in #749) — they are structurally correct but not yet fully spec-compliant.
2. The names `SignAck` / `ValidateAck` are oriented around synchronous responses. Rather than reusing them by name, the better approach is to extract the shared 4-line signing string construction into a common internal function and have both the ACK response path and the solicited callback path call it explicitly — making the intent clear at each call site.

nirmay:
## Summary

Fixes #752.

Solicited callbacks (`on_*` actions) were signed with the standard 3-line signing string and carried the CN's original signature only as a non-spec 7th `Authorization` header attribute. Both sides were internally consistent, but the implementation deviated from NFH-004 §3.3 and would fail to interoperate with any spec-compliant peer in either direction.

### Root cause
`signStep.Run` called `signer.Sign` for all outgoing messages, producing a 3-line signing string. The CN's original signature was looked up from the PayloadStore but only appended as `request-signature="..."` in the header — it was never included in the data that was actually signed. On the receive side, `validateHeaders` always called `validator.Validate` (3-line reconstruction), and a separate `validateRequestSignatureChain` did a plain string comparison against the non-spec header attribute rather than cryptographic verification.

### Fix

**Send side (`signStep.Run`):**
- Looks up the CN's stored outbound signature _before_ the signer span.
- Calls `signer.SignAck` (4-line signing string, NFH-004 §3.3) for solicited callbacks; `signer.Sign` (3-line) for all other messages.
- Degrades gracefully to `Sign` when PayloadStore has no entry for the message ID (non-fatal, same as existing `lookupRequestSignature` behaviour).
- `generateAuthHeader` no longer emits the non-spec `request-signature="..."` attribute — Authorization header is now spec-compliant 6 attributes.

**Receive side (`validateHeaders`):**
- New `lookupCallbackRequestSig` helper: for v2 solicited callbacks (v2 protocol, `headers` attribute declares `request-signature`, PayloadStore configured), fetches the stored CN outbound signature. Degrades to empty string (3-line path) when store is nil or protocol is pre-v2.
- `validate` now takes a `requestSig` parameter; dispatches to `validator.ValidateAck` (4-line) when non-empty, `validator.Validate` (3-line) otherwise. Both branches share a single error format.
- `validateRequestSignatureChain` removed — chain integrity is now proven by the Ed25519 signature itself.
- `extractRequestSignatureValue` removed — its only caller was the removed chain check.

### Behaviour table after this fix

| Case | Signing string | Spec | Before | After |
|---|---|---|---|---|
| CN request / PN-initiated | 3-line | `(created) (expires) digest` | ✓ | ✓ |
| **Solicited callback** | **4-line** | `(created) (expires) digest request-signature` | ✗ 3-line signed, non-spec 7th attr | ✓ |
| Synchronous ACK response | 4-line | `(created) (expires) digest request-signature` | ✓ | ✓ |

### Out of scope
- `BLAKE-512=` digest prefix (spec says `BLAKE2b-512=`) — tracked in #749.
- `SignAck`/`ValidateAck` naming refactor mentioned in #752 — no functional impact, separate PR.

## Test plan

- [x] `TestSignStep_Run_SolicitedCallback_UsesSignAck` — `SignAck` called, not `Sign`, for `on_*` actions; no `request-signature` attribute in emitted header
- [x] `TestSignStep_Run_NonCallback_UsesSign` — `Sign` called for regular actions
- [x] `TestSignStep_Run_SolicitedCallback_NoStoredEntry_FallsBackToSign` — missing store entry degrades to `Sign` without error
- [x] `TestSignStep_Run_SignAckError_ReturnsError` — `SignAck` failure propagated correctly
- [x] `TestValidateHeaders_SolicitedCallback_UsesValidateAck` — `ValidateAck` dispatched for v2 solicited callbacks with stored entry
- [x] `TestValidateHeaders_ProviderInitiated_UsesValidate` — `Validate` dispatched for provider-initiated callbacks
- [x] `TestValidateHeaders_SolicitedCallback_NilStore_FallsBackToValidate` — degrades gracefully when PayloadStore not configured
- [x] `TestValidateHeaders_SolicitedCallback_PreV2_FallsBackToValidate` — no 4-line path for pre-v2 protocol
- [x] `TestValidateHeaders_SolicitedCallback_NoEntry_ReturnsError` — unsolicited callback rejected
- [x] `TestValidateHeaders_SolicitedCallback_StoreError_ReturnsError` — store error propagated
- [x] Full suite: `go test ./...` — all green, no regressions

### DIFF
--- a/core/module/handler/responsestep_test.go
+++ b/core/module/handler/responsestep_test.go
@@ -441,13 +441,16 @@ func TestNack_1(t *testing.T) {
 
 // mockSigner satisfies definition.Signer for testing.
 type mockSigner struct {
+	signCalled    bool
 	signAckCalled bool
 	signAckErr    error
-	returnSig     string
+	returnSig     string // returned by SignAck
+	returnSignSig string // returned by Sign (default "")
 }
 
 func (m *mockSigner) Sign(_ context.Context, _ []byte, _ string, _, _ int64) (string, error) {
-	return "", nil
+	m.signCalled = true
+	return m.returnSignSig, nil
 }
 
 func (m *mockSigner) SignAck(_ context.Context, _ []byte, _ string, _ string, _, _ int64) (string, error) {
--- a/core/module/handler/step.go
+++ b/core/module/handler/step.go
@@ -27,10 +27,10 @@ type signStep struct {
 }
 
 // newSignStep initializes and returns a new signing step.
-// payloadStore may be nil; when non-nil and the outgoing action is a callback
-// (prefixed with "on_"), signStep looks up the originating request's signature
-// in the store and includes it as request-signature="..." in the Authorization
-// header, enabling the peer to verify the solicited-callback chain (NFH-004 §6).
+// payloadStore may be nil; when non-nil and the outgoing action is a solicited
+// callback (prefixed with "on_"), signStep looks up the originating request's
+// signature and signs with SignAck (4-line signing string, NFH-004 §3.3),
+// cryptographically binding the callback to the CN's original request.
 func newSignStep(signer definition.Signer, km definition.KeyManager, payloadStore definition.PayloadStore) (definition.Step, error) {
 	if signer == nil {
 		return nil, fmt.Errorf("invalid config: Signer plugin not configured")
@@ -62,22 +62,28 @@ func (s *signStep) Run(ctx *model.StepContext) error {
 	}
 
 	{
+		// Look up the CN's original signature before signing so we can choose
+		// Sign (3-line) vs SignAck (4-line) based on whether this is a solicited
+		// callback (NFH-004 §3.3).
+		requestSig := s.lookupRequestSignature(ctx)
+
 		signerCtx, signerSpan := tracer.Start(ctx.Context, "sign")
 		createdAt := time.Now().Unix()
 		validTill := time.Now().Add(5 * time.Minute).Unix()
-		sign, err := s.signer.Sign(signerCtx, ctx.Body, keySet.SigningPrivate, createdAt, validTill)
+		var sign string
+		var err error
+		if requestSig != "" {
+			sign, err = s.signer.SignAck(signerCtx, ctx.Body, requestSig, keySet.SigningPrivate, createdAt, validTill)
+		} else {
+			sign, err = s.signer.Sign(signerCtx, ctx.Body, keySet.SigningPrivate, createdAt, validTill)
+		}
 		signerSpan.End()
 		if err != nil {
 			return fmt.Errorf("failed to sign request: %w", err)
 		}
 
-		// For solicited callbacks (action starts with "on_") look up the original
-		// request's signature in the store and include it as request-signature="..."
-		// so the receiving peer can verify the chain (NFH-004 §6).
-		requestSig := s.lookupRequestSignature(ctx)
-
 		authHeader := s.generateAuthHeader(ctx.SubID, keySet.UniqueKeyID, createdAt, validTill, sign, requestSig)
-		log.Debugf(ctx, "Signature generated: %v (request-signature included: %v)", sign, requestSig != "")
+		log.Debugf(ctx, "Signature generated: %v (4-line signing string: %v)", sign, requestSig != "")
 		header := model.AuthHeaderSubscriber
 		if ctx.Role == model.RoleGateway {
 			header = model.AuthHeaderGateway
@@ -114,9 +120,10 @@ func (s *signStep) lookupRequestSignature(ctx *model.StepContext) string {
 	return entry.Signature
 }
 
-// generateAuthHeader constructs the authorization header for the signed request.
-// When requestSig is non-empty (solicited callback path) it appends the
-// request-signature field and extends the headers coverage string accordingly.
+// generateAuthHeader constructs the Authorization header for the signed request.
+// When requestSig is non-empty (solicited callback path) it declares
+// "request-signature" in the headers list (NFH-004 §4); the value itself is
+// in the signing string, not as a separate header attribute.
 func (s *signStep) generateAuthHeader(subID, keyID string, createdAt, validTill int64, signature, requestSig string) string {
 	base := fmt.Sprintf(
 		"Signature keyId=\"%s|%s|ed25519\",algorithm=\"ed25519\",created=\"%d\",expires=\"%d\"",
@@ -124,8 +131,8 @@ func (s *signStep) generateAuthHeader(subID, keyID string, createdAt, validTill 
 	)
 	if requestSig != "" {
 		return base + fmt.Sprintf(
-			",headers=\"(created) (expires) digest request-signature\",signature=\"%s\",request-signature=\"%s\"",
-			signature, requestSig,
+			",headers=\"(created) (expires) digest request-signature\",signature=\"%s\"",
+			signature,
 		)
 	}
 	return base + fmt.Sprintf(",headers=\"(created) (expires) digest\",signature=\"%s\"", signature)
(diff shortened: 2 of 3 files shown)


nirmay:
<html><head></head><body><h2>Code Review: PR #756 (second pass)</h2>
<h3>Summary</h3>
<p>All four findings from the first review are addressed cleanly. The diff is correct, well-structured, and ready to merge.</p>
<hr>
<h3>Previous findings — all resolved</h3>

# | Finding | Status
-- | -- | --
1 | Missing test: on_* action with no stored entry falls back to Sign | ✅ TestSignStep_Run_SolicitedCallback_NoStoredEntry_FallsBackToSign added
2 | Missing test: SignAck error propagation | ✅ TestSignStep_Run_SignAckError_ReturnsError added
3 | Missing debug log in lookupCallbackRequestSig empty-action branch | ✅ log.Debugf reinstated
4 | Duplicate "sign validation failed: %w" in validate | ✅ Unified with var validErr error dispatch


<hr>
<h3>New pass — nothing blocking found</h3>
<p>One very minor observation worth noting for completeness:</p>
<p><code>lookupCallbackRequestSig</code> strips the <code>on_</code> prefix to get the store lookup key, but only guards against the result being <code>""</code>. If a peer sends <code>request-signature</code> in the <code>headers</code> attribute with a non-<code>on_*</code> action body (e.g. <code>"search"</code>), the <code>TrimPrefix</code> leaves the action unchanged and the store lookup proceeds with <code>"search"</code> as the key. If an entry happens to exist, <code>ValidateAck</code> is called and will reject the request because the signature over the <code>search</code> body won't match. So the outcome is still a correct rejection — just via a slightly indirect path. This is pre-existing behaviour from <code>validateRequestSignatureChain</code> and nothing to fix here.</p>
<hr>
<h3>What looks good</h3>
<ul>
<li><strong>Core logic</strong>: <code>SignAck</code>/<code>ValidateAck</code> wiring is precise and symmetric — same conditions govern both the send and receive paths.</li>
<li><strong><code>validate</code> unification</strong>: clean, single error format, easier to read.</li>
<li><strong><code>lookupCallbackRequestSig</code></strong>: the comment enumerating all four fall-through conditions is clear and correct. The restored debug log makes the silent skip traceable.</li>
<li><strong>Test coverage</strong>: 12 tests covering all dispatch paths, all fallback/degradation paths, and all error paths for both <code>signStep</code> and <code>validateHeaders</code>. No gaps.</li>
<li><strong>Backward compatibility</strong>: <code>mockSigner.returnSignSig</code> defaults to <code>""</code> — all pre-existing <code>ackSignerStep</code> tests pass unchanged.</li>
</ul>
<hr>
<h3>Verdict</h3>
<p><strong>Approve — ready to merge.</strong></p></body></html>

PR #757, opened 2026-06-01
fix(signvalidator): enforce ed25519 algorithm in signature validation

### LINKED ISSUE #753: [Proto-v2.0] fix: algorithm field in Authorization header is never validated
### Summary

NFH-007 requires the `algorithm` field in the `Authorization` header to be validated on every inbound request — any value other than `ed25519` must be rejected with `401 NackUnauthorized`. The current implementation parses the `algorithm` field but never checks its value, allowing messages with any algorithm string to pass through to signature verification.

---

### What the spec says

**NFH-007 §1 — Cryptographic Primitives:**
> No other signature or hash algorithm is permitted. An NP that receives a message carrying an `algorithm` value other than `ed25519` MUST reject it with a `401 NackUnauthorized`.

**NFH-007 §7 Step 3 — Verifying a Callback:**
> Confirm `algorithm` is `ed25519`. Reject any other value.

**NFH-007 §8 Step 3 — Verifying a Request:**
> Confirm `algorithm` is `ed25519`.

**CON-004-03:**
> The signature algorithm MUST be `ed25519`. No other value is permitted.

---

### What the code does

`parseHeader` in `core/module/handler/step.go:332–360` correctly extracts all three keyId components including `Algorithm`:

```go
return &authHeader{
    SubscriberID: strings.TrimSpace(keyIDComponents[0]),
    UniqueID:     strings.TrimSpace(keyIDComponents[1]),
    Algorithm:    strings.TrimSpace(keyIDComponents[2]),  // extracted but never used
}, nil
```

However, the `Algorithm` field on the returned `authHeader` struct is never read after this point. In `validate()` (`step.go:210–224`), only `SubscriberID` and `UniqueID` are used:

```go
func (s *validateSignStep) validate(ctx *model.StepContext, value string) error {
    headerVals, err := parseHeader(value)
    ...
    signingPublicKey, _, err := s.km.LookupNPKeys(ctx, headerVals.SubscriberID, headerVals.UniqueID)
    // headerVals.Algorithm is never checked
    ...
}
```

Similarly, `signvalidator.go`'s `parseAuthHeader` parses the `algorithm` field from the header into a map but never validates it:

```go
signatureMap[key] = value  // "algorithm" → "ed25519" (or anything else)
// signatureMap["algorithm"] is never read
```

---

### Impact

A message carrying `algorithm="rsa"`, `algorithm="hs256"`, or any arbitrary string passes the algorithm check and proceeds to Ed25519 verification. While a crafted payload is unlikely to pass Ed25519 verification with the wrong algorithm, the absence of an explicit check means the implementation is not rejecting at the correct point with the correct error, and leaves open any future code path that might branch on the algorithm value.

---

### Fix

In `validate()` (`step.go`), after calling `parseHeader`, add an explicit check:

```go
if headerVals.Algorithm != "ed25519" {
    return model.NewSignValidationErr(fmt.Errorf("unsupported algorithm %q: only ed25519 is permitted", headerVals.Algorithm))
}
```

The same check should be added in `signvalidator.go`'s `parseAuthHeader` or `Validate` to keep the validation self-contained within the validator package.

nirmalnr:
**Closes #753**

### Summary

- Rejects any `Authorization` header whose `keyId` algorithm component (3rd pipe-delimited segment) is not `ed25519` in `validate()` — returns an error before key lookup or signature verification is attempted (NFH-007 §1, §7 Step 3, §8 Step 3, CON-004-03)
- Rejects any `Authorization` header whose outer `algorithm=` attribute is not `ed25519` inside `signvalidator.parseAuthHeader()`, making the validator plugin self-contained regardless of caller
- Adds `TestValidate_NonEd25519Algorithm_ReturnsError` confirming the signature validator is never invoked when the algorithm is rejected
- Adds `"Unsupported Algorithm"` failure case to `TestVerifyFailure`; updates existing signvalidator test headers to include `algorithm="ed25519"` so each case exercises its intended failure mode

### What changed

| File | Change |
|------|--------|
| `core/module/handler/step.go` | Guard in `validate()` — checks `headerVals.Algorithm != "ed25519"` before key lookup |
| `pkg/plugin/implementation/signvalidator/signvalidator.go` | Guard in `parseAuthHeader()` — checks `signatureMap["algorithm"] != "ed25519"` before timestamp parsing |
| `core/module/handler/step_test.go` | New test `TestValidate_NonEd25519Algorithm_ReturnsError` |
| `pkg/plugin/implementation/signvalidator/signvalidator_test.go` | New `"Unsupported Algorithm"` case; existing headers updated with `algorithm="ed25519"` |

### Why two validation points

The two checks cover **different fields** in the header:

```
Signature keyId="sub|key|ed25519", algorithm="ed25519", ...
                          ↑                   ↑
                   step.go checks      signvalidator checks
                   (keyId component)   (outer attribute)
```

A header where only one field is wrong would bypass a single-point check.

### Test plan

- [x] `go test ./core/module/handler/...` — all pass including `TestValidate_NonEd25519Algorithm_ReturnsError`
- [x] `go test ./pkg/plugin/implementation/signvalidator/...` — all pass including `TestVerifyFailure/Unsupported_Algorithm`
- [x] Manually send a request with `algorithm="rsa"` in the `Authorization` header → confirm `401` response

### DIFF
--- a/core/module/handler/step.go
+++ b/core/module/handler/step.go
@@ -252,6 +252,8 @@ func (s *validateSignStep) validate(ctx *model.StepContext, value, requestSig st
 	if err != nil {
 		return fmt.Errorf("failed to parse header")
 	}
+	// Guards the keyId component (sub|key|alg); parseAuthHeader in signvalidator
+	// guards the outer algorithm= attribute — the two cover different header fields.
 	if headerVals.Algorithm != "ed25519" {
 		return fmt.Errorf("unsupported algorithm %q: only ed25519 is permitted", headerVals.Algorithm)
 	}
--- a/core/module/handler/step.go
+++ b/core/module/handler/step.go
@@ -252,6 +252,9 @@ func (s *validateSignStep) validate(ctx *model.StepContext, value, requestSig st
 	if err != nil {
 		return fmt.Errorf("failed to parse header")
 	}
+	if headerVals.Algorithm != "ed25519" {
+		return fmt.Errorf("unsupported algorithm %q: only ed25519 is permitted", headerVals.Algorithm)
+	}
 	log.Debugf(ctx, "Validating Signature for subscriberID: %v", headerVals.SubscriberID)
 	signingPublicKey, _, err := s.km.LookupNPKeys(ctx, headerVals.SubscriberID, headerVals.UniqueID)
 	if err != nil {
--- a/core/module/handler/step_test.go
+++ b/core/module/handler/step_test.go
@@ -461,6 +461,27 @@ func TestValidateHeaders_SolicitedCallback_StoreError_ReturnsError(t *testing.T)
 	}
 }
 
+func TestValidate_NonEd25519Algorithm_ReturnsError(t *testing.T) {
+	sv := &mockSignValidatorBasic{}
+	km := &mockKMBasic{publicKey: "pubKey=="}
+	step, _ := newValidateSignStep(sv, km, nil)
+	vStep := step.(*validateSignStep)
+
+	badAlgHeader := `Signature keyId="bpp.example.com|key-1|rsa",algorithm="rsa",` +
+		`created="1700000000",expires="1700003600",signature="sig=="`
+
+	ctx := makeValidateStepCtx("2.0.0", "msg-alg-001", "bap.example.com",
+		badAlgHeader,
+		`{"context":{"action":"on_search","messageId":"msg-alg-001","version":"2.0.0"}}`)
+
+	if err := vStep.validate(ctx, badAlgHeader, ""); err == nil {
+		t.Fatal("expected error for non-ed25519 algorithm")
+	}
+	if sv.validateCalled || sv.validateAckCalled {
+		t.Error("signature validator must not be called when algorithm is rejected")
+	}
+}
+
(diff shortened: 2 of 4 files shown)


nirmay:
The change is correct, well-scoped, and well-tested. The dual-guard approach is sound and clearly explained. The main suggestion before merging is:

Suggested (non-blocking): Add a "Missing Algorithm" case to TestVerifyFailure in signvalidator_test.go.
Optional: Add a short inline comment in step.go explaining that this check is complementary to the one in the plugin (guards the keyId component, not the algorithm= attribute).

Safe to merge as-is; the nits can be addressed as a follow-up if preferred @nirmalnr

nirmay (approved):
approved.

nirmay:
Overall Assessment
All previous feedback has been addressed. The PR is:

Correct — both validation points guard the right fields, in the right order, before any expensive operations.
Well-tested — happy path, unsupported algorithm, missing algorithm, and all pre-existing failure cases are covered and cleanly isolated.
Convention-compliant — commit messages follow fix:/test: convention, error handling uses the established model.NewSignValidationErr pattern, logging placement is unchanged.
✅ Approved — ready to merge.

PR #762, opened 2026-06-02
feat(config): introduce signed beckn-constants file

### LINKED ISSUE #592: Beckn Infra services related config to be separated into a signed file
Global urls (dedi and cds etc.) to be out of the box when they are used. changes to those need to be out of the box as ell. 

dont make the onix users edit them, but make them extend it. this means such global urls come from a separate file that is digitally signed.

nirmay:
## Summary

Introduces a signed \`pkg/beckndefaults/beckn-constants.yaml\` file carrying Beckn-defined config values. The plugin manager loads, signature-verifies, and enforces these at plugin creation time, preventing accidental operator modification of Beckn-defined values.

**Locked constants** — startup fails if operator config contradicts:
- \`dediregistry.url\`

**Overridable constants** — injected automatically; if user sets a different value it is accepted and logged at WARN:
- \`schemav2validator.type\` (default: \`url\`)
- \`schemav2validator.location\` (default: Beckn Protocol v2.0.0-lts spec URL)

**Key design decisions:**
- Constants are enforced in \`pkg/plugin/manager\` at plugin creation time — only on \`Registry\` and \`SchemaValidator\` which handle the two in-scope plugin IDs
- \`becknConstants\` nests under \`pluginManager\` in adapter YAML; adapter has zero beckn constants logic
- Ships a baseline constants file embedded in the binary; attempts remote refresh from repo at startup, falls back to shipped file if unavailable
- Adds \`beckn_constants_info\` OTLP gauge (one series per non-canonical value detected), driven dynamically by the manager — no hardcoded key list in telemetry
- Adds \`tools/sign\` CLI for key generation and file signing (used by CI)
- Adds \`.github/workflows/sign-beckn-constants.yml\` that re-signs on changes to \`main\`
- Removes the three in-scope keys from operator-facing config files

## Test plan

- [x] Build passes: \`go build ./...\`
- [x] Tests pass: \`go test ./...\`
- [x] Start adapter with any existing config — confirm \`dediregistry.url\` and \`schemav2validator.type\`/\`location\` are logged as injected at startup
- [x] Add a conflicting \`dediregistry.url\` to a config file — confirm startup fails with a clear locked constant error
- [x] Set a different \`schemav2validator.location\` — confirm adapter starts, WARN logged with canonical and actual values
- [ ] Set \`pluginManager.becknConstants.disableRemoteRefresh: true\` — confirm no outbound call attempted
- [ ] Add \`BECKN_CONSTANTS_SIGNING_KEY\` GitHub secret and push a change to \`beckn-constants.yaml\` on \`main\` — confirm workflow re-signs and commits the \`.sig\`
- [ ] Confirm \`beckn_constants_info\` gauge appears in OTLP metrics when a non-canonical value is set; absent when all values are canonical

Closes #592


pkg/beckndefaults/beckndefaults_test.go
```diff
@@ -0,0 +1,97 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, "v1", bc.Version)
+}
+
+// TestFetchRemote_SizeLimit verifies that a response exceeding the size cap
+// is rejected without panicking.
+func TestFetchRemote_SizeLimit(t *testing.T) {
+	oversized := make([]byte, maxConstantsFileBytes+1)
+	assert.Greater(t, len(oversized), maxConstantsFileBytes)
+}
```
nirmalnr, line 117: **`TestFetchRemote_SizeLimit` does not test anything real.**

The test only asserts that a byte slice is larger than a constant — it never exercises the actual size-limit enforcement inside `fetchRemote`:

```go
func TestFetchRemote_SizeLimit(t *testing.T) {
    oversized := make([]byte, maxConstantsFileBytes+1)
    assert.Greater(t, len(oversized), maxConstantsFileBytes) // just checks math, not behaviour
}
```

Replace with an `httptest.Server` that actually serves an oversized body and verifies `fetchRemote` rejects it:

```go
func TestFetchRemote_SizeLimit(t *testing.T) {
    srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        w.Write(make([]byte, maxConstantsFileBytes+1))
    }))
    defer srv.Close()

    _, _, err := fetchRemoteFromURLs(context.Background(), srv.URL, srv.URL)
    require.Error(t, err)
    assert.Contains(t, err.Error(), "exceeds maximum allowed size")
}
```

This requires extracting the URLs as parameters in `fetchRemote` (or a thin helper) so the test can inject the mock server URL.

.github/workflows/sign-beckn-constants.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Sign constants file
+        env:
+          BECKN_SIGNING_KEY: ${{ secrets.BECKN_CONSTANTS_SIGNING_KEY }}
+        run: |
+          echo "$BECKN_SIGNING_KEY" > /tmp/beckn_private.key
+          go run ./tools/sign sign \
+            --key /tmp/beckn_private.key \
+            --input pkg/beckndefaults/beckn-constants.yaml \
+            --output pkg/beckndefaults/beckn-constants.yaml.sig
+          rm -f /tmp/beckn_private.key
```
nirmalnr: **Private key not cleaned up if `go run` fails.**

The `rm -f` is inside the same `run` block as the signing step. If `go run` exits with an error, the shell stops executing the remaining lines — the private key file is left on the runner for the rest of the job.

```yaml
run: |
  echo "$BECKN_SIGNING_KEY" > /tmp/beckn_private.key
  go run ./tools/sign sign ...   # if this fails ↓ line never runs
  rm -f /tmp/beckn_private.key   # ← not reached on failure
```

Move cleanup to a separate step with `if: always()` so it runs regardless of whether signing succeeds:

```yaml
- name: Sign constants file
  env:
    BECKN_SIGNING_KEY: ${{ secrets.BECKN_CONSTANTS_SIGNING_KEY }}
  run: |
    echo "$BECKN_SIGNING_KEY" > /tmp/beckn_private.key
    go run ./tools/sign sign \
      --key /tmp/beckn_private.key \
      --input pkg/beckndefaults/beckn-constants.yaml \
      --output pkg/beckndefaults/beckn-constants.yaml.sig

- name: Cleanup signing key
  if: always()
  run: rm -f /tmp/beckn_private.key
```

nirmalnr (approved):
We could think if we should have some sort of versioning to the constant file in case we would want different constants for different releases just in case. But I think there is no need for that currently and should not be in scope for this specific PR.

I have been thinking through if pulling the latest config by an older version of the app could cause issues but I think since `applyConstants` will just ignore any unknown plugins it shouldn't be a problem. In the future for any plugins I we should be mindful of this to make sure introducing an unknown config will not cause any issues for older versions of the plugin. 

This PR is good to merge.

PR #794, opened 2026-06-08
feat(schemav2validator): auxiliary spec support for non-Beckn and extended-verb deployments

### LINKED ISSUE #780: feat: schemav2validator — support multiple OpenAPI specs to enable non-Beckn and extended-verb deployments
### Problem

\`schemav2validator\` is configured with a single OpenAPI spec (\`type\` + \`location\`). The validator builds its action index by scanning that one document and keying schemas on \`context.action\`. Any message whose action is not present in that spec is rejected as \`unsupported action\`.

This design works perfectly for a standard Beckn network (one \`beckn.yaml\`), but breaks down in two real scenarios:

1. **Non-Beckn networks or custom deployments** — A domain, app, or network that has extended the Beckn API surface with additional verbs (e.g. \`subscribe\`, \`renew\`, \`allocate\`) cannot use ONIX at all today, because those verbs have no entry in the upstream \`beckn.yaml\`.
2. **Multi-spec networks** — An operator running several sub-networks (e.g. energy + retail + mobility) each with their own OpenAPI complement cannot load all of them into a single validator instance.

The tight \`location → single spec\` binding means ONIX is effectively locked to the Beckn core protocol file, contradicting the plugin architecture's goal of being network-agnostic middleware.

### Relevant code

- [`schemav2validator.go:loadSpec`](pkg/plugin/implementation/schemav2validator/schemav2validator.go#L201) — loads exactly one document per validator instance.
- [`schemav2validator.go:buildActionIndex`](pkg/plugin/implementation/schemav2validator/schemav2validator.go#L380) — merges all path operations into one flat `action → schema` map with no provision for multiple source specs.
- [`cmd/plugin.go`](pkg/plugin/implementation/schemav2validator/cmd/plugin.go#L32) — config accepts only a single `type` / `location` pair.
- The `dir` type is stubbed at line 219 but not implemented.

nirmay:
## Summary

Closes #780

- Adds `auxiliary_types` / `auxiliary_locations` comma-separated config keys to `schemav2validator`, allowing operators to layer additional OpenAPI specs on top of the primary (signed) Beckn spec
- Primary spec is now optional — non-Beckn deployments can operate with auxiliary specs only
- Implements the previously-stubbed `dir` type: loads all top-level `*.yaml`/`*.yml`/`*.json` files in a directory as specs
- Hard-rejects at startup if any auxiliary action collides with the primary (auxiliary specs may only add, never shadow)
- On TTL refresh, merged index is rebuilt from scratch; previous valid index is retained on failure

## Config shape

```yaml
plugins:
  schemav2validator:
    type: url
    location: https://.../beckn.yaml          # primary — signed, injected by beckndefaults
    auxiliary_types: "file,url,dir"
    auxiliary_locations: "/etc/onix/energy.yaml,https://example.com/custom.yaml,/etc/onix/domain/"
```

## Test plan

- [ ] `TestAuxiliary_AddsNewActions` — primary + auxiliary actions all validate
- [ ] `TestAuxiliary_ShadowPrimaryHardRejects` — collision on startup returns error
- [ ] `TestAuxiliary_DirType` — all files in dir loaded, actions available
- [ ] `TestAuxiliary_NoPrimary_NonBeckn` — no primary, auxiliary carries full action set
- [ ] `TestAuxiliary_BadLocationSkipped` — bad auxiliary skipped, primary actions intact
- [ ] `TestProvider_AuxiliaryConfig` — config parsing: valid, mismatched lengths, empty entries
- [ ] All existing tests pass unmodified

### DIFF
--- a/pkg/plugin/implementation/schemav2validator/cmd/plugin_test.go
+++ b/pkg/plugin/implementation/schemav2validator/cmd/plugin_test.go
@@ -178,12 +178,12 @@ func TestProvider_AuxiliaryConfig(t *testing.T) {
 		errMsg  string
 	}{
 		{
-			name: "valid auxiliary config",
+			name: "auxiliary collides with primary",
 			config: map[string]string{
-				"type":               "url",
-				"location":           primary.URL,
-				"auxiliary_types":    "url",
-				"auxiliary_locations": aux.URL,
+				"type":              "url",
+				"location":          primary.URL,
+				"auxiliaryTypes":    "url",
+				"auxiliaryLocations": aux.URL,
 			},
 			// aux and primary serve the same spec — collision on "test" action.
 			wantErr: true,
@@ -194,8 +194,8 @@ func TestProvider_AuxiliaryConfig(t *testing.T) {
 			config: map[string]string{
 				"type":               "url",
 				"location":           primary.URL,
-				"auxiliary_types":    "url,file",
-				"auxiliary_locations": aux.URL,
+				"auxiliaryTypes":     "url,file",
+				"auxiliaryLocations": aux.URL,
 			},
 			wantErr: true,
 			errMsg:  "same number",
@@ -203,33 +203,33 @@ func TestProvider_AuxiliaryConfig(t *testing.T) {
 		{
 			name: "empty auxiliary entry",
 			config: map[string]string{
-				"type":                "url",
-				"location":            primary.URL,
-				"auxiliary_types":     "url,",
-				"auxiliary_locations": aux.URL + ",",
+				"type":               "url",
+				"location":           primary.URL,
+				"auxiliaryTypes":     "url,",
+				"auxiliaryLocations": aux.URL + ",",
 			},
 			wantErr: true,
 			errMsg:  "must not be empty",
 		},
 		{
-			name: "auxiliary_types set without auxiliary_locations",
+			name: "auxiliaryTypes set without auxiliaryLocations",
 			config: map[string]string{
-				"type":            "url",
-				"location":        primary.URL,
-				"auxiliary_types": "url",
+				"type":           "url",
+				"location":       primary.URL,
+				"auxiliaryTypes": "url",
 			},
 			wantErr: true,
-			errMsg:  "auxiliary_locations is missing",
+			errMsg:  "auxiliaryLocations is missing",
 		},
 		{
-			name: "auxiliary_locations set without auxiliary_types",
+			name: "auxiliaryLocations set without auxiliaryTypes",
 			config: map[string]string{
-				"type":                "url",
-				"location":            primary.URL,
-				"auxiliary_locations": aux.URL,
+				"type":               "url",
+				"location":           primary.URL,
+				"auxiliaryLocations": aux.URL,
 			},
 			wantErr: true,
-			errMsg:  "auxiliary_types is missing",
+			errMsg:  "auxiliaryTypes is missing",
 		},
 		{
 			name: "no auxiliary keys — primary only",
--- a/pkg/plugin/implementation/schemav2validator/cmd/plugin_test.go
+++ b/pkg/plugin/implementation/schemav2validator/cmd/plugin_test.go
@@ -160,6 +160,100 @@ func TestProvider_New(t *testing.T) {
 	}
 }
 
+func TestProvider_AuxiliaryConfig(t *testing.T) {
+	primary := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte(testSpec))
+	}))
+	defer primary.Close()
+
+	aux := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte(testSpec))
+	}))
+	defer aux.Close()
+
+	tests := []struct {
+		name    string
+		config  map[string]string
+		wantErr bool
+		errMsg  string
+	}{
+		{
+			name: "valid auxiliary config",
+			config: map[string]string{
+				"type":               "url",
+				"location":           primary.URL,
+				"auxiliary_types":    "url",
+				"auxiliary_locations": aux.URL,
+			},
+			// aux and primary serve the same spec — collision on "test" action.
+			wantErr: true,
+			errMsg:  "already defined",
+		},
+		{
+			name: "mismatched auxiliary list lengths",
+			config: map[string]string{
+				"type":               "url",
+				"location":           primary.URL,
+				"auxiliary_types":    "url,file",
+				"auxiliary_locations": aux.URL,
+			},
+			wantErr: true,
+			errMsg:  "same number",
+		},
+		{
+			name: "empty auxiliary entry",
+			config: map[string]string{
+				"type":               "url",
+				"location":           primary.URL,
+				"auxiliary_types":    "url,",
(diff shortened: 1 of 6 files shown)


nirmalnr:
## Collision behaviour in `dir` type specs

The PR handles action collisions differently depending on where the collision occurs:

| Collision type | Behaviour |
|---|---|
| Auxiliary spec vs primary spec | Hard error — adapter refuses to start |
| Auxiliary spec vs earlier auxiliary spec | Hard error — adapter refuses to start |
| File A vs File B within the same `dir` | Warning only — last file wins silently |

The within-dir soft behaviour may be intentional, but it sits at odds with the strict collision policy everywhere else. An operator who accidentally duplicates an action across two files in a `dir` gets a silent overwrite rather than a startup failure — which could be a surprising debugging experience.

Would it be worth making within-dir collisions a hard error as well, to keep the policy consistent? If the soft behaviour is deliberate (e.g., to allow override files), that's totally reasonable — but it would be good to call that intent out explicitly in the docs, since `CONFIG.md` currently doesn't mention the collision behaviour at all, and neither doc mentions that "later" means alphabetically later.

Ideally I feel like within the same dir also the collisions should result in a hard error because it can lead to unintended consequences for the user. Happy to discuss if there's a specific reason for the asymmetry!

nirmalnr:
#### Config key naming: `auxiliary_types` / `auxiliary_locations`

Minor style note — the new keys use snake_case, while the existing config keys use either camelCase (`cacheTTL`) or a camelCase prefix with underscore separator (`extendedSchema_enabled`, `extendedSchema_cacheTTL`).

Would it be worth renaming these to `auxiliaryTypes` / `auxiliaryLocations` to stay consistent with the established convention? Happy to defer if there's a reason for the difference.

pkg/plugin/implementation/schemav2validator/cmd/plugin_test.go
```diff
@@ -160,6 +160,120 @@ func TestProvider_New(t *testing.T) {
...
+	defer aux.Close()
+
+	tests := []struct {
+		name    string
+		config  map[string]string
+		wantErr bool
+		errMsg  string
+	}{
+		{
+			name: "valid auxiliary config",
```
nirmalnr: The test case named `"valid auxiliary config"` actually expects `wantErr: true` — it errors because both the primary and auxiliary servers serve the same spec, causing a collision on the `"test"` action.

The name implies the config is valid and should succeed, which is the opposite of what the test asserts. Something like `"auxiliary collides with primary"` would better reflect the intent.

nirmalnr:
### Adapter can start and serve with a fully empty action index

If all specs fail to load at startup (e.g. no primary configured and all auxiliaries skipped due to load errors), `loadAllSpecs` does not return an error — it logs a warning and commits an empty action index with `specsLoaded = true`.

A warning is currently logged:
> `schemav2validator: no actions indexed — all requests will fail schema validation`

However this is easy to miss in practice — it fires once at startup and is then silent, while every subsequent request fails with `"unsupported action: <action>"` giving the operator no clear signal that the root cause is a startup misconfiguration rather than a bad request.

Would it be worth turning this into a hard error instead? As the adaptor will be pretty much useless in this state.

nirmay:
Here's a summary of what was fixed for each review comment:

**Comment 1 — Within-dir collision consistency (nirmalnr)**
`loadSpecFromDir` now returns a hard error on intra-dir action collision instead of warn-and-overwrite. Behavior: the dir spec is skipped (same skip policy as any other load failure), the collision is logged at ERROR, and `New()` fails with "no actions indexed" if no other spec is available. Two new tests cover this path.

**Comment 2 — Config key naming**
`auxiliary_types`/`auxiliary_locations` → `auxiliaryTypes`/`auxiliaryLocations` throughout `plugin.go`, both test files, `CONFIG.md`, and `README.md`.

**Comment 3 — Empty action index should be a startup failure**
The `log.Warnf` in `loadAllSpecs` is now `return fmt.Errorf(...)` — adapter refuses to start rather than starting in a state where every request fails silently.

nirmalnr:
## Auxiliary spec load failure on TTL refresh silently drops actions

On TTL refresh a soft auxiliary failure can silently removes previously valid actions from the index. Even if this was a short temporary issue, an endpoint that was previously working might end up returning error till the next ttl refresh.

| | Startup | TTL Refresh |
|---|---|---|
| **Primary fails to load** | Hard error — adapter refuses to start | Hard error — old index retained |
| **Auxiliary fails to load** | Soft — skipped, adapter starts without it | Soft — skipped, new index committed **without** its actions |
| **Action collision** | Hard error — adapter refuses to start | Hard error — old index retained |

The auxiliary failure case, on refresh an auxiliary that was successfully contributing actions at startup can silently disappear from the index on the next TTL cycle (e.g. if a file is deleted or a URL is temporarily down), causing requests that were previously valid to start returning `"unsupported action"`. When this happens, an error is logged for the failed auxiliary load but until the next ttl cycle the aux endpoints will not be supported.

Is the current behaviour intentional here? I understand that we are prioritising getting the primary update and skipping the aux in case of error. Would it be worth retaining the last successfully loaded auxiliary index when a refresh fails or is that considered unnecessary complexity given that the primary stays current?

nirmalnr:
Found another issue while testing some of the cache refresh scenarios.

### TTL refresh is effectively a no-op for file and URL specs due to kin-openapi's global URI cache

`kin-openapi` v0.137.0 defines `DefaultReadFromURI` as a package-level global with a `URIMapCache` closure:

```go
var DefaultReadFromURI = URIMapCache(ReadFromURIs(ReadFromHTTP(http.DefaultClient), ReadFromFile))
```

`URIMapCache` holds a `map[string][]byte` that lives for the **entire process lifetime**. Every `openapi3.NewLoader()` call — including the ones created on each TTL reload — reuses this same global cache. Once a spec is loaded at startup, every subsequent TTL reload returns the cached bytes without ever touching the filesystem or making an HTTP request.

This affects all three source types:
- **`file`** — file is never re-read from disk after startup
- **`url`** — remote spec is never re-fetched after startup
- **`dir`** — file additions and deletions are picked up (via `os.ReadDir`), but changes to existing files in the dir are silently ignored

This isn't introduced by this PR — the same issue exists on `main`. This can be fixed as a separate issue but I found this while testing the refresh conditions we discussed as part of this PR.

nirmay:
**Re: auxiliary TTL silently drops actions**

This is already handled by commit `68682e3`. `reloadAllSpecs` calls `loadAllSpecs(ctx, true)` — the `failOnAuxError=true` flag causes an immediate `return error` on any auxiliary failure, before the write lock is ever acquired. The old `actionSchemas`/`bodylessActions` maps are never replaced. Test coverage: `TestAuxiliary_RefreshRetainsIndexOnAuxFailure` forces a reload while the auxiliary is down and asserts both primary and auxiliary actions are still served.

The table in the comment describes the behaviour before that fix. After the fix the TTL refresh column for auxiliary failures reads: **Hard — old index retained**.

---

**Re: kin-openapi global URI cache**

Tracked as #795. Fix is straightforward — replace `openapi3.NewLoader()` with a loader that sets a per-call `ReadFromURIFunc` bypassing the global `URIMapCache`. Will be picked up in the next iteration.

nirmalnr (approved):
With the new issue #795 raised for the cache issue, the PR looks good and the basic cases have been tested. Once #795 has been fixed the cache/refresh related scenarios can also be tested. The PR is good to be merged.

Issue #770, opened 2026-06-03
epic: schema version mediator plugin


nirmay:
## Overview
The core of the schema version mediation feature. A new `SchemaVersionMediator` named plugin slot that runs as a built-in pipeline step. At construction it receives `ManifestLoader` and a registry of `Translator` plugins as injected dependencies. At runtime it sits in the handler's step pipeline — the operator places `"mediateSchema"` in their `cfg.Steps` list.

The plugin owns the full mediation loop: walk the payload, check schema object compatibility against the relevant manifest, fetch translation artifacts from schema.beckn.io, dispatch to the appropriate translator by artifact content-type, and enforce the configured data-loss policy.

Translation execution is deliberately separated into a `Translator` plugin interface. Today's only implementation is JSONata (`jsonatatranslator`). Future implementations (SHACL, agent-based) can be added as new `.so` files with a config YAML change — no mediator code changes required. This also handles the case where a single payload contains schema objects that require different translation mechanisms.

## Tasks
- [ ] SchemaVersionMediator interface, SchemaVersionMediatorProvider, Translator interface and TranslatorProvider `[definition]`
- [ ] Message tree walker — extract (context_url, type) pairs from payload `[schema]`
- [ ] Compatibility check — compare extracted pairs against a NodeManifest `[schema]`
- [ ] JSONataTranslator plugin — single-pass composition and execution for application/jsonata artifacts `[plugin]`
- [ ] Translation map manager — fetch artifacts from schema.beckn.io, capture Content-Type for dispatcher `[schema]`
- [ ] Translation policy model and loader with wildcard and expiry support `[model]`
- [ ] Mediation loop — inbound path with per-artifact-type translator dispatch `[plugin]`
- [ ] Mediation loop — outbound path with nil body guard and per-artifact-type translator dispatch `[plugin]`
- [ ] Cold-start guard — reject inbound when local manifest is absent or empty `[plugin]`
- [ ] Data-loss policy enforcement (drop_and_signal / reject) `[plugin]`
- [ ] Handler wiring — PluginCfg, PluginManager interface, stdHandler field, step factory, Translator slice dependency `[core]`
- [ ] SchemaVersionMediator plugin entry point (.so) and build-plugins.sh registration `[build]`
- [ ] JSONataTranslator plugin entry point (.so) and build-plugins.sh registration `[build]`

## Dependencies
[...]

Issue #769, opened 2026-06-03
epic: node manifest in manifestloader


nirmay:
## Overview
Extends the existing `ManifestLoader` plugin and DeDi registry client to support node-level manifests alongside the existing network manifests. No new plugin is introduced — the entire fetch/verify/cache machinery is reused. Only the lookup path and model types change.

A node manifest is the capability declaration published by each Beckn participant. It declares which schema object versions their application handles. It is a sibling to the existing `network-manifest` type, sharing the same DeDi registry placement convention, signing policy, and manifest loader infrastructure.

The `subscriberId` in a node manifest is a fully-qualified three-part DeDi reference (`namespace/registry/recordId`) — the same value that appears as `bapId`/`bppId` in the Beckn transaction context. This maps directly to a DeDi lookup endpoint with no keyID or networkID required, mirroring how network manifest discovery works via the existing 2-segment registry lookup.

See the schema version mediation approach document for full design context.

## Tasks
- [ ] #763 NodeManifest model and SchemaObject type
- [ ] #764 Add SubscriberID field to ManifestDocument
- [ ] #765 Extend RegistryMetadataLookup interface with LookupSubscriberMeta
- [ ] #766 Implement LookupSubscriberMeta on DeDi client
- [ ] #767 Extend ManifestLoader interface with GetBySubscriberID
- [ ] #768 Implement GetBySubscriberID on ManifestLoader

## Notes
- All tasks can be developed and tested against mock DeDi responses immediately — no dependency on the DeDi team shipping their side first
- #766 will return empty results in production until the DeDi team ships the new 3-segment subscriber metadata endpoint — handled gracefully via `ErrNoManifestPublished`

nirmay:
merged with #825

PR #796, opened 2026-06-09
feat(definition): SchemaVersionMediator and Translator plugin interfaces

nirmay:
## Summary

- Adds `SchemaVersionMediator` interface and `SchemaVersionMediatorProvider` in `pkg/plugin/definition/schemaversionmediator.go`
- Adds `Translator` interface and `TranslatorProvider` in `pkg/plugin/definition/translator.go`

## Design decisions

- `Mediate(ctx *model.StepContext) error` is direction-agnostic — both the receiver handler (inbound) and caller handler (outbound) invoke the same method via their respective `StepContext`. Direction-specific guards (cold-start, nil-body) are implementation concerns, not interface concerns.
- `ManifestLoader` is injected into `SchemaVersionMediatorProvider.New` — the mediator does not own the manifest fetch/cache lifecycle.
- `Translator` is stateless (`Translate(ctx, artifact, payload)`). The mediator selects the implementation by artifact `Content-Type`, enabling a single mediation pass to dispatch different payload fragments to different translators (e.g. JSONata for one schema delta, SHACL for another).
- `CompatibilityCheck` is intentionally not a separate interface — it is an internal concern of the mediator implementation with no standalone plugin utility.

## Dependencies

Closes part of #770
Requires #769 (ManifestLoader interface) — already available in `definition/manifestloader.go`

## Test plan

- [ ] `go build ./pkg/plugin/definition/...` passes (verified locally)
- [ ] Implementation branches (walker, mediation loop) will add unit tests against these interfaces

### DIFF
--- a/pkg/plugin/definition/schemaversionmediator.go
+++ b/pkg/plugin/definition/schemaversionmediator.go
@@ -1,0 +1,25 @@
+package definition
+
+import (
+	"context"
+
+	"github.com/beckn-one/beckn-onix/pkg/model"
+)
+
+// SchemaVersionMediator mediates schema version differences in a Beckn payload.
+// It walks the payload, checks compatibility against the node manifest, fetches
+// translation artifacts, dispatches to the appropriate translator, and enforces
+// the configured data-loss policy.
+//
+// Mediate is direction-agnostic: both the receiver handler (inbound) and the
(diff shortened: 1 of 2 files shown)


nirmay:
Merging, tested locally.

PR #799, opened 2026-06-10
fix(schemav2validator): bypass kin-openapi global URI cache for TTL reloads

### LINKED ISSUE #795: fix(schemav2validator): kin-openapi global URI cache makes TTL refresh a no-op for file and url specs
## Problem

`kin-openapi` v0.137.0 exposes a package-level global loader cache:

```go
// github.com/getkin/kin-openapi/openapi3/loader.go
var DefaultReadFromURI = URIMapCache(ReadFromURIs(ReadFromHTTP(http.DefaultClient), ReadFromFile))
```

`URIMapCache` wraps a `map[string][]byte` that lives for the **entire process lifetime**. Every `openapi3.NewLoader()` call in `schemav2validator` — including the ones created on each TTL reload — reuses this global cache. Once a spec is fetched at startup, every subsequent TTL reload returns the cached bytes without re-reading the file or making a network request.

### Effect by source type

| Source type | TTL refresh behaviour |
|---|---|
| `url` | Remote spec is **never re-fetched** after startup — HTTP request is skipped |
| `file` | File is **never re-read** from disk after startup |
| `dir` | File **additions and deletions** are picked up via `os.ReadDir`, but **changes to existing files** are silently ignored |

### Impact

- Operators who update a spec file or redeploy a remote spec cannot pick up the change without restarting the adapter.
- The `cacheTTL` setting gives a false impression that the spec is periodically refreshed.

## Context

Found while testing TTL refresh behaviour during review of PR #794 (auxiliary spec support). The issue pre-dates that PR and exists on `main`.

## Proposed fix

Replace `openapi3.NewLoader()` with a loader that bypasses the global cache. The simplest approach is to set a custom `ReadFromURIFunc` on each loader instance that always reads fresh:

```go
loader := openapi3.NewLoader()
loader.IsExternalRefsAllowed = true
loader.ReadFromURIFunc = openapi3.ReadFromURIs(
    openapi3.ReadFromHTTP(http.DefaultClient),
    openapi3.ReadFromFile,
)
```

This removes the global cache for all loaders created by `schemav2validator`, making TTL refresh actually re-fetch specs. Note that `$ref` resolution within a single load call will still cache within that call's scope (which is desirable).

## Acceptance criteria

- [ ] `schemav2validator` creates loaders with a per-call `ReadFromURIFunc` that bypasses the global cache
- [ ] A test verifies that a TTL refresh picks up a changed spec (e.g. serve v1 at startup, update handler to serve v2, trigger `reloadAllSpecs`, assert v2 actions are available)
- [ ] No regression in existing tests

nirmalnr:
## Problem

`kin-openapi` v0.137.0 wraps `DefaultReadFromURI` in a package-level `URIMapCache` that persists for the entire process lifetime. Every `openapi3.NewLoader()` call in `schemav2validator` reused this global cache, making TTL-driven reloads a silent no-op:

- **url**: remote spec fetched once at startup; HTTP request never repeated
- **file**: file read once at startup; disk never re-read
- **dir**: file additions/deletions picked up via `os.ReadDir`, but edits to existing files silently ignored
- **extended schema**: same bug — domain schema re-fetches after LRU TTL expiry still served stale bytes from the global cache

Operators who updated a spec file or redeployed a remote spec could not pick up the change without restarting the adapter, despite `cacheTTL` being configured.

## Fix

Introduced two package-level helpers in `schemav2validator`:

- `freshReadFromURI` — reads bytes directly from disk or network on every call, bypassing the `URIMapCache` entirely
- `newFreshLoader` — returns an `openapi3.Loader` with `ReadFromURIFunc` set to `freshReadFromURI`; used everywhere instead of `openapi3.NewLoader()`

Applied in two places:

| File | Change |
|---|---|
| `schemav2validator.go` | `loadSingleSpec` now uses `newFreshLoader()` — covers primary, auxiliary (`url`/`file`/`dir`) |
| `extended_schema.go` | `loadSchemaFromPath` uses `newFreshLoader()`; `$ref` fallback in localSchema path replaced `DefaultReadFromURI` with `freshReadFromURI` |

## Tests added

- `TestTTLRefresh_PicksUpChangedURLSpec` — serves v1 at startup, switches httptest server to v2, forces `reloadAllSpecs`, asserts v2 action is available
- `TestTTLRefresh_PicksUpChangedFileSpec` — same flow using a temp file overwritten with `os.WriteFile`
- `TestLoadSchemaFromPath_TTLExpiry_FetchesFresh` — loads a domain schema with a 1 ms TTL, waits for expiry, switches httptest server to v2, reloads, asserts v2 title is returned

`TestAuxiliary_RefreshRetainsIndexOnAuxFailure` updated to remove the `NOTE: … until
#795 is fixed` comment — the test now properly exercises the retain-old-index path (server returns 503, reload fails, stale index is retained).

## Notes

- Setting `ReadFromURIFunc` causes kin-openapi to skip its own `IsExternalRefsAllowed` guard (the custom function becomes the sole access policy). Since `freshReadFromURI` imposes no restrictions and we always set `IsExternalRefsAllowed = true`, behaviour is unchanged — but the flag is now a no-op and is kept only for documentation intent.
- `freshReadFromURI` uses `http.Get` (i.e. `http.DefaultClient`) with no explicit timeout, matching the library's own `ReadFromHTTP` behaviour. A per-request timeout for spec loading can be wired in as a separate improvement.

Closes #795

### DIFF
--- a/pkg/plugin/implementation/schemav2validator/extended_schema.go
+++ b/pkg/plugin/implementation/schemav2validator/extended_schema.go
@@ -315,8 +315,7 @@ func (c *schemaCache) loadSchemaFromPath(ctx context.Context, schemaPath string,
 		return nil, fmt.Errorf("failed to parse schema URL %s: %w", schemaPath, parseErr)
 	}
 
-	loader := openapi3.NewLoader()
-	loader.IsExternalRefsAllowed = true
+	loader := newFreshLoader()
 
 	var doc *openapi3.T
 	var err error
@@ -335,7 +334,7 @@ func (c *schemaCache) loadSchemaFromPath(ctx context.Context, schemaPath string,
 		if ok {
 			log.Debugf(ctx, "Loading from memory: %s -> %s", schemaPath, relPath)
 
-			// $ref lookups: memory first, network fallback.
+			// $ref lookups: memory first, fresh-fetch fallback (bypasses global URIMapCache).
 			loader.ReadFromURIFunc = func(l *openapi3.Loader, refURL *url.URL) ([]byte, error) {
 				refRelPath := extractRelativeSchemaPath(refURL)
 				c.mu.RLock()
@@ -346,7 +345,7 @@ func (c *schemaCache) loadSchemaFromPath(ctx context.Context, schemaPath string,
 					return data, nil
 				}
 				log.Debugf(ctx, "$ref not in memory, fetching from network: %s", refURL.String())
-				return openapi3.DefaultReadFromURI(l, refURL)
+				return freshReadFromURI(l, refURL)
 			}
 
 			doc, err = loader.LoadFromData(schemaBytes)
--- a/pkg/plugin/implementation/schemav2validator/extended_schema.go
+++ b/pkg/plugin/implementation/schemav2validator/extended_schema.go
@@ -316,6 +316,7 @@ func (c *schemaCache) loadSchemaFromPath(ctx context.Context, schemaPath string,
 	}
 
 	loader := newFreshLoader()
+	loader.Context = ctx
 
 	var doc *openapi3.T
 	var err error
--- a/pkg/plugin/implementation/schemav2validator/extended_schema_test.go
+++ b/pkg/plugin/implementation/schemav2validator/extended_schema_test.go
@@ -1014,7 +1014,7 @@ func TestLoadSchemaFromPath_TTLExpiry_FetchesFresh(t *testing.T) {
 	assert.Equal(t, "Schema v1", doc1.Info.Title)
 
 	// Wait for the LRU entry to expire, then switch the server to v2.
-	time.Sleep(5 * time.Millisecond)
+	time.Sleep(10 * time.Millisecond)
 	serveV2.Store(true)
 
 	// Re-load — LRU miss (expired), freshReadFromURI fetches from the server and gets v2.
--- a/pkg/plugin/implementation/schemav2validator/extended_schema_test.go
+++ b/pkg/plugin/implementation/schemav2validator/extended_schema_test.go
@@ -2,9 +2,12 @@ package schemav2validator
 
 import (
 	"context"
+	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"os"
 	"path/filepath"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


nirmay:
1. http.Get with no timeout — freshReadFromURI uses http.DefaultClient with no deadline. The PR notes this but accepts it as-is. Given this runs in the hot path of every spec reload (and potentially on every $ref resolution), a hanging remote server can stall the reload goroutine indefinitely. Recommend at minimum extracting a package-level specHTTPClient with a sensible timeout (e.g., 30 s), or wiring the caller's ctx in. This is a real production risk for remote specs.
2. //nolint:noctx on http.Get — The suppression is honest but it means context cancellation (e.g., adapter shutdown) won't interrupt an in-flight spec fetch. If a context.Context is available at the call site (loadSchemaFromPath and loadSingleSpec both receive ctx), threading it through freshReadFromURI via http.NewRequestWithContext would be cleaner than nolinting.

The fix is correct, the approach is sound, and the test additions are meaningful. The no-timeout HTTP client and the noctx suppression are worth addressing before merge — everything else looks very good.

@nirmalnr

nirmalnr:
## Updates

Addressed both review comments and added two further hardening changes.

### Review comment fixes

**1. HTTP timeout + context propagation** (no-timeout / noctx feedback)

- Added `specHTTPClient = &http.Client{Timeout: 30 * time.Second}` — prevents a
  hanging remote server from stalling the reload goroutine indefinitely.
- Replaced `http.Get` with `http.NewRequestWithContext` using `loader.Context` when
  set, falling back to `context.Background()`. Removed the `//nolint:noctx` comment.
- `loadSingleSpec` now sets `loader.Context = ctx` after `newFreshLoader()`, so
  adapter shutdown / context cancellation is propagated into in-flight spec fetches.
- `extended_schema.go` already set `loader.Context` before `LoadFromURI` — it gets
  the context and timeout benefits for free through `freshReadFromURI`.

### Additional hardening

**2. Missing `dir` spec regression test**

Added `TestTTLRefresh_PicksUpChangedDirSpec` — overwrites an existing file inside a
`dir` spec in-place and asserts the changed action is available after `reloadAllSpecs`.
This is the exact case issue #795 called out as silently ignored.

**3. Unbounded `io.ReadAll` memory risk**

`freshReadFromURI` previously called `io.ReadAll(resp.Body)` with no size limit — a
misconfigured proxy or adversarial endpoint could send an arbitrarily large body and
exhaust heap before any error was returned.

- Added `maxSpecBodyBytes int64 = 32 * 1024 * 1024` (32 MB, declared `var` for test
  overriding). No legitimate OpenAPI spec approaches this limit.
- Replaced `io.ReadAll` with `io.LimitReader(resp.Body, maxSpecBodyBytes+1)` + explicit
  size check — reading one byte over the cap distinguishes "at limit" from "over limit"
  without a second read.
- Added `TestFreshReadFromURI_BodyTooLarge` — lowers the cap to 10 bytes, serves a
  larger body, asserts an error is returned.

**4. Test flake reduction**

`TestLoadSchemaFromPath_TTLExpiry_FetchesFresh` sleep increased from 5 ms to 10 ms to
reduce false failures on heavily loaded CI runners with a 1 ms TTL.

nirmay (approved):
Approved.

PR #800, opened 2026-06-10
fix: cache registry lookup results to prevent repeated network calls

### LINKED ISSUE #774: bug: registry lookup results are not cached, causing repeated calls to the registry service
### Summary
ONIX makes repeated network calls to the registry service for every inbound request, even when the same subscriber's keys have already been fetched. This is because the cache write-back in `LookupNPKeys` was never implemented — the cache read is attempted but the result is never written back after a successful registry lookup.

---

### Background
Both `KeyMgr` and `SimpleKeyMgr` implement `LookupNPKeys` which is called on every inbound request to fetch a remote participant's signing public key for signature verification. The method attempts a cache read first, but **never writes the result back to cache** after a successful registry lookup. A search across the entire codebase confirms that `Cache.Set` is never called in either key manager implementation, meaning the cache check at the start of `LookupNPKeys` can never succeed.

```go
func (km *KeyMgr) LookupNPKeys(...) (string, string, error) {
    cachedData, err := km.Cache.Get(ctx, cacheKey)  // always misses — cache is never populated
    if err == nil {
        // ...return cached value — never reached
    }
    subscribers, err := km.Registry.Lookup(...)      // always called
    // result never written to cache
    return subscribers[0].SigningPublicKey, subscribers[0].EncrPublicKey, nil
}
```

---

### Impact

For every inbound request, `LookupNPKeys` is called **twice** for the same `subscriberID + keyID`:
- Once in `validateSignStep` — to verify the inbound request signature
- Once in `validateAckSign` — to verify the ACK response signature

Since the cache is never populated, both calls always result in a live network request to the registry service, fetching identical data.

nirmalnr:
## Summary

Fixes #774

Registry lookup results were never written back to cache after a successful network call, causing `LookupNPKeys` to make live HTTP requests to the registry on every inbound request — twice per request (once in `validateSignStep`, once in `validateAckSign`).

## What changed

Moved cache responsibility from the key manager layer down to the registry layer, where the network call actually lives.

- `RegistryClient.Lookup` and `DeDiRegistryClient.Lookup` now accept a `definition.Cache` at construction time, read from it before making an HTTP call, and write results back after a successful response
- `KeyMgr` and `SimpleKeyMgr` no longer own a `Cache` field or contain cache logic — they delegate entirely to the registry
- `RegistryLookupProvider.New` gains a `Cache` parameter; `KeyManagerProvider.New` loses it
- TTL is resolved from the response where available (`ValidUntil` for registry, `ttl` field for DeDi), falling back to a configured `cacheTTL` or a default of 5 minutes
- OTel spans added to both registry clients: `"cache lookup"` (covers `Get` + `Unmarshal`) and `"http lookup"` (covers the HTTP call only)

## Behaviour

- Cache hit → HTTP call skipped entirely
- Cache miss → result fetched and written back; `cache.Set` failure is non-fatal (warns and returns result)
- Corrupt cached value → falls through to HTTP
- `nil` cache → behaves as before, no cache operations

## Test plan

- [ ] `TestRegistryClient_Lookup_Cache` — cache hit, miss, corrupt value, set error, nil cache, TTL variants
- [ ] `TestDeDiRegistryClient_Lookup_Cache` — same coverage for DeDi client
- [ ] All 46 test packages pass
- [ ] Local testing to ensure registry API calls are not being made repeatedly for same NP.

### DIFF
--- a/cmd/adapter/main_test.go
+++ b/cmd/adapter/main_test.go
@@ -80,7 +80,7 @@ func (m *MockPluginManager) Registry(ctx context.Context, cache definition.Cache
 }
 
 // KeyManager returns a mock implementation of the KeyManager interface.
-func (m *MockPluginManager) KeyManager(ctx context.Context, cache definition.Cache, rLookup definition.RegistryLookup, cfg *plugin.Config) (definition.KeyManager, error) {
+func (m *MockPluginManager) KeyManager(ctx context.Context, rLookup definition.RegistryLookup, cfg *plugin.Config) (definition.KeyManager, error) {
 	return nil, nil
 }
 
--- a/cmd/adapter/main_test.go
+++ b/cmd/adapter/main_test.go
@@ -75,7 +75,7 @@ func (m *MockPluginManager) Cache(ctx context.Context, cfg *plugin.Config) (defi
 }
 
 // Registry returns a mock implementation of the RegistryLookup interface.
-func (m *MockPluginManager) Registry(ctx context.Context, cfg *plugin.Config) (definition.RegistryLookup, error) {
+func (m *MockPluginManager) Registry(ctx context.Context, cache definition.Cache, cfg *plugin.Config) (definition.RegistryLookup, error) {
 	return nil, nil
 }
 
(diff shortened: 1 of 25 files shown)


nirmay (approved):
Approved.

PR #802, opened 2026-06-11
fix(schemav2validator): strengthen domain allowlist validation

nirmalnr:
### Summary

Improves the robustness of domain allowlist enforcement in `isAllowedDomain` and `validateReferencedObject` to ensure only intended domains and URL schemes are accepted when an allowlist is configured.

### Changes

- Refactored `isAllowedDomain` to accept a pre-parsed `*url.URL` and match against `u.Host` using exact and subdomain-anchored matching instead of raw string matching
- Added scheme enforcement (`http`/`https` only) in `validateReferencedObject` when an allowlist is active
- URL is now parsed once and reused for both the scheme check and the domain check, removing a redundant parse
- Both checks are gated on `len(allowedDomains) > 0`, preserving unrestricted behaviour when no allowlist is configured
- Added `strings.TrimSpace` and empty-entry skip on allowlist entries to handle accidental whitespace in config values

### Tests added

- Host-level domain matching: exact match, subdomain, and invalid host cases
- Scheme validation when allowlist is active (including a case where the host would otherwise pass the domain check)
- Allowlist entry whitespace trimming
- Empty/nil allowlist skips all checks (regression)
- Valid `http://` URL with allowlisted host passes end-to-end (regression)

### DIFF
--- a/pkg/plugin/implementation/schemav2validator/extended_schema.go
+++ b/pkg/plugin/implementation/schemav2validator/extended_schema.go
@@ -488,7 +488,11 @@ func isAllowedDomain(schemaURL string, allowedDomains []string) bool {
 		return false // unparseable or no host (file://, bare path) → deny
 	}
 	for _, domain := range allowedDomains {
-		if strings.Contains(u.Host, domain) {
+		domain = strings.TrimSpace(domain)
+		if domain == "" {
+			continue
+		}
+		if u.Host == domain || strings.HasSuffix(u.Host, "."+domain) {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


nirmalnr:
@711ayush711 Please review.

nirmay:
This looks like a comprehensive fix,

Minor concern: The new isAllowedDomain returns false for an empty allowlist, which is the opposite of the old behavior (true). This is by design (caller now guards), but it's a silent footgun if someone adds a future call site without reading the comment. Consider making the guard explicit inside the function itself (early return true for empty list) and removing the burden from callers — the current split is a bit fragile:

```// Option: keep old semantics inside the function
func isAllowedDomain(u *url.URL, allowedDomains []string) bool {
    if len(allowedDomains) == 0 {
        return true
    }
    ...
}```

The call site in validateReferencedObject already has its own `len(allowedDomains) > 0` gate for the scheme check, so there's no duplication concern.
@nirmalnr

nirmalnr:
@abhey-WIL Please review.

nirmay (approved):
approved.

PR #807, opened 2026-06-15
fix(signvalidator): cross-check signing subscriber identity against request context

### LINKED ISSUE #806: fix: signing subscriber identity not cross-checked against request context
## Summary
During signature validation, the adapter verifies that the `Authorization` header contains a valid signature from a registered subscriber. However, it does not verify that the signing subscriber's identity (extracted from the `keyId` field in the Authorization header) matches the corresponding subscriber ID declared in the request body's `context` object (`context.bap_id` on BAP receivers, `context.bpp_id` on BPP receivers).

## Impact
A registered network participant could sign a request while claiming to be a different network participant in the context, potentially leading to spoofed identity in downstream processing, audit logs, and policy decisions.

## Expected Behavior
After signature validation succeeds, the `subscriber_id` parsed from the `keyId` in the Authorization header should be asserted to equal `context.bap_id` (on a BAP receiver) or `context.bpp_id` (on a BPP receiver).

## Affected Components 
- `signvalidator` plugin (`validateSign` step)

nirmalnr:
## Summary

Closes #806.

During signature validation the adapter verified that the `Authorization` header contained a valid signature from a registered subscriber, but did not assert that the signing subscriber (`subscriber_id` from `keyId`) matched the caller identity declared in the request body (`context.bap_id` on a BPP receiver, `context.bpp_id` on a BAP receiver). A registered participant could therefore sign a request while impersonating a different participant in the context, leading to spoofed identity in downstream processing, audit logs, and policy decisions.

### `pkg/model/model.go`
- Added `ResolveCallerID` — returns the expected signing subscriber ID from a Beckn context map for a given role (opposite-role lookup: BAP→bpp_id/bppId/receiverId, BPP→bap_id/bapId/senderId)
- Added `ResolveSubscriberID` — returns the node's own subscriber ID from a Beckn context map (same-role lookup)
- Both functions handle the triple-key alias chain (legacy snake_case, camelCase, Beckn spec v2 names) and return `""` for Gateway role or missing fields

### `core/module/handler/step.go`
- `validate()` now returns `(*authHeader, error)` so the parsed header is available to the caller without re-parsing
- Added `checkSubscriberIdentity(ctx, *authHeader)` — asserts `parsed.SubscriberID == expected`, where `expected` is resolved via fast path (reqpreprocessor-cached `ContextKeyRemoteID`) or slow path (direct body parse via `model.ResolveCallerID`)
- Check is skipped (with debug log) when no caller ID field is present in the body or the role is Gateway; malformed body is logged and skipped rather than errored, since schema validation handles that separately
- `checkSubscriberIdentity` is called in `validateHeaders` immediately after signature validation succeeds

### `pkg/plugin/implementation/reqpreprocessor/reqpreprocessor.go`
- Replaced the local `firstNonNil(any...)` pattern with `model.ResolveSubscriberID` and `model.ResolveCallerID`
- Removed the `firstNonNil` helper
- Guard conditions changed from `!= nil` to `!= ""` to match the new typed string returns

## Tests

- `pkg/model/model_test.go` (new): 13 table-driven cases for `ResolveCallerID` and 9 for `ResolveSubscriberID`, covering precedence, empty strings, non-string values, and unsupported roles
- `core/module/handler/step_test.go`: 9 new cases covering fast path (context) and slow path (body), both roles, both spec versions (bap_id vs senderId/receiverId), Gateway role skip, missing field skip, and malformed body skip

### DIFF
--- a/core/module/handler/responsestep.go
+++ b/core/module/handler/responsestep.go
@@ -389,7 +389,7 @@ func (v *validateAckSignatureStep) RunOnResponse(ctx *model.StepContext, rctx *m
 		return nil
 	}
 
-	if err := v.signValidator.ValidateAck(ctx, rctx.Body, sigHeader, outboundAuth, publicKey); err != nil {
+	if err := v.signValidator.ValidateAck(ctx, rctx.Body, sigHeader, outboundAuth, publicKey, false); err != nil {
 		log.Warnf(ctx, "validateAckSign: Signature verification failed (status=%d): %v — degraded trust", rctx.StatusCode, err)
 		return nil
 	}
--- a/core/module/handler/responsestep_test.go
+++ b/core/module/handler/responsestep_test.go
@@ -481,11 +481,11 @@ type mockSignValidatorWithAck struct {
 	validateAckErr    error
 }
 
-func (m *mockSignValidatorWithAck) Validate(_ *model.StepContext, _ string, _ string) error {
+func (m *mockSignValidatorWithAck) Validate(_ *model.StepContext, _ string, _ string, _ bool) error {
 	return nil
 }
 
-func (m *mockSignValidatorWithAck) ValidateAck(_ *model.StepContext, _ []byte, _, _, _ string) error {
+func (m *mockSignValidatorWithAck) ValidateAck(_ *model.StepContext, _ []byte, _, _, _ string, _ bool) error {
 	m.validateAckCalled = true
 	return m.validateAckErr
 }
--- a/core/module/handler/responsestep_test.go
+++ b/core/module/handler/responsestep_test.go
@@ -481,11 +481,11 @@ type mockSignValidatorWithAck struct {
 	validateAckErr    error
 }
 
-func (m *mockSignValidatorWithAck) Validate(_ context.Context, _ []byte, _ string, _ string) error {
+func (m *mockSignValidatorWithAck) Validate(_ *model.StepContext, _ string, _ string) error {
 	return nil
 }
 
-func (m *mockSignValidatorWithAck) ValidateAck(_ context.Context, _ []byte, _, _, _ string) error {
+func (m *mockSignValidatorWithAck) ValidateAck(_ *model.StepContext, _ []byte, _, _, _ string) error {
 	m.validateAckCalled = true
 	return m.validateAckErr
 }
--- a/core/module/handler/step.go
+++ b/core/module/handler/step.go
@@ -213,6 +213,10 @@ func (s *validateSignStep) validateHeaders(ctx *model.StepContext) error {
 		ctx.RespHeader.Set(model.UnaAuthorizedHeaderSubscriber, unauthHeader)
 		return model.NewSignValidationErr(fmt.Errorf("failed to validate %s: %w", model.AuthHeaderSubscriber, err))
 	}
+	if err := s.checkSubscriberIdentity(ctx, headerValue); err != nil {
+		ctx.RespHeader.Set(model.UnaAuthorizedHeaderSubscriber, unauthHeader)
+		return model.NewSignValidationErr(err)
+	}
 	return nil
 }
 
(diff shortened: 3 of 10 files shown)


nirmay:
The identity check can be done with significantly fewer changes. The `validate()` method in `step.go` already has everything needed in scope: `ctx.Role`, `ctx.Body`, and `headerVals.SubscriberID` (parsed from `keyId`). The check can live at the bottom of `validate()` itself, right after signature validation passes.

This eliminates the need for:
- `ResolveCallerID` / `ResolveSubscriberID` helpers in `model.go`
- Return type change on `validate()`
- `checkSubscriberIdentity` method
- Any changes to `reqpreprocessor.go`

Suggested change — add this block at the end of `validate()` in `step.go`, before the final `return nil`:

```go
if ctx.Role == model.RoleBPP || ctx.Role == model.RoleBAP {
    var payload struct {
        Context map[string]interface{} `json:"context"`
    }
    if err := json.Unmarshal(ctx.Body, &payload); err == nil && payload.Context != nil {
        var keys []string
        if ctx.Role == model.RoleBPP {
            keys = []string{"bap_id", "bapId", "senderId"}
        } else {
            keys = []string{"bpp_id", "bppId", "receiverId"}
        }
        for _, k := range keys {
            if declared, ok := payload.Context[k].(string); ok && declared != "" {
                if declared != headerVals.SubscriberID {
                    return fmt.Errorf("subscriber identity mismatch: signing subscriber %q does not match declared context identity %q",
                        headerVals.SubscriberID, declared)
                }
                break
            }
        }
    }
}
```

The reqpreprocessor refactor (removing `firstNonNil`) is a separate concern and should either be dropped from this PR or moved to a follow-up, so the security fix stays minimal and easy to audit.

@nirmalnr

nirmalnr:
This fix was initially implemented in `step.go` and has since been refactored and moved into the `signvalidator` plugin, where it now lives alongside the cryptographic verification it extends.

During signature validation the adapter verified that the `Authorization` header contained a valid signature from a registered subscriber, but did not assert that the signing subscriber (`subscriber_id` from `keyId`) matched the caller identity declared in the request body (`context.bap_id` on a BPP receiver, `context.bpp_id` on a BAP receiver). A registered participant could therefore sign a request while impersonating a different participant in the context, leading to spoofed identity in downstream processing, audit logs, and policy decisions.

### Why the check belongs in the sign validator

A signature is only meaningful if it binds to an identity. Cryptographic verification alone proves that *someone* with a registered key signed the payload — but if the signing entity is claiming to be a *different* entity in the message body, the signature is semantically invalid regardless of whether the math checks out. The identity assertion is not an authorization policy layered on top of signature validation; it is part of what it means for a signature to be valid. Keeping this check inside the `signvalidator` plugin ensures the full definition of "this signature is valid" is co-located with the cryptographic verification rather than split across layers.

## Changes

### `pkg/model/model.go`
- Added `ResolveCallerID` — returns the expected signing subscriber ID from a Beckn context map for a given role (opposite-role lookup: BAP→bpp_id/bppId/receiverId, BPP→bap_id/bapId/senderId)
- Added `ResolveSubscriberID` — returns the node's own subscriber ID from a Beckn context map (same-role lookup)
- Both functions handle the triple-key alias chain (legacy snake_case, camelCase, Beckn spec v2 names) and return `""` for Gateway role or missing fields

### `pkg/plugin/implementation/signvalidator/signvalidator.go`
- `Validate` and `ValidateAck` signatures updated: `context.Context` replaced with `*model.StepContext` so the body, role, and request headers are available inside the plugin without re-passing them explicitly
- Added `extractSubscriberIDFromHeader` — extracts the subscriber ID (first `|`-delimited component of `keyId`) from a Beckn Authorization header
- Added `checkSubscriberIdentity` — after cryptographic verification succeeds, asserts that the signer matches the caller identity declared in the request body; resolves the expected identity via fast path (reqpreprocessor-cached `ContextKeyRemoteID`) or slow path (direct body parse via `model.ResolveCallerID`)
- Check is skipped (with debug log) when no caller ID field is present in the body or the role is Gateway; malformed body is logged and skipped rather than errored, since schema validation handles structural correctness separately
- Identity check is gated so it only fires when validating the `Authorization` header, not the `X-Gateway-Authorization` header

### `pkg/plugin/definition/signvalidator.go`
- Updated `SignValidator` interface to match the new `*model.StepContext`-based signatures for `Validate` and `ValidateAck`

### `pkg/plugin/implementation/reqpreprocessor/reqpreprocessor.go`
- Replaced the local `firstNonNil(any...)` helper with `model.ResolveSubscriberID` and `model.ResolveCallerID`
- Guard conditions changed from `!= nil` to `!= ""` to match the new typed string returns

## Tests

- `pkg/model/model_test.go` (new): 13 table-driven cases for `ResolveCallerID` and 9 for `ResolveSubscriberID`, covering precedence, empty strings, non-string values, and unsupported roles
- `pkg/plugin/implementation/signvalidator/signvalidator_test.go`: 9 new cases for `Validate` identity checks and 5 for `ValidateAck` identity checks, covering fast path (context) and slow path (body), both roles, both spec versions (bap_id vs senderId/receiverId), Gateway role skip, missing field skip, malformed body skip, and gate-closed (ACK response path) skip

nirmay (approved):
Approve. The main concern across all review rounds — the gate mechanism — is now properly fixed with the checkIdentity bool parameter. The double-parse is gone. Security fix is correct, layering is clean, test coverage is thorough. The points above are all low-severity and can be follow-up cleanup.

nirmalnr:
## Follow-up cleanups

### 1. Eliminate duplicate header parsing

`checkSubscriberIdentity` called `extractSubscriberIDFromHeader` to extract the subscriber ID from the `keyId` field — re-scanning the same raw string that `parseAuthHeader` had already split into a key-value map. Extended `parseAuthHeader` to return `subscriberID` directly from `signatureMap["keyId"]` in the same pass and deleted `extractSubscriberIDFromHeader`.

### 2. Replace implicit gate with explicit `checkIdentity bool`

The identity check inside `Validate` and `ValidateAck` was gated on `ctx.Request.Header.Get(model.AuthHeaderSubscriber) == header`. This worked because `step.go` happened to pass the raw `Authorization` header value, but it was non-obvious and fragile. The caller already knows whether identity should be checked, so that intent now lives explicitly at the call site as `checkIdentity bool`. Gateway path passes `false`, subscriber path passes `true`, ACK response path passes `false`.

### Changes
- `pkg/plugin/definition/signvalidator.go` — added `checkIdentity bool` to `Validate` and `ValidateAck` interface methods
- `pkg/plugin/implementation/signvalidator/signvalidator.go` — `parseAuthHeader` returns `subscriberID`; removed `extractSubscriberIDFromHeader`; replaced string comparison gate with `if checkIdentity`
- `core/module/handler/step.go` — `validate()` takes `checkIdentity bool`; gateway path passes `false`, subscriber path passes `true`
- `core/module/handler/responsestep.go` — `ValidateAck` call passes `false`
- `core/module/handler/step_test.go`, `responsestep_test.go` — updated mock signatures
- `pkg/plugin/implementation/signvalidator/signvalidator_test.go` — simplified `makeCtx`/`makeCtxWithCallerID` helpers; updated all call sites with explicit bool

PR #816, opened 2026-06-17
fix(router): forward inbound query parameters to upstream URL

### LINKED ISSUE #815: fix: query parameters are silently dropped and not forwarded to upstream
## Problem

When an incoming request has query parameters (e.g. `/beckn/catalog/subscription?subscriptionId=<id>`), ONIX silently discards them. They are **not forwarded** to the upstream BPP/BAP endpoint.

## Root Cause

Three places in the code drop query parameters:

1. **`handleProtocolMapping`** ([router.go:363](https://github.com/beckn/beckn-onix/blob/main/pkg/plugin/implementation/router/router.go#L363)) — constructs `targetURL` then sets only `targetURL.Path` via `joinPath`. `RawQuery` from the incoming request URL is never copied onto `targetURL`.

2. **`routeBodyless`** — same issue for GET/DELETE requests that carry no JSON body.

3. **`proxy` director** ([stdHandler.go:339](https://github.com/beckn/beckn-onix/blob/main/core/module/handler/stdHandler.go#L339)) — sets `req.URL = target` where `target` is the `*url.URL` returned by the router. Since the router never populates `RawQuery`, it is absent at proxy time too.

## Impact

- Any Beckn endpoint that relies on query parameters (e.g. subscription IDs, pagination cursors, filter tokens) will silently fail — the upstream receives the request without the expected parameters.
- No error is returned to the caller; the drop is invisible.

## Expected Behaviour

Query parameters present on the inbound request must be forwarded verbatim to the upstream URL. If the routing rule or registry-resolved URI also carries query parameters, they should be merged (inbound wins on key collision).

## Fix

- In `handleProtocolMapping` and `routeBodyless`: copy `reqURL.RawQuery` onto `targetURL.RawQuery` when building the route.
- In `proxy` director: ensure `req.URL.RawQuery` is set from `ctx.Route.URL.RawQuery` (this will be correct once the router fix is in place, but should be verified).
- Add unit tests in `router_test.go` covering requests with query strings.
- Add integration test in `stdHandler_test.go` asserting query params reach the proxied backend.

## Affected Files

- `pkg/plugin/implementation/router/router.go`
- `core/module/handler/stdHandler.go`
- `pkg/plugin/implementation/router/router_test.go`
- `core/module/handler/stdHandler_test.go`

## Milestone

1.6.0

nirmay:
## Summary

- Query parameters on inbound requests (e.g. `?subscriptionId=<id>`) were silently dropped and never forwarded to the upstream BPP/BAP endpoint
- `handleProtocolMapping` and `routeBodyless` now copy `reqURL.RawQuery` onto the constructed target URL
- Static URL route entries are cloned before setting `RawQuery` to avoid mutating the shared baked-in route object
- The `proxy` director required no change — it sets `req.URL = target`, so populating `RawQuery` on the route URL is sufficient

## Test plan

- [ ] `TestRouteQueryParamsForwarded` — 5 table-driven cases covering `bpp_uri`, `bap_uri`, static URL, empty query, and gateway fallback URL routing
- [ ] `TestRouteBodylessQueryParamsForwarded` — bodyless GET path with `?subscriptionId=test123`
- [ ] `TestProxy_QueryParamsForwardedToUpstream` — end-to-end integration test: real `httptest.Server` asserts the upstream receives `subscriptionId=test123&page=2` verbatim
- [ ] All existing router and handler tests continue to pass

Fixes #815

### DIFF
--- a/pkg/plugin/implementation/router/router.go
+++ b/pkg/plugin/implementation/router/router.go
@@ -299,7 +299,7 @@ func (r *Router) Route(ctx context.Context, reqURL *url.URL, body []byte) (*mode
 	}
 	// For static URL targets, copy inbound query params onto a URL clone so the
 	// upstream receives them. The baked-in URL has no RawQuery of its own.
-	if route.TargetType == targetTypeURL && rawQuery != "" && route.URL != nil {
+	if rawQuery != "" && route.TargetType == targetTypeURL && route.URL != nil {
 		clone := *route.URL
 		clone.RawQuery = rawQuery
 		return &model.Route{TargetType: targetTypeURL, URL: &clone}, nil
@@ -340,7 +340,7 @@ func (r *Router) routeBodyless(endpoint, rawQuery string) (*model.Route, error) 
 		if rawQuery != "" && route.TargetType == targetTypeURL && route.URL != nil {
 			clone := *route.URL
 			clone.RawQuery = rawQuery
-			return &model.Route{TargetType: route.TargetType, URL: &clone}, nil
+			return &model.Route{TargetType: targetTypeURL, URL: &clone}, nil
 		}
 		return route, nil
 	}
@@ -386,10 +386,7 @@ func handleProtocolMapping(route *model.Route, npURI, endpoint, rawQuery string)
 	if rawQuery != "" {
 		targetURL.RawQuery = rawQuery
 	}
-	return &model.Route{
-		TargetType: targetTypeURL,
-		URL:        targetURL,
-	}, nil
+	return &model.Route{TargetType: targetTypeURL, URL: targetURL}, nil
 }
 
 func joinPath(u *url.URL, endpoint string) string {
--- a/pkg/plugin/implementation/router/router.go
+++ b/pkg/plugin/implementation/router/router.go
@@ -383,7 +383,9 @@ func handleProtocolMapping(route *model.Route, npURI, endpoint, rawQuery string)
 		return nil, fmt.Errorf("invalid %s URI - %s in request body for %s: %w", role, target, endpoint, err)
 	}
 	targetURL.Path = joinPath(targetURL, endpoint)
-	targetURL.RawQuery = rawQuery
+	if rawQuery != "" {
+		targetURL.RawQuery = rawQuery
+	}
 	return &model.Route{
 		TargetType: targetTypeURL,
 		URL:        targetURL,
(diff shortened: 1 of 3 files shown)


pkg/plugin/implementation/router/router.go
```diff
@@ -288,9 +293,16 @@ func (r *Router) Route(ctx context.Context, reqURL *url.URL, body []byte) (*mode
...
 	case targetTypeBPP, targetTypeReceiver:
-		return handleProtocolMapping(route, bppURI, endpoint)
+		return handleProtocolMapping(route, bppURI, endpoint, rawQuery)
 	case targetTypeBAP, targetTypeSender:
-		return handleProtocolMapping(route, bapURI, endpoint)
+		return handleProtocolMapping(route, bapURI, endpoint, rawQuery)
+	}
+	// For static URL targets, copy inbound query params onto a URL clone so the
+	// upstream receives them. The baked-in URL has no RawQuery of its own.
+	if route.TargetType == targetTypeURL && rawQuery != "" && route.URL != nil {
```
nirmalnr: There is already a switch for route.TargetType above. Should the if be moved inside?

nirmalnr:
The above is just a minor styling related nit. Tested the flow and can see the query parameters being carried from ONIX to ONIX as well as from ONIX to backend service.

pkg/plugin/implementation/router/router_test.go
```diff
@@ -1015,3 +1015,134 @@ func TestRouteBodyless(t *testing.T) {
...
+	_, err := router.Route(context.Background(), nil, []byte(`{}`))
+	if err == nil || !strings.Contains(err.Error(), "reqURL must not be nil") {
+		t.Errorf("Route(nil URL) = %v, want error containing 'reqURL must not be nil'", err)
+	}
+}
+
+// TestRouteBodylessPublisherUnaffectedByQueryParams confirms that publisher-type
+// bodyless routes are returned unchanged when RawQuery is present — publisher
+// routes carry no URL, so there is nothing to attach the query string to.
+func TestRouteBodylessPublisherUnaffectedByQueryParams(t *testing.T) {
```
nirmalnr, line 1129: The test mentions body less but body is defined `{"context":{"domain":"ONDC:TRV10","version":"1.1.0"}}`

nirmalnr: I think there is a gap here as well. A nil-body request where the matched rule has targetType: publisher and rawQuery is non-empty. That hits return route, nil directly, and no test exercises it.

nirmalnr (approved):
All comments have been addressed and the PR has been tested locally. Good to merge.

nirmay:
Both issues are fixed in subsequent commits, pls review @nirmalnr

PR #819, opened 2026-06-22
fix(signvalidator): apply configurable clock-skew tolerance to created field

### LINKED ISSUE #818: fix(signvalidator): apply configurable clock-skew tolerance to `created` field per Auth & Trust spec
## Problem

The Beckn Auth & Trust spec defines two distinct timestamp validation rules:

1. **`created` — clock skew tolerance applies.** The recommended default is **5 seconds**. NFOs MAY configure a stricter or more permissive value for their subnet; all NPs in that subnet MUST use the configured value.
2. **`expires` — zero tolerance, always.** An expired signature MUST be rejected unconditionally, regardless of any subnet configuration.

`checkTimestampWindow` in `signvalidator` currently applies **zero tolerance to both fields**:

```go
if createdTimestamp > current {
    return model.NewSignValidationErr(...)   // rejects if created is even 1s in the future
}
if current > expiredTimestamp {
    return model.NewSignValidationErr(...)   // correct — strict
}
```

This means two NPs whose clocks differ by even a few seconds will fail signature validation despite exchanging cryptographically valid messages. There is also no way for an NFO to configure the subnet-level tolerance that the protocol mandates.

## Expected behaviour

| Field | Rule |
|---|---|
| `created` | Accept if `created ≤ now + clockSkewTolerance` (default 5 s, configurable per deployment) |
| `expires` | Reject strictly if `now > expires` — no tolerance, no config knob |

## Proposed fix

1. Add `ClockSkewTolerance time.Duration` to `signvalidator.Config` with a default of **5 seconds**.
2. Change the `created` check to: `if time.Unix(createdTimestamp, 0).After(now.Add(config.ClockSkewTolerance))`.
3. Keep the `expires` check unchanged (strict zero tolerance).
4. Expose the field in the adapter YAML plugin config block so NFOs can override it per subnet deployment.

## References

- Beckn Auth & Trust spec — Clock Skew section
- `pkg/plugin/implementation/signvalidator/signvalidator.go` — `checkTimestampWindow` (line 190)
- Related: fix(signvalidator): split timestamp check into distinct expiry/not-yet (#803)

nirmay:
Fixes #818

## Summary

- `checkTimestampWindow` now accepts a `clockSkewTolerance` parameter and applies it **only** to the `created` field — accepting signatures where `created ≤ now + tolerance` to accommodate NTP drift between network participants
- `expires` retains strict zero tolerance with no config knob, per the Beckn Auth & Trust spec ("MUST be rejected without exception")
- `ClockSkewTolerance` added to `signvalidator.Config`, defaulting to **5 s** when unset
- Startup warning logged if configured tolerance exceeds **10 s** (wider replay window)
- New config key `clockSkewToleranceSeconds` (integer, seconds) exposed in the plugin config block; validated at startup (non-negative integer only)
- CONFIG.md updated to document the new parameter

## Test plan

- [x] `go test ./pkg/plugin/implementation/signvalidator/...` passes
- [x] `created` 3 s in future accepted with default 5 s tolerance
- [x] `created` 7 s in future rejected with default 5 s tolerance
- [x] `expires` 1 s in past rejected even with large tolerance configured
- [x] Custom 10 s tolerance accepted via config
- [x] Invalid config values (`"5s"`, `"-1"`) rejected at startup
- [x] Deploy with no `config:` block — default 5 s behaviour confirmed

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -807,7 +807,7 @@ Auxiliary specs are additive — they may only introduce new action verbs not al
 signValidator:
   id: signvalidator
   config:
-    clockSkewToleranceSeconds: "5"   # optional — see parameters below
+    clockSkewToleranceSeconds: 5   # optional — see parameters below
 ```
 
 **Parameters**:
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -806,9 +806,16 @@ Auxiliary specs are additive — they may only introduce new action verbs not al
 ```yaml
 signValidator:
   id: signvalidator
+  config:
+    clockSkewToleranceSeconds: "5"   # optional — see parameters below
 ```
 
-**Parameters**: None required. Uses key manager for public key lookup.
+**Parameters**:
+
+- `clockSkewToleranceSeconds` *(optional, integer, default: `5`)*: Maximum number of seconds the `created` timestamp in an inbound signature header may be ahead of the server clock. Accommodates NTP drift between network participants. Per the Beckn Auth & Trust spec the recommended default is 5 s; NFOs may configure a stricter or more permissive value for their subnet.
+  - A value of `"0"` enforces strict same-second validation.
+  - Values greater than `10` are accepted but trigger a startup warning — large tolerances widen the replay window.
+  - **The `expires` field always uses zero tolerance regardless of this setting.** An expired signature is rejected unconditionally.
 
 ---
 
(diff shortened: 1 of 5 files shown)


nirmalnr (approved):
The core design is sound and spec-aligned. PR has been tested and good to merge.

PR #831, opened 2026-06-29
feat(plugin): add vcvalidator Verifiable Credential step

ameetdesh:
## Summary

Adds a new **`vcvalidator`** processing Step plugin that verifies the W3C Verifiable Credentials embedded in a request body. For the configured beckn action(s) it verifies every embedded credential and rejects the request with a **NACK** (before routing) when verification fails.

This plugin has been running as an out-of-tree plugin bundled into the `onix-adapter-deg` image; this PR upstreams it so it builds and ships like any other beckn-onix plugin.

## What it checks

For every embedded credential (any JSON object carrying both `proof` and `credentialSubject`):

1. **Proof signature** — verifies a VC-JWT (`proof.jwt`) against the issuer DID's public key, resolved from:
   - `did:key` — `Ed25519` (`z6Mk…`), `P-256` (`zDn…`), `secp256k1` (`zQ3…`)
   - `did:jwk` — embedded JWK
   - `did:web` — fetches `https://<host>/[path/]did.json`
2. **Issuer binding** — the JWT signer (`kid` controller) must equal the declared `issuer`, and the header `alg` must match the resolved key (alg-confusion protection).
3. **Validity window** — `validFrom`/`validUntil` and JWT `nbf`/`exp`.
4. **Revocation** — `credentialStatus` via StatusList2021/BitstringStatusList, DEDI, or a generic revoked indicator.

JSON-LD Data Integrity proofs (`proofValue`, requiring URDNA2015 canonicalisation) are not verified; with the default `requireProof: true` such credentials are rejected.

## Rejections go through the standard signed-NACK path

The plugin implements `definition.StepProvider` and runs inside the handler's step pipeline, so a rejection is just an error returned from `Run` — the handler builds and **signs** the NACK exactly as it does for `validateSign` / `validateSchema`. There is no plugin-local NACK envelope or HTTP status mapping.

Failures map to the standard model error types, with the machine-readable failure class preserved at the start of the NACK error message:

| failure class | meaning | model error → HTTP |
|------|---------|------|
| `INVALID_CREDENTIAL` | malformed / missing issuer | `BadReqErr` → 400 |
| `INVALID_PROOF` | bad/missing signature or alg mismatch | `SignValidationErr` → 401 |
| `ISSUER_MISMATCH` | proof signer ≠ issuer | `SignValidationErr` → 401 |
| `CREDENTIAL_EXPIRED` | outside validity window | `SignValidationErr` → 401 |
| `DID_RESOLUTION_FAILED` | issuer/VM DID did not resolve | `SignValidationErr` → 401 |
| `CREDENTIAL_REVOKED` | revoked per `credentialStatus` | `SignValidationErr` → 401 |

## Conformance to plugin conventions

- Implements `definition.StepProvider`; `cmd/plugin.go` exports `Provider` (with a compile-time interface assertion).
- Two implementation files, matching sibling plugin layout: `vcvalidator.go` (Step, Config, credential extraction) and `verify.go` (verification engine: proof/JWT checks, DID resolution, revocation, base58).
- Logging via `pkg/log`.
- Standard layout otherwise: `cmd/`, `README.md`, table-driven tests, `testdata/`.
- Added to `install/build-plugins.sh` → `vcvalidator.so`.
- Wired as a `plugins.steps` entry referenced by id in the handler's `steps` list, typically after `validateSign` (see plugin README for config).

## Dependencies

The only new direct dependencies — `github.com/lestrrat-go/jwx/v3` and `github.com/decred/dcrd/dcrec/secp256k1/v4` — were already present as **indirect** requirements; `go mod tidy` only promotes them. No new modules are introduced.

## Tests

Fully offline (`go test ./pkg/plugin/implementation/vcvalidator/...`). Includes committed **mock credentials covering did:key, did:jwk and did:web in both revoked and not-revoked states**, plus a real externally-issued `did:key` (P-256) VC-JWT, and negative tests (tampered signature, expired/not-yet-valid, issuer mismatch, did:web unreachable fail-closed/fail-open, DEDI revocation, Data Integrity rejection). Step-level tests cover pass-through (disabled, non-gated action, no credentials) and the rejection-to-model-error mapping. The mock vectors are reproducibly regenerated via `testdata/gen` (fixed key seeds + stable JSON ordering).

Verified locally: all tests pass and `go build -buildmode=plugin` produces `vcvalidator.so`.


nirmay:
@ameetdesh thanks for this work, this is very useful plugin.

I have 2 high level points

1. The proposed design uses the middleware plugin for VC. Because middleware sits outside the Step pipeline, vcvalidator can't use h.signNackResponse. So it hand-rolls its own NACK envelope in pkg/plugin/implementation/vcvalidator/middleware.go (nackResponse struct, manual json.Marshal, manual status-code mapping) — and that NACK is unsigned, unlike every other NACK in the system (validateSignStep, validateSchemaStep) which goes through the signed envelope path. That's an inconsistency I'd raise with the author: This should probably be a Step so rejections go through the standard signed-NACK path.
2. the plugin code is highly fragmented into many files. Just for consistency with other plugin implementations, we should try to consolidate that into 1-2 files.

I will do a detailed review and trials on this later.

Thanks
Mayuresh

ameetdesh:
> @ameetdesh thanks for this work, this is very useful plugin.
> 
> I have 2 high level points
> 
> 1. The proposed design uses the middleware plugin for VC. Because middleware sits outside the Step pipeline, vcvalidator can't use h.signNackResponse. So it hand-rolls its own NACK envelope in pkg/plugin/implementation/vcvalidator/middleware.go (nackResponse struct, manual json.Marshal, manual status-code mapping) — and that NACK is unsigned, unlike every other NACK in the system (validateSignStep, validateSchemaStep) which goes through the signed envelope path. That's an inconsistency I'd raise with the author: This should probably be a Step so rejections go through the standard signed-NACK path.
> 2. the plugin code is highly fragmented into many files. Just for consistency with other plugin implementations, we should try to consolidate that into 1-2 files.
> 
> I will do a detailed review and trials on this later.
> 
> Thanks Mayuresh

Hi @nirmay 
addressed your comments (moved from middleware to step plugin, consolidated main files). Thanks. I also removed this plugin from DEG repo here https://github.com/beckn/DEG/pull/420, rebuilt with the beckn-onix from this branch, and tested with payloads from DEG/devkits/data-exchange/uc1-meter-data pasting below.

1. did:web revoked: 
```
{
    "message": {
        "status": "NACK",
        "messageId": "cba1fdd8-cafc-4742-84e5-a089da6b7fc6",
        "error": {
            "code": "Unauthorized",
            "message": "Signature Validation Error: CREDENTIAL_REVOKED: credential revoked via https://api.dedi.global/dedi/lookup/ameetdesh.github.io/vc-revocation-registry"
        }
    }
}
```

2. did:web unrevoked
```
{
    "message": {
        "status": "ACK",
        "messageId": "6d11ac3f-392f-410d-b124-d7226ef87a05"
    }
}
```

3. did:web without revocation status link:
```
{
    "message": {
        "status": "ACK",
        "messageId": "ceea790c-7a89-4f57-be3a-54da48d048e9"
    }
}
```

4. expired
```
{
    "message": {
        "status": "NACK",
        "messageId": "6bc484a5-3c07-4e51-97a4-2bf09b7c6388",
        "error": {
            "code": "Unauthorized",
            "message": "Signature Validation Error: CREDENTIAL_EXPIRED: credential expired (validUntil=2026-01-01T00:00:00Z)"
        }
    }
}
```

5. tampered
```
{
    "message": {
        "status": "NACK",
        "messageId": "2e6f5e6a-25bf-4887-b769-f3edac4a9672",
        "error": {
            "code": "Unauthorized",
            "message": "Signature Validation Error: INVALID_PROOF: signature verification failed: jws.Verify: could not verify message using any of the signatures or keys: jws.Verify: failed to verify signature #1 with key *ecdsa.PublicKey: invalid ECDSA signature\njws.Verify: signature #1: tried 1 key(s) but none verified successfully"
        }
    }
}
```

6. wrong-issuer
```
{
    "message": {
        "status": "NACK",
        "messageId": "0da5a9ed-7c47-41af-b44c-93f1474e43f1",
        "error": {
            "code": "Unauthorized",
            "message": "Signature Validation Error: ISSUER_MISMATCH: proof signer \"did:key:zDnaeVtysyVoSjwzhpHXEaLqGB2T5KgfDtANX9D5KN6bcMHR8\" does not match issuer \"did:key:z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH\""
        }
    }
}
```

nirmay:
Nice work on the Step conversion — confirmed the signed-NACK path, ran the full suite, and the consolidated verify.go diffs clean against the original split files (only real change is swapping two hand-rolled helpers for stdlib bytes.HasPrefix/min).

A few things before merge:

1. Step naming. Every built-in step in the pipeline is a verb (validateSign, validateSchema, addRoute). vcvalidator reads as a noun/package name and stands out in the steps: list and in traces/metrics (the id is used directly as the span/telemetry label). Suggest renaming the plugin id (and ideally the exported New) to something like validateVC or verifyCredential to match convention.

2. Docs are missing in the two central references. Every other plugin — including recent ones like ManifestLoader/PayloadStore — gets an entry in README.md "Plugin Types" and a numbered section in CONFIG.md. vcvalidator has neither; right now it's only documented in its own plugin README. Please add both, plus a wiring example in one of the config/*.yaml reference configs (not just the plugin README).

3. SSRF via attacker-influenced fetch URLs. Both did:web resolution (didWebURL → fetch) and revocation checks (statusListRevoked/genericRevoked → fetch(ctx, e.StatusListCredential)) issue outbound GETs to URLs taken directly from the request body (the credential's issuer or credentialStatus.statusListCredential/id), with no hostname/IP allowlist and a default http.Client that follows redirects. This step runs after validateSign, so it requires a valid signature, but any legitimate-or-compromised network participant can use it to make the adapter fetch arbitrary internal URLs (including cloud metadata endpoints). Recommend at minimum: block private/loopback/link-local destinations, and disable redirect-following (or cap+re-validate each hop).

4. Unbounded credential count per request. Run loops over every extracted credential sequentially, each potentially costing up to httpTimeout (default 10s) for did:web resolution plus another for revocation. Nothing caps how many embedded credentials a single request can carry, so a crafted request could tie up a handler goroutine for a long time. Suggest a maxCredentials config knob with a sane default.
@ameetdesh

ameetdesh:
All four points addressed in f474008:

1. **Step naming** — the plugin now builds as `validateVC.so`, so pipelines wire it as the `validateVC` step (the .so basename is the plugin id, hence the span/telemetry label). `install/build-plugins.sh` gained a `<dir>:<output>` entry form for this; the Go package stays `vcvalidator`, and the exported constructor was already just `New`.

2. **Docs** — added the README.md "Plugin Types" bullet and a numbered CONFIG.md "Available Plugins" section, plus a wiring example in `config/local-dev.yaml` (bppTxnReceiver, shipped disabled so the reference config's behavior is unchanged).

3. **SSRF** — the shared HTTP client now rejects private, loopback, link-local (incl. 169.254.169.254), multicast, unspecified and broadcast destinations. The check runs in the dialer's `Control` hook on the resolved IP, so it also covers DNS rebinding, and redirects are capped at 3 hops with every hop re-validated by the same guard. A new `allowPrivateNetworks` knob (default `false`) exists for local/devkit deployments whose issuers and registries live on a private docker network.

4. **Credential cap** — new `maxCredentials` knob (default 10); a request exceeding it is rejected with a Bad Request NACK before any network I/O.

One correction on the threat framing: the step running after `validateSign` is a config choice, not a guarantee — some deployments wire it first — so the SSRF guard doesn't assume an authenticated caller.

New tests cover the dial guard (blocked/allowed IP table + live loopback block), the redirect cap, the credential cap boundary, and the new config knobs; full suite passes.

nirmay (approved):
approved.

PR #838, opened 2026-07-01
test: improve unit test coverage for pkg/beckndefaults

### LINKED ISSUE #814: chore: improve test coverage for pkg/beckndefaults
## Summary

Improve unit test coverage for the `pkg/beckndefaults` package to ensure reliability and catch regressions.

## Scope

Package: `pkg/beckndefaults`

areebahmeddd:
## Summary

* adds 8 new unit tests to `pkg/beckndefaults`, increasing coverage from 60.5% to 95.3%
* covers happy paths, error cases, and edge cases across `fetchFromURLs`, `fetchRemote`, `Load` and `loadAndVerify`
* the only remaining uncovered path (~5%) is when remote fetch succeeds but signature verification fails, which would require refactoring `fetchRemote` to allow injecting a mock url

Closes #814

## AI Model Disclosure

used vscode copilot (claude sonnet 4.6) to help identify missing test cases and improve coverage. changes were self-reviewed and verified with:

```bash
go test ./pkg/beckndefaults/... -v -cover
```

### DIFF
--- a/pkg/beckndefaults/beckndefaults.go
+++ b/pkg/beckndefaults/beckndefaults.go
@@ -25,9 +25,12 @@ var shippedConstantsSig []byte
 //go:embed beckn_public_key.pem
 var becknPublicKeyPEM []byte
 
-const (
+var (
 	remoteConstantsURL    = "https://raw.githubusercontent.com/beckn/beckn-onix/main/pkg/beckndefaults/beckn-constants.yaml"
 	remoteConstantsSigURL = "https://raw.githubusercontent.com/beckn/beckn-onix/main/pkg/beckndefaults/beckn-constants.yaml.sig"
+)
+
+const (
 	remoteTimeout         = 10 * time.Second
 	maxConstantsFileBytes = 64 << 10 // 64 KiB — well above any realistic constants file
 )
--- a/pkg/beckndefaults/beckndefaults.go
+++ b/pkg/beckndefaults/beckndefaults.go
@@ -25,12 +25,9 @@ var shippedConstantsSig []byte
 //go:embed beckn_public_key.pem
 var becknPublicKeyPEM []byte
 
-var (
+const (
 	remoteConstantsURL    = "https://raw.githubusercontent.com/beckn/beckn-onix/main/pkg/beckndefaults/beckn-constants.yaml"
 	remoteConstantsSigURL = "https://raw.githubusercontent.com/beckn/beckn-onix/main/pkg/beckndefaults/beckn-constants.yaml.sig"
-)
-
-const (
 	remoteTimeout         = 10 * time.Second
 	maxConstantsFileBytes = 64 << 10 // 64 KiB — well above any realistic constants file
 )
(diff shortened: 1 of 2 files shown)


nirmay:
Thanks for this — real coverage gain (60.5% → 95.3%) and the new tests are well-targeted and correctly asserted. One thing I'd like fixed before merging:

**`TestLoad_RemoteRefresh` makes a live network call to `raw.githubusercontent.com`, and our CI doesn't skip it.**

Every other test in this file (existing and new) uses `httptest.NewServer` to stay hermetic — this is the only one that reaches the real internet. It's gated behind `testing.Short()`, but our CI runs `go test -coverprofile=coverage.out ./...` with no `-short` flag ([[ci.yml:23](https://github.com/beckn/beckn-onix/blob/main/.github/workflows/ci.yml#L23)](https://github.com/beckn/beckn-onix/blob/main/.github/workflows/ci.yml#L23)), so that guard never actually engages — the live call always happens.

Concretely, this means:
- If GitHub is rate-limited, briefly unreachable, or the CI runner has restricted egress, this one test fails and blocks unrelated PRs, with no actual code defect behind it.
- It also asserts `bc.Version == "v1"` against the *remote* file on `main` — if `becknConstantsVersion` ever gets bumped there in an unrelated change, this test breaks for a reason that has nothing to do with the code it's testing.

Given the PR description already flags that covering "remote fetch succeeds but signature verification fails" needs a way to inject a mock URL into `fetchRemote`, could we fold this into the same small refactor and point `TestLoad_RemoteRefresh` at an `httptest.NewServer` instead? That would close both gaps with one DI change and keep the whole suite hermetic.

Everything else looks good — `gofmt`/`go vet` clean, no other issues found.
@areebahmeddd

areebahmeddd:
sure, let me refactor the live test with httptest.NewServer way 👍🏻

nirmay:
Thanks for iterating on this — the switch to `httptest.NewServer` for `TestLoad_RemoteRefresh` and the new `TestLoad_RemoteRefreshBadSig_FallsBackToShipped` test are exactly right, and coverage is now at 97.7% with no live network calls (verified locally, including under `-race`). Nice work closing that gap.

One thing I'd like changed before merging, though:

**Don't turn `remoteConstantsURL` / `remoteConstantsSigURL` into `var` just to make them test-injectable.**

These two constants are security-relevant — they're the only place we decide where the binary reaches out to fetch the file that later gets signature-verified and injected into plugin configs. Right now the change is safe in practice (they're unexported, and nothing in production code reassigns them), but making them mutable weakens a guarantee the compiler used to enforce for us, purely for test convenience. That's the kind of thing that's fine today and a foot-gun later — a future helper, debug hook, or careless refactor now has a legal path to repoint that fetch at runtime, where before it simply couldn't.

Suggest moving the seam one level up instead, at `fetchRemote`, which is already just a one-line wrapper:

```go
var fetchRemote = func(ctx context.Context) ([]byte, []byte, error) {
    return fetchFromURLs(ctx, remoteConstantsURL, remoteConstantsSigURL)
}
```

Tests override the `fetchRemote` func var to point at the test server; `remoteConstantsURL`/`remoteConstantsSigURL` stay `const`; `Load`'s call site is untouched. Same testability, without loosening the invariant on the URLs themselves.

Everything else looks good — `gofmt`/`go vet` clean, tests are well-targeted and correctly assert on the paths they claim to cover.

areebahmeddd:
updated! 🙂

nirmay (approved):
approved.

PR #839, opened 2026-07-01
fix: validate context.network_id against caller network memberships 

### LINKED ISSUE #837: fix(signvalidator): validate context.networId against caller memberships and receiver allowlist
## Problem

Onix currently validates that the signing subscriber belongs to a network in `allowedNetworkIDs` (registry membership check in `dediregistry`), but it never cross-checks the `context.network_id` field in the request body against the caller's actual network memberships.

This means a caller registered on NFO1 can send a request with `context.network_id: NFO2` and it will pass validation — the body's declared network identity is entirely unverified.

## Required Behavior

A request should be allowed only when **both** of the following hold:

1. **Allowlist check** (existing): if `allowedNetworkIDs` is configured, the caller's `network_memberships` must intersect it.
2. **Context network check** (new): if `context.network_id` is present, it must appear in both the caller's `network_memberships` **and** the configured `allowedNetworkIDs` (if set).

## Decision Table

| # | Caller memberships | Receiver allowlist | context.network_id | Current | Required | Reason |
|---|--------------------|--------------------|-------------------|---------|----------|--------|
| 1 | NFO1 | NFO1 | NFO1 | allow | allow | Caller belongs to declared network; allowlist permits it |
| 2 | NFO1 | NFO1 | NFO2 | allow | **block** | Caller doesn't belong to declared network |
| 3 | NFO1 | NFO1 | nil | allow | allow | No network declared; caller is in allowlist |
| 4 | NFO1 | NFO2 | NFO1 | block | block | Caller not in allowlist |
| 5 | NFO1 | NFO2 | NFO2 | block | block | Caller not in allowlist |
| 6 | NFO1 | NFO2 | nil | block | block | Caller not in allowlist |
| 7 | NFO1 | nil | NFO1 | allow | allow | No allowlist; caller belongs to declared network |
| 8 | NFO1 | nil | NFO2 | allow | **block** | Caller doesn't belong to declared network |
| 9 | NFO1 | nil | nil | allow | allow | No restrictions applicable |
| 10 | NFO1, NFO2 | NFO1 | NFO1 | allow | allow | Declared network in both caller memberships and allowlist |
| 11 | NFO1, NFO2 | NFO1 | NFO2 | allow | **block** | Declared network not in allowlist |
| 12 | NFO1, NFO2 | NFO1 | nil | allow | allow | No network declared; caller is in allowlist |
| 13 | (none) | NFO1 | any | block | block | Caller not in allowlist |
| 14 | (none) | nil | NFO1 | allow | **block** | Caller doesn't belong to network declared in context |
| 15 | (none) | nil | nil | allow | allow | No restrictions applicable |
| 16 | NFO1 | NFO1, NFO2 | NFO2 | allow | **block** | Caller doesn't belong to network declared in context |
| 17 | NFO1 | NFO1, NFO2 | NFO1 | allow | allow | Declared network in both caller memberships and allowlist |
| 18 | NFO1 | NFO1, NFO2 | nil | allow | allow | No network declared; caller is in allowlist |

## Implementation Notes

- The new check belongs in the sign validator step, after the registry lookup resolves the caller's `network_memberships`.
- `context.network_id` being absent (`nil` or empty) must remain valid — not all Beckn flows require it.
- The check must apply to both the `Validate` and `ValidateAck` paths.

nirmalnr:
Closes #837

## Summary

- A caller registered on `nfo1.com/retail` could send a request with `context.network_id: nfo2.com/retail` and pass signature validation. Onix checked that the signing subscriber belonged to an allowed network, but never cross-checked the request body's `context.network_id` against the subscriber's actual `network_memberships` from the registry.

- `dediregistry.Lookup` now calls `validateMemberships` on both the cache-hit and HTTP paths, which enforces two guards in sequence:
  1. Static guard — subscriber's `network_memberships` must intersect `allowedNetworkIDs` (pre-existing check, now shared)
  2. Per-request guard — `context.network_id` (when present) must be in the subscriber's `network_memberships` **and** in `allowedNetworkIDs`

- `reqpreprocessor` extracts `network_id`/`networkId` from the request body and stores it as `model.ContextKeyNetworkID` in the Go context before any OTel span wrapping. `extractContextNetworkID` reads this value first, with a fallback to direct body parsing via type assertion.

- `simplekeymanager.LookupNPKeys` was wrapping the full `context.Context` interface with an OTel span before passing it to `Registry.Lookup`, causing `ctx.(*model.StepContext)` to silently fail downstream. Fixed by wrapping only the embedded `sc.Context` field so `*model.StepContext` is preserved as the outermost type passed to the registry.

- Cache key changed from `lookup_<sub>_<key>` to `dedi_lookup_<sub>_<key>` to invalidate pre-deploy entries that have no `network_memberships` field.

## Test plan

- [ ] `TestContextNetworkIDValidation` — 16 of 18 decision table rows. The two omitted rows (12 and 18) are the network_id-absent variants of groups already covered by rows 3 and 9 — when `network_id` is empty the per-request guard exits immediately, so they exercise no new code path.
- [ ] `TestDeDiRegistryClient_Lookup_Cache` — 9 sub-tests including cache-hit path for both guards active simultaneously
- [ ] `TestExtractContextNetworkID` — context value primary path, body fallback path, precedence, absent cases
- [ ] `go test ./...` — full suite green

### DIFF
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -176,6 +176,18 @@ func (r *Role) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+// ResolveNetworkID returns context.network_id from a parsed Beckn context map,
+// trying "network_id" (snake_case) then "networkId" (camelCase).
+// Returns "" when absent, empty, or not a string under either alias.
+func ResolveNetworkID(reqContext map[string]interface{}) string {
+	for _, k := range []string{"network_id", "networkId"} {
+		if v, ok := reqContext[k].(string); ok && v != "" {
+			return v
+		}
+	}
+	return ""
+}
+
 // ResolveCallerID returns the ID of the other party in a Beckn exchange — i.e. who
 // sent the inbound message — from a parsed Beckn context map.
 // For BPP role: the caller is the BAP (bap_id / bapId / senderId).
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -20,17 +20,17 @@ type Subscriber struct {
 
 // Subscription represents subscription details of a network participant.
 type Subscription struct {
-	Subscriber       `json:",inline"`
-	KeyID            string    `json:"key_id,omitzero" format:"uuid"`
-	SigningPublicKey string    `json:"signing_public_key,omitzero"`
-	EncrPublicKey    string    `json:"encr_public_key,omitzero"`
-	ValidFrom        time.Time `json:"valid_from,omitzero" format:"date-time"`
-	ValidUntil       time.Time `json:"valid_until,omitzero" format:"date-time"`
-	Status           string    `json:"status,omitzero" enum:"INITIATED,UNDER_SUBSCRIPTION,SUBSCRIBED,EXPIRED,UNSUBSCRIBED,INVALID_SSL"`
-	Created          time.Time `json:"created,omitzero" format:"date-time"`
-	Updated          time.Time `json:"updated,omitzero" format:"date-time"`
-	Nonce              string   `json:"nonce,omitzero"`
-	NetworkMemberships []string `json:"network_memberships,omitempty"`
+	Subscriber         `json:",inline"`
+	KeyID              string    `json:"key_id,omitzero" format:"uuid"`
+	SigningPublicKey   string    `json:"signing_public_key,omitzero"`
+	EncrPublicKey      string    `json:"encr_public_key,omitzero"`
+	ValidFrom          time.Time `json:"valid_from,omitzero" format:"date-time"`
+	ValidUntil         time.Time `json:"valid_until,omitzero" format:"date-time"`
+	Status             string    `json:"status,omitzero" enum:"INITIATED,UNDER_SUBSCRIPTION,SUBSCRIBED,EXPIRED,UNSUBSCRIBED,INVALID_SSL"`
+	Created            time.Time `json:"created,omitzero" format:"date-time"`
+	Updated            time.Time `json:"updated,omitzero" format:"date-time"`
+	Nonce              string    `json:"nonce,omitzero"`
+	NetworkMemberships []string  `json:"network_memberships,omitempty"`
 }
 
 // RegistryMetadata represents metadata configured on a registry itself rather than on a specific record.
@@ -241,12 +241,12 @@ type Keyset struct {
 // StepContext holds context information for a request processing step.
 type StepContext struct {
 	context.Context
-	Request         *http.Request
-	Body            []byte
-	Route           *Route
-	SubID           string
-	Role            Role
-	RespHeader      http.Header
+	Request              *http.Request
+	Body                 []byte
+	Route                *Route
+	SubID                string
+	Role                 Role
+	RespHeader           http.Header
 	ProtocolVersion      string // Protocol version parsed from context.version (e.g. "2.0.0")
 	MessageID            string // Message ID parsed from context.messageId in the request body
 	InboundAuthSignature string // Raw Base64 signature from the inbound Authorization header's signature="..." attribute
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -29,7 +29,8 @@ type Subscription struct {
 	Status           string    `json:"status,omitzero" enum:"INITIATED,UNDER_SUBSCRIPTION,SUBSCRIBED,EXPIRED,UNSUBSCRIBED,INVALID_SSL"`
 	Created          time.Time `json:"created,omitzero" format:"date-time"`
 	Updated          time.Time `json:"updated,omitzero" format:"date-time"`
-	Nonce            string    `json:"nonce,omitzero"`
+	Nonce              string   `json:"nonce,omitzero"`
+	NetworkMemberships []string `json:"network_memberships,omitempty"`
 }
 
 // RegistryMetadata represents metadata configured on a registry itself rather than on a specific record.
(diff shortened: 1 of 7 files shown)


nirmay:
## PR #839 Review — `fix: validate context.network_id against caller network memberships`

**What it does:** Adds a per-request guard so that `context.network_id` in an inbound Beckn payload must match the signing subscriber's actual `network_memberships` (from the registry) and the configured `allowedNetworkIDs`, closing a spoofing hole where a caller from one network could claim to be on another. Implemented via a new `dediregistry.validateMemberships`/`validateContextNetworkID`, a `reqpreprocessor` extraction step that stashes `network_id` into the request context, and a `simplekeymanager` fix so OTel span-wrapping doesn't erase the `*model.StepContext` type. Comes with substantial new test coverage (18-row decision table, cache-hit paths).

### Findings (most severe first)

**1. (pkg/plugin/implementation/reqpreprocessor/reqpreprocessor.go) and the identical loop in (pkg/plugin/implementation/dediregistry/dediregistry.go)'s `extractContextNetworkID` fallback — `break` instead of `continue` on a bad type assertion silently disables the new guard.** 
The key-alias loop does `if !ok { log.Warnf(...); break }` when a key's value fails the `.(string)` assertion, instead of `continue`ing to try the other spelling. A payload like `{"context":{"network_id":12345,"networkId":"other-network.com/retail"}}` hits `"network_id"` first, fails the string check, and aborts the whole loop — `networkId` is never read, `networkID` stays `""`, and the new per-request guard silently no-ops as if `network_id` were absent. No Beckn schema in `schemas/` constrains `network_id`'s type, and in `config/local-dev.yaml` `validateSign` runs before `validateSchema` anyway, so nothing downstream would catch this first. A caller can bypass the very check this PR adds by sending a malformed `network_id` alongside a valid `networkId`.

**2. Duplicate extraction logic between `reqpreprocessor.go` and `dediregistry.go`.** 
The snake_case/camelCase alias-loop is hand-rolled twice with copy-pasted logic, instead of extending the existing alias-resolution pattern already in `pkg/model/model.go` (`ResolveCallerID`/`ResolveSubscriberID`, or the `contextKeys`/`ParseContextKey` mechanism used for `transaction_id`/`message_id`). Two places to keep in sync.

**3. `extractContextNetworkID`'s body-parsing fallback (dediregistry.go) and the `*model.StepContext`-preserving rewrap in `simplekeymanager.go`'s `LookupNPKeys` appear to solve an already-solved problem.** 
`reqpreprocessor` sets `ContextKeyNetworkID` on the request context before any handler or OTel span wraps it, and `context.Value()` always walks the parent chain regardless of wrapping depth — so the "primary path" in `extractContextNetworkID` already covers every real request. The ~20-line body-reparse fallback and the ~15-line StepContext-preserving span rewrap in `simplekeymanager` (plus dedicated fallback tests) appear to be dead code in production. Worth confirming there's genuinely no path where `reqpreprocessor` is skipped before stripping this, since it adds real complexity (a manual struct-copy that's easy to get subtly wrong if `StepContext` gains fields later).

**4. Unrelated gofmt-only realignment mixed into a security-fix diff.** *(minor conventions note)*
`model.go`'s `StepContext` fields, `simplekeymanager.go`'s `Config` fields, and ~15 struct literals in `dediregistry_test.go` get pure whitespace realignment unrelated to this change, adding review noise to a security-sensitive diff.

**Bottom line:** #1 should block merge — #1 because it's a bypass of the new check itself. #2–#4 are worth a follow-up but not blocking.

@nirmalnr

nirmalnr:
Thanks for the review! 1 & 2 has been addressed.

Regarding 3: The body-parsing fallback in extractContextNetworkID and the *model.StepContext-preserving OTel span rewrap in simplekeymanager.LookupNPKeys are NOT dead code. reqpreprocessor middleware is currently optional (omitempty in config); without it, the primary ctx.Value() path returns "" and the fallback body-parse is the only coverage.

I think the complexity arises from this plugin and some other legacy plugins not being defined using StepContext while they are called using StepContext. I think this refactoring/cleanup can be a seperate PR.

nirmay:
Verified the latest commits — both fixes look correct:

- **`network_id` extraction bypass** ([[99aabcf2](https://github.com/beckn/beckn-onix/commit/99aabcf28d033d742babb50441465e4b9f438788)](https://github.com/beckn/beckn-onix/commit/99aabcf28d033d742babb50441465e4b9f438788)): `break` → `continue` fix confirmed, with a test covering the non-string-`network_id`-falls-through-to-`networkId` case. 
- **Duplicate extraction logic** ([[82e97949](https://github.com/beckn/beckn-onix/commit/82e97949d6bfaf7e053dd517a631985dbb1b9594)](https://github.com/beckn/beckn-onix/commit/82e97949d6bfaf7e053dd517a631985dbb1b9594)): consolidated into `model.ResolveNetworkID`, consistent with `ResolveCallerID`/`ResolveSubscriberID`. 

On the `dediregistry`-only scope: agreed this is fine to leave as-is — the plain `registry` plugin is for generation 1 networks, so extending the same check there isn't worth the effort. Noting it here for the record: until `registry` is retired, adapters configured with `id: registry` (e.g. the default `onix-bap`/`onix-bpp` configs) don't get the `context.network_id` membership check this PR adds — only `dediregistry`-backed deployments do.

The `StepContext`-preservation rewrap in `simplekeymanager` and the body-parsing fallback in `extractContextNetworkID` are justified — confirmed `reqpreprocessor` is optional per adapter config, so that fallback is the only enforcement path when it's omitted. Agreed the StepContext/context.Context typing inconsistency across plugins is worth its own cleanup PR.

Approving — nice work tightening this up.

nirmay (approved):
approved.

PR #854, opened 2026-07-06
fix(schemav2validator): remove core-schema skip in extended schema validation

### LINKED ISSUE #843: Tightening core schema check in schemav2validator plugin.
This happened during one of our workshops. There was a json packet being sent and it had some errors. So the ONIX validator was preventing it from being sent. The participant (by mistake I guess), changed the @context value to a non-existant path and it worked. Ideally ONIX should have encountered a 404 when it tried to fetch the non-existant path and should have prevented the json payload from being validated. The reason why ONIX did not do this was that the @context that was being passed had a "/core/schema" string somewhere in the path. Extended validation in ONIX checks for this path and considers it as core schema and does not validate the rest of the object. 

https://github.com/beckn/beckn-onix/blob/main/pkg/plugin/implementation/schemav2validator/extended_schema.go#L102C1-L104C2

This seems like an unintended effect. If the isCoreSchema method cannot be tightened, then we should clearly specify to schema authors that they should not have the string "core/schema" anywhere in their path. I am not sure if that is the right solution either as even if the schema authors follow it, it will be difficult to control NPs using it as skip value.

nirmay:
## Summary
- Removes `isCoreSchema` and its call site in `findReferencedObjects` — every object with `@context` + `@type` now always goes through Extended Schema validation instead of being conditionally skipped.
- `isCoreSchema` was a raw `strings.Contains(contextURL, "/schema/core/")` check on sender-controlled input, with no host or path verification — a sender could bypass Extended Schema validation for any object entirely just by including that substring in a bogus `@context`.
- The check no longer corresponds to anything in the current, actually-loaded core spec: `protocol-specifications-v2` (`core-v2.0.0-lts`, pinned via `beckndefaults`) only ever uses `@context`/`@type` on the JSON-LD `Attributes` extension point, which by the spec's own description is always domain-specific ("The @context property MUST reference a valid JSON-LD context URI ... identify the domain type being expressed"). The check was modeled on an older `protocol-specifications-new` draft repo layout (`schema/core/...`) that no longer exists — confirmed the fixture URL used in the old test now 404s.
- Updated `extended_schema_test.go` (removed `TestIsCoreSchema`, fixed two `TestFindReferencedObjects` cases that assumed skipping) and `README.md` (removed the documented "Filter Core Schemas" step) to match.

Full analysis posted on the issue: #843

Fixes #843

## Test plan
- [x] `go vet ./pkg/plugin/implementation/schemav2validator/...`
- [x] `go test ./pkg/plugin/implementation/schemav2validator/...` — all pass
- [x] Reviewer sanity-check: confirm no production network relies on the old core-context-skip behavior

### DIFF
--- a/pkg/plugin/implementation/schemav2validator/README.md
+++ b/pkg/plugin/implementation/schemav2validator/README.md
@@ -99,8 +99,7 @@ A startup warning is logged if no schemas are found, which usually means the dir
 
 **Extended Schema Validation** (if `extendedSchema_enabled: "true"` AND core validation passed):
 5. **Scan for @context**: Recursively traverse `message` field for objects with `@context` and `@type`
-6. **Filter Core Schemas**: Skip objects with `/schema/core/` in `@context` URL
-7. **Validate Each Domain Object**:
+6. **Validate Each Domain Object**:
    - Check domain whitelist (if `allowedDomains` configured)
    - Transform `@context` URL: `context.jsonld` → `attributes.yaml`
    - Load schema from URL/file (check cache first, download if miss)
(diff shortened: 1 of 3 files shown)


nirmalnr (approved):
The changes look good and have been tested locally. Approved.

PR #860, opened 2026-07-13
fix(dediregistry): exempt network-agnostic subscribers (e.g. catalog) from network membership checks

### LINKED ISSUE #859: dediregistry: context.network_id membership check (#839) breaks cross-network shared services like catalog
### Background
#839 added validation of `context.network_id` (from the inbound Beckn message) against the
counterparty subscriber's `network_memberships` returned by DeDi, in addition to the existing
`allowedNetworkIDs` config check. This is correct for regular NP-to-NP traffic, where both
participants genuinely belong to a specific network.

### Problem
Some subscribers — e.g. a catalog service — are meant to be common infrastructure shared across
multiple networks, not a member of any single one. Today, `network_memberships` is populated
only by explicit action from a network owner (there is no self-declared or wildcard membership
mechanism in DeDi). When such a shared service is looked up as the counterparty, and the request
carries a `context.network_id` for a network the service hasn't been explicitly enrolled in,
`validateContextNetworkID` in `pkg/plugin/implementation/dediregistry/dediregistry.go` rejects
the lookup — breaking legitimate traffic to/from these shared services.

### Where this shows up
`DeDiRegistryClient.validateMemberships` / `validateContextNetworkID`
(pkg/plugin/implementation/dediregistry/dediregistry.go, added in #839).

### Open question
How should the adapter/registry model represent a subscriber that is legitimately valid across
multiple (or all) networks, without requiring it to be explicitly enrolled as a member of every
individual network? No fix is proposed here — needs design discussion.

cc @nirmalnr

nirmalnr:
## Summary
Fixes #859. 

`validateMemberships`/`validateContextNetworkID` in `dediregistry.go` rejected lookups for shared cross-network services (e.g. the catalog) because `network_memberships` only reflects explicit per-network enrollment.
Adds a hardcoded `networkAgnosticSubscriberIDs` set (`staging.catalg.fabric.nfh.global`, `fabric.nfh.global`) and short-circuits `validateMemberships` for these subscriber IDs, skipping both the static `allowedNetworkIDs` guard and the per-request `context.network_id` check on the cache-hit and HTTP lookup paths.

## Test plan
- [x] `go test ./pkg/plugin/implementation/dediregistry/...` passes
- [x] Added `TestLookup` case: catalog subscriber bypasses a mismatched `allowedNetworkIDs` on the HTTP path
- [x] Added `TestDeDiRegistryClient_Lookup_Cache` case: catalog subscriber bypasses a mismatched `context.network_id` on the cache-hit path
- [x] Added `TestIsNetworkAgnosticSubscriber` unit test for the new helper
- [x] Added `TestContextNetworkIDValidation_NetworkAgnosticSubscriberBypass` covering disjoint memberships/allowlist/context.network_id
- [x] Test with staging catalog service details

### DIFF
--- a/pkg/plugin/implementation/dediregistry/dediregistry.go
+++ b/pkg/plugin/implementation/dediregistry/dediregistry.go
@@ -25,6 +25,21 @@ const defaultCacheTTL = 5 * time.Minute
 // to search across all cached registries in Beckn One. This value must not be configured externally.
 const dediAllRegistriesWildcard = "subscribers.beckn.one"
 
+// networkAgnosticSubscriberIDs identifies the common catalog service. These subscriber
+// IDs are treated as belonging to every network, since the catalog service is shared
+// across all networks by design — network-membership validation is skipped entirely
+// for lookups of these subscribers. Must not be configured externally.
+var networkAgnosticSubscriberIDs = map[string]struct{}{
+	"staging.catalg.fabric.nfh.global": {},
+	"fabric.nfh.global":                {},
+}
+
+// isNetworkAgnosticSubscriber reports whether subscriberID identifies the common catalog service.
+func isNetworkAgnosticSubscriber(subscriberID string) bool {
+	_, ok := networkAgnosticSubscriberIDs[subscriberID]
+	return ok
+}
+
 // Config holds configuration parameters for the DeDi registry client.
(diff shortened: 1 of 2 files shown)


nirmay (approved):
approved.

Issue #872, opened 2026-07-16
feat: cryptographic build attestation for adapter version identity


nirmay:
## Problem

#808 adds self-reported build identity (version, commit, tree-state, build-date) to the adapter's startup log, `/health` response, and OTel resource attributes (`onix.build.*`). Those fields solve honest-operator drift detection — spotting a stale, mismatched, or locally-patched deployment — but they don't stop a deliberately malicious network participant from faking them: anyone who controls their own build can hardcode any commit hash or tree-state they like via the same `-ldflags` mechanism.

Since beckn-onix is open source and deployed by independent operators across a Beckn network, network observability and any compliance/audit tooling need a way to tell "this operator is running a verified, unmodified release" from "this operator is self-reporting a version string that may or may not be true."

## Expected Behavior

A network observer (or the network's observability backend) should be able to cryptographically verify that a running adapter instance matches a known-good, official build — not just read a self-asserted string.

## Open Design Questions

- **Who signs the build?** e.g. an official CI pipeline signs release artifacts; self-built/dev binaries would be unsigned or flagged as such.
- **Where does the trust root live?** e.g. a public key set published by the network registry / a Beckn-wide authority, vs. per-network trust configuration.
- **Where is verification performed?** At OTel ingestion into the observability backend, at network registry subscription/handshake time, or both.
- **What's actually signed?** Likely the same identity bundle from #808 (commit, tree-state, build date) plus perhaps a binary/artifact hash — the signature should cover exactly the fields #808 already emits, not replace them.
- **Rollout path:** unsigned builds need to remain functional (this can't be a hard requirement for every deployment on day one) — likely an additive trust signal rather than a gate, at least initially.

## Relationship to #808

This is a deliberate scope split agreed on while implementing #808: honest-operator identity now, adversarial attestation as a separate, larger design effort. See #808 for the fields this would attest to.

PR #873, opened 2026-07-16
feat: log adapter version at startup and expose it at runtime (#808)

### LINKED ISSUE #808: feat: log adapter version at startup and expose it at runtime
## Problem

ONIX does not log the adapter version during startup, and there is no runtime mechanism to determine which version is currently running.

This makes it difficult to:
- Confirm which release is deployed in an environment
- Correlate logs and incidents with a specific version
- Debug environment-specific issues where the deployed version may differ from what is expected

## Current Behavior

The startup logs only include config path, logger initialisation, plugin manager initialisation, and HTTP server initialisation. No version information is emitted anywhere during boot or accessible at runtime.

## Expected Behavior

The running adapter version should be identifiable without having to inspect the binary, git history, or deployment manifests.

nirmay:
## Summary
- Adds a `pkg/version` package (`Version`, `GitCommit`, `GitTreeState`, `BuildDate`) populated via `-ldflags -X` at build time in `install/setup.sh`, both Dockerfiles, and the CI workflow (falls back to `dev`/`unknown` for unflagged builds).
- Logs build identity at startup, adds a `build` object to `/health`, and adds `onix.build.*` OTel resource attributes (plus defaulting `otelsetup`'s `service.version` to the build-time version) — documented in `OBSERVABILITY.md`.
- Scope is deliberately limited to self-reported build identity for spotting stale/mismatched/locally-patched deployments. Cryptographic build attestation (so a malicious operator can't fake these fields) is scoped separately as a follow-up: #872.

## Test plan
- [x] `go build` / `go test` pass for all touched packages (`pkg/version`, `core/module/handler`, `pkg/plugin/implementation/otelsetup`, `cmd/adapter`)
- [x] Manually built the adapter binary with test `-ldflags` values and confirmed the startup log line reflects them correctly
- [x] Reviewer sign-off on `/health` response shape and OTel attribute naming before merge

Closes #808

### DIFF
--- a/.github/workflows/deploy-to-gke-BS.yml
+++ b/.github/workflows/deploy-to-gke-BS.yml
@@ -12,6 +12,12 @@ jobs:
     steps:
     - name: Checkout Code
       uses: actions/checkout@v3
+      with:
+        # Full history + tags, so `git describe --tags` in version-vars.sh
+        # can actually resolve a release tag instead of always falling back
+        # to the bare commit SHA (the default shallow checkout fetches no
+        # tag refs at all).
+        fetch-depth: 0
 
     - name: Authenticate to Google Cloud
       uses: google-github-actions/auth@v2
@@ -33,12 +39,9 @@ jobs:
     - name: Build Docker Image
       run: |
         IMAGE_NAME=${{ secrets.GCP_REGION }}-docker.pkg.dev/${{ secrets.GCP_PROJECT }}/${{ secrets.GCP_REPO }}/beckn-onix:${{ github.sha }}
-        GIT_VERSION="$(git describe --tags --always 2>/dev/null || echo dev)"
-        GIT_COMMIT="$(git rev-parse --short HEAD)"
-        GIT_TREE_STATE="clean"
-        BUILD_DATE="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+        source install/scripts/version-vars.sh
         docker build -f Dockerfile.adapter \
-          --build-arg ONIX_VERSION="$GIT_VERSION" \
+          --build-arg ONIX_VERSION="$ONIX_VERSION" \
           --build-arg GIT_COMMIT="$GIT_COMMIT" \
           --build-arg GIT_TREE_STATE="$GIT_TREE_STATE" \
           --build-arg BUILD_DATE="$BUILD_DATE" \
--- a/.github/workflows/deploy-to-gke-BS.yml
+++ b/.github/workflows/deploy-to-gke-BS.yml
@@ -33,7 +33,16 @@ jobs:
     - name: Build Docker Image
       run: |
         IMAGE_NAME=${{ secrets.GCP_REGION }}-docker.pkg.dev/${{ secrets.GCP_PROJECT }}/${{ secrets.GCP_REPO }}/beckn-onix:${{ github.sha }}
-        docker build -f Dockerfile.adapter -t $IMAGE_NAME .
+        GIT_VERSION="$(git describe --tags --always 2>/dev/null || echo dev)"
+        GIT_COMMIT="$(git rev-parse --short HEAD)"
+        GIT_TREE_STATE="clean"
+        BUILD_DATE="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+        docker build -f Dockerfile.adapter \
+          --build-arg ONIX_VERSION="$GIT_VERSION" \
+          --build-arg GIT_COMMIT="$GIT_COMMIT" \
+          --build-arg GIT_TREE_STATE="$GIT_TREE_STATE" \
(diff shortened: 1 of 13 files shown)


nirmalnr:
**Minor doc/comment inconsistency** — pkg/version/version.go:366 and OBSERVABILITY.md's onix.build.tree_state row both describe only "clean"/"dirty" as possible values, but the actual zero-value default (unflagged builds) is "unknown" — same as the other three fields. Worth a one-line update so the docs match the real value space.

```
{"level":"info","time":"2026-07-17T11:55:02+05:30","message":"Starting Beckn-ONIX version=dev commit=unknown tree=unknown built=unknown"}
```

nirmalnr:
**Test coverage gap** — `otelsetup_test.go` wasn't updated to assert on the new behavior: no test checks `DefaultConfig().ServiceVersion == version.Version`, and no test verifies the three new `onix.build.*` resource attributes actually get attached to the OTel `Resource`. Existing tests (e.g. `TestSetup_New_DefaultValues`) only check that a provider is non-nil, not the resource attribute contents. Given this is the primary runtime surface for the new build-identity feature, a targeted assertion here would be worthwhile.

nirmalnr:
**Minor nit: Possibly-unmaintained workflow left untouched** — `build-and-deploy-plugins.yml` builds plugins directly (`go build -buildmode=plugin -buildvcs=false ...`) without going through `install/build-plugins.sh`, so plugins built via this path (including `otelsetup`) won't carry the new build-identity ldflags. I'm not sure whether this workflow is still actively used — it's manual-dispatch-only, targets a `beckn-onix-v1.0-develop`-style branch naming scheme that looks older than `main`, and has a hardcoded branch name in its clone step that doesn't even respect its own `target_branch` input, which suggests it may be stale/unmaintained. If it's still in active use, it should probably also source `version-vars.sh` for consistency; if it's dead, it might be worth removing rather than leaving it silently out of sync.

nirmalnr:
**Unable to test using `install/setup.sh`**: Step 4's `SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"` (line 135) runs *after* the `cd ..` on line 85, so with the intended invocation (`cd install && ./setup.sh`), it resolves to the repo root instead of `install/`. `source "${SCRIPT_DIR}/scripts/version-vars.sh"` then fails silently (no error check), `ONIX_LDFLAGS` stays empty, and `go build -ldflags "${ONIX_LDFLAGS}"` embeds nothing — so running `setup.sh` still produces `version=dev commit=unknown tree=unknown built=unknown`. Fix: compute `SCRIPT_DIR` once at the top of the script, before any `cd`.

I was only able to test this by running `version-vars.sh` 
```
source install/scripts/version-vars.sh
go build -ldflags "${ONIX_LDFLAGS}" -o beckn-adapter cmd/adapter/main.go
```

nirmay:
Thanks for the thorough review — addressed in 290a400:

**1. Doc/comment inconsistency (`GitTreeState` possible values)** — Fixed. Updated the comment in `pkg/version/version.go` and the `onix.build.tree_state` row in `OBSERVABILITY.md` to mention `unknown` as the zero-value default alongside `clean`/`dirty`.

**2. Test coverage gap** — Fixed. Extracted the resource-attribute construction into `buildBaseAttrs()` (the OTel SDK's `MeterProvider`/`TracerProvider` don't expose a getter for their `Resource`, so this was the direct way to make it assertable) and added two tests: one confirming `DefaultConfig().ServiceVersion` follows `pkg/version.Version`, and one confirming `buildBaseAttrs()` includes the three `onix.build.*` attributes with the expected values.

**3. `build-and-deploy-plugins.yml` not going through the new ldflags path** — Confirmed this workflow is dead, not just possibly stale: unmodified for 14 months, its 5 most recent runs (over a year ago) all failed, no runs since, and its `target_branch` input isn't even wired to what actually gets built (it clones a hardcoded `beckn-onix-v1.0-develop` branch regardless). Leaving it out of scope here since it doesn't touch `install/build-plugins.sh` or anything else this PR changes — happy to open a separate cleanup PR to remove it if others agree it's unused, but didn't want to bundle an unrelated deletion into this diff.

**4. Unable to test via `install/setup.sh`** — Confirmed and fixed, thank you for catching this — good catch. `SCRIPT_DIR` was computed inside Step 4, which runs *after* the script's own `cd ..`, so with the documented invocation (`cd install && ./setup.sh`) it resolved to the repo root instead of `install/`, and `source .../version-vars.sh` failed silently (no error check on that line). Moved `SCRIPT_DIR` to the top of the script before any `cd`, and added an error check so a future path issue fails loudly instead of silently defaulting to `dev`/`unknown`. Verified by simulating the exact `cd install && ./setup.sh` invocation.

nirmalnr (approved):

PR #893, opened 2026-07-29
catalogpublisher: decentralized-catalog publisher plugin + catalog/publish handler

nirmay:
## Summary
- New `catalogpublisher` core plugin (`pkg/plugin/implementation/catalogpublisher`) implementing `definition.CatalogPublisher`: given a publisher's catalog submissions, produces a signed manifest and a catalog index matching the Decentralized Catalog file spec, with incremental diffing (baseline vs. change file vs. no-op) against caller-supplied prior state.
- New `catalogPublish` HTTP handler (`core/module/handler/catalogPublishHandler.go`), a DS-internal, unsigned trigger mirroring `catalogPullHandler`'s conventions -- no `validateSign`/`addRoute`/`signAck` pipeline, since the caller is the operator's own tooling.
- Shared `localstore` package for reading/writing the manifest, catalog index, and catalog files to a local output root, used by both the handler and a demo CLI (`cmd/catalogpublisherctl`).
- Request body matches beckn.yaml's real `CatalogPublishAction` envelope (`context`/`message.catalogs[]`/`message.publishDirectives[]`) rather than a synthesized wrapper, so schema validation runs against the real request.
- Optional, reused-as-is plugin wiring for the handler: `schemaValidator` (`schemav2validator`) and `policyChecker` (`opapolicychecker`) validate/police the raw request; `manifestLoader` (`manifestloader`) checks read-only whether this node's manifest already declares its catalog index, staging a proposed update locally (never writing to the real manifest) and warning if not.
- `subscriberId` is derived automatically from `keyManager.config.subscriberId` (no duplicate declaration); a distinct `manifestSubscriberId` config covers DeDi's `namespace/registry/recordName` identity format for the manifest self-lookup, kept separate since it's a different shape than the plain Beckn `subscriberId` used for signing.
- `config/local-beckn-one-bpp.yaml` now also wires a `catalogPublish` module (it previously only existed in `local-beckn-one-bap.yaml`), matching how `starter-kit`'s `generic-bpp.yaml` wires it -- `schemaValidator`/`manifestLoader` active by default, not just commented examples.

## Fixes from self-review
- A catalog submission (or change-file upsert list) containing two items with the same `id` no longer produces a duplicated upsert entry in the emitted baseline/change file or when reapplying one (`catalogpublisher.go`'s `itemsByIDOrdered`, `catalogfile.go`'s `applyDiffBlock`); last write wins, cleanly.
- `NewCatalogPublishHandler` now loads `manifestLoader` before `policyChecker` and threads it through, so a manifest-backed policy type can actually resolve under this handler instead of always failing with "requires ManifestLoader plugin to be configured".
- A `PublishError.Fatal` per-catalog error now flips the handler's overall response `status` to `FAILED` instead of being silently reported identically to a routine rejection.
- Retired-catalog tombstone ordering in the published index is now deterministic (iterates the request's `retire` list instead of a Go map), and a catalogId that's both submitted and retired in the same call -- where Publish's own rule is "submission wins" -- no longer gets a spurious extra "retired" result in the HTTP response alongside its real outcome.
- `diffCatalogs`/`diffArrayField`/`diffCatalogAttributes` now parse each submission's prior/next catalog JSON once instead of 3-6 times per submission.
- Removed dead commented-out manifest-write code in `localstore.Write`; fixed a stale README cross-reference that only mentioned the bap config example, not the newer bpp one.

## Test plan
- [x] `go build ./...` and `go test ./...` clean.
- [x] Exercised end-to-end against the `starter-kit` generic devkit (see companion PR beckn/starter-kit#17): publish, incremental change file, retire, schema-validation rejection, manifest-link warning.
- [x] Added regression tests for the duplicate-id dedup fix, the manifestLoader/policyChecker threading fix, the Fatal-status wiring, and the submitted-and-retired-same-id response fix.
- [ ] Reviewer sanity check of the plugin's signing/diffing logic and the handler's error/response shape.


nirmalnr:
`localstore.Write` is not atomic/ordered safely (localstore.go:150)
The index file is written first, unconditionally, before any per-catalog content file. If a later `os.WriteFile` for a baseline/change file fails (disk full, permission error, process killed), the index on disk already references a baseline/changes entry — complete with digest and signature — for a file that was never written. Any later localstore.Load for that catalog will fail reading it (reconstructState), and any external crawler fetching that URL gets a 404 despite the (correctly signed) index claiming it exists.

PR #894, opened 2026-07-29
Phase-1 decentralized catalog crawler (onix plugin + module)

ameersohel45:
## Summary

Phase-1 **decentralized catalog crawler** — a framework-agnostic module (`pkg/crawler`) plus a thin onix plugin and a DS-internal `/crawl` handler — built on top of the initial `catalog-crawler` scaffold.

This branch builds on `catalog-crawler`'s last 4 commits and reworks/extends them into the full Phase-1 implementation:

- `167eac3` refactor(catalogcrawler): rewrite to match the file spec
- `81e8981` fix(dedi): real detached-JWS sign/verify per §7
- `c38080a` fix(catalogcrawler): match manifest catalog-index entry on the registry
- `36f8f2b` catalog-crawler: add crawler plugin and DS-facing catalog/pull handler

## Behaviour change vs `catalog-crawler` (the base)

The base branch implemented a **synchronous, pull-based** crawler:

- **`Crawler.CrawlSubscriber(CrawlRequest) → CrawlResult`** — one call fetched + verified a subscriber's manifest/index and **composed the full catalog** (`CatalogResult.Catalog` = the baseline with every change file applied).
- A DS-internal **`catalog/pull` handler** invoked it **synchronously** and **returned the catalogs inline** in the HTTP response — a `CatalogPullCallbackAction`-shaped body `{status: COMPLETED|FAILED, catalogs: […], error}` ("no ACK/callback split; a single synchronous response suffices").
- Change detection was **digest-based** (`CrawlModeIncremental` skipped a part whose digest was unchanged); scope was one subscriber per request.

This branch reworks that into a **background, push-based** crawler:

- The crawler is a **daemon**: `Start` / `Stop` + an on-demand `CrawlNow(indexURL) → runId`. Scheduled index/catalog jobs run continuously from plugin init.
- The endpoint is now an **async trigger**: `POST /crawl {indexUrl}` returns **`202 ACCEPTED` + a `runId`** — it does **not** return the catalog inline. The crawler **pushes** each resolved catalog to Discovery's **`/catalog/push`** (→ Kafka → publish job → Elasticsearch); outcomes are observed via logs/metrics and the `crawler_*` tables, not the trigger's response.
- Change detection is **version-driven** (a persisted per-catalog cursor); the digest is kept purely as a fetch-time **integrity** check.
- State is durable: a Postgres work queue (claim/lease + retry/park), MERGE-only delta pushes, and per-index scheduling.

(The `/crawl` handler file is named `catalogPullHandler.go` after a file-only rename; its struct and route are still `crawl`.)

## What's in it

**Crawl model**
- Postgres work queue (`crawler_index` / `crawler_queue` / `crawler_catalog`) with `FOR UPDATE SKIP LOCKED`, a claim token + lease; the cursor advances only on a successful push.
- **Version-driven** change detection; the per-file digest is verified as an integrity gate **before** decode.
- **MERGE-only delta push** (`CRAWLER_MERGE_ONLY`): the delta is built straight from the change file's catalog metadata envelope — no baseline fetch. First sync / re-baseline still fetches the baseline. A change file that omits the envelope is a permanent content fault (no silent baseline fallback).
- Pushes to Discovery `/catalog/push` with `publishDirectives` (`catalogType`, `updateMode`, `visibleTo`).

**Fetch / decode**
- SSRF-guarded fetch with **dial-time IP validation** (DNS-rebind safe), artifact size caps, and a decompression-bomb guard.
- Conditional-GET (ETag / If-Modified-Since → 304 skip).
- `json` + `json.gzip` via an extensible codec registry.

**Reliability / lifecycle**
- Typed fault taxonomy driving park-vs-retry: digest mismatch / SSRF / oversize / content-invalid **park** (ERROR, operator signal); 5xx and 408/425/429 **retry** with capped backoff.
- §6b lifecycle vocabulary (SyncOutcome / SyncPhase / IndexOutcome / DaemonState).

**Observability**
- Component/stage/message structured logging with `run_id` / `pass_id`; `CRAWLER_LOG_LEVEL`.
- Ops metrics: `crawler_sync_outcome_total{outcome,fault}`, a `seconds_since_last_success` liveness gauge, `catalogs_parked` / `catalogs_tracked`, `sync_lag`, `index_poll`, `queue_depth`, and push/index latency histograms.

**Scheduler**
- An on-demand `/crawl`'d index **joins the recurring schedule**: the scheduled pass re-polls every persisted index (deduped, each gated by its own `next_crawl_at`), so a later change file is picked up with no second `/crawl`.

**Structure**
- Domain packages: `pkg/crawler/{catalog,fetch,decode,publish,source,store,runner,config}` + a composition root; full `catalogcrawler` → `crawler` rename across module / cmd / plugin. `/crawl` handler file renamed `crawlHandler.go` → `catalogPullHandler.go` (file-only; symbols/route unchanged).

## Testing
- TDD throughout. `go test ./pkg/crawler/...` green (incl. DB-gated store/engine tests via `CRAWLER_TEST_DB_DSN`), plus `gofmt` / `go build` / `go vet`.

## Deferred to Phase 2
- Signature-tuple verification, per-catalog scope enforcement, removals / FULL mode, and a concrete registry source client.


nirmay:
Thanks for the writeup and the work here — the crawl engine's core mechanics (queue claim/lease, SSRF dial-time guard, decompression cap) are solid. I don't think this is mergeable in its current shape, though — there are two equally-weighted sets of blockers: structure/placement, and correctness/security. Both need to be addressed in the same pass, not deferred.

## Structure / placement

**1. `cmd/crawler` shouldn't be a new top-level binary.**
The repo's only existing top-level binaries are `cmd/adapter`, `cmd/catalogpublisherctl`, `cmd/test`. This PR adds `cmd/crawler/main.go` as a second, independent daemon entrypoint that duplicates what `pkg/plugin/implementation/crawler/crawler.go` already does as a plugin. Every other plugin runs only through the plugin/handler path — none has a standalone `cmd`. If a standalone worker mode is genuinely needed, it should be a mode on the existing adapter, not a second binary with its own lifecycle.

**2. Hard-coded Postgres coupling.**
`pkg/crawler/store/open.go` calls `sql.Open("pgx", dsn)` directly, and the plugin requires `CRAWLER_DB_DSN` to start at all — there's no storage interface a different backend could satisfy. This makes the plugin unusable for anyone not running Postgres. Please refactor so `pkg/crawler` depends on a `Store` interface, with Postgres as one pluggable implementation of it, consistent with how other stateful plugins are structured.

**3. `docs/` is a brand-new, inconsistent top-level convention.**
The base branch has no `docs/` directory. This PR introduces one with 5 crawler-only files, several of which read as internal review notes (`catalog-crawler-phase1-review.md`, `catalog-crawler-gzip-support-review.md`, `catalog-crawler-restructure-cr.md`) rather than shipped documentation. Every other plugin documents itself via its own `README.md` under `pkg/plugin/implementation/<name>/`. Please fold the real docs into the plugin README and drop the review-note files from the diff entirely.

**4. `pkg/crawler` shouldn't be a new top-level package tree.**
Every plugin-specific implementation currently lives entirely under `pkg/plugin/implementation/<name>/`. This PR instead creates `pkg/crawler/{catalog,fetch,decode,publish,source,store,runner,config}` as a sibling to genuinely cross-cutting packages like `pkg/security`, `pkg/model`, `pkg/telemetry`. Please fold this back under the plugin's own directory (e.g. `pkg/plugin/implementation/catalogcrawler/internal/...`). If there's a real "framework-agnostic second driver" need behind this, that's worth an explicit discussion — it doesn't justify a stray top-level package *and* a stray top-level binary landing together.

## Correctness / security

**5. SSRF pre-check has no timeout.**
`checkPublicURL`'s `net.LookupIP` call runs before the request is built and has no context bound to it — unlike the dial-time guard, which correctly uses a context-bound resolver. A slow or malicious resolver for an attacker-supplied publisher host can stall this check indefinitely, bypassing `FetchTimeout` entirely.

**6. Retry cap doesn't actually park.**
`scheduleRetry` is documented as "give up after `MaxAttempts` failed pushes," but the code never parks on exceeding it for transient faults — it logs and keeps retrying forever with capped backoff. A permanently-failing catalog never surfaces as an operator-actionable state.

**7. Digest verification fails open.**
A `FileEntry` with a blank/missing `digest` is treated as verified, so `FetchFile` returns unverified bytes. Since signature verification is deferred to Phase 2, digest is the *only* integrity gate in this phase — and it's optional as written. A malicious or MITM'd index that simply omits `digest` on a change file gets its content pushed to Discovery as authentic.

**8. No per-index lock between on-demand and scheduled crawls.**
`CrawlNow` (the `/crawl` trigger) and the scheduled ticker can run concurrently against the same index with no lock, risking duplicate queue enqueues and lost-update races on the version cursor. Untested.

**9. Signature verification silently dropped, not just an unused arg.**
The prior `catalogcrawler` verified per-file Ed25519 signature tuples and the manifest's detached-JWS proof before trusting fetched content. This engine only checks a SHA-256 digest and defers signing to "Phase 2" — that's a real security capability regression, not a constructor-arg cleanup, and should be called out plainly rather than folded into a scoping note. the branch that was created earlier was already using the signer plugin and include verification artifacts that should have been used instead.

**10. `/crawl` has no source allowlist.**
Combined with #5, an unauthenticated internal endpoint that accepts any `indexUrl` and triggers a server-side fetch is close to an unauthenticated SSRF primitive. There's no check that `indexUrl` is one of the configured `CRAWLER_INDEX_URLS`.

Please restructure per 1–4 and fix 5–10 in the same pass — happy to re-review once it's reshaped.

nirmay:
Thanks @ameersohel45 
This PR overall looks good, we have some more changes in the spec to be implemented. I will come back on the plan for the same.

ameersohel45:
## `/crawl` payload change — now registry-based (commit `f9a7037`)

The on-demand `POST /crawl` trigger no longer takes a raw index URL. It now takes a **registry URL + network(s)** and runs the same DeDi `/query` discovery the scheduled pass uses, so on-demand and scheduled crawls share one input model.

### Request

**Before**
```json
{ "indexUrl": "https://publisher.example/catalog-index.json" }
```

**After**
```json
{
  "registryUrl": "https://fabric.nfh.global/registry/dedi",
  "networkId": "beckn.one/testnet"
}
```

| Field | Required | Notes |
|-------|----------|-------|
| `registryUrl` | ✅ | Absolute `http`/`https`. DeDi registry base; the crawler calls `GET {registryUrl}/query/{networkId}` to discover the provider catalog indexes, then fetches + verifies + pushes each. |
| `networkId` | ✅\* | Single network id, e.g. `beckn.one/testnet`. |
| `networkIds` | — | List form, e.g. `["beckn.one/testnet","beckn.one/mainnet"]`. Merged with `networkId`, de-duplicated, order preserved. |

\* at least one of `networkId` / `networkIds` must be present.

### Response — `202 Accepted`
```json
{
  "status": "ACCEPTED",
  "registryUrl": "https://fabric.nfh.global/registry/dedi",
  "networkIds": ["beckn.one/testnet"],
  "runId": "…"
}
```

Validation: a missing/blank `registryUrl`, a non-`http(s)` scheme, or no network → `400 Bad Request`. The crawl runs asynchronously; progress surfaces in the crawler's telemetry, correlated by `runId`.

### Scheduled pass (unchanged input model)
The background pass reads the same source from config — `CRAWLER_REGISTRY_URL` + `CRAWLER_NETWORK_IDS` — and discovers via `/query` every `CRAWLER_INDEX_INTERVAL`. Setting a registry base with no networks falls back to the static `CRAWLER_INDEX_URLS` list.

nirmay:
@Atri-sarika This is the crawler work in onix, pls do share your thoughts,

Issue #870, opened 2026-07-14
encrypter/decrypter/publisher/vcvalidator: classify BadReqErr usage into correct codes


nirmay:
Part of #861. Blocked by #862 (core `ErrorCode` type).

## Scope
`encrypter`, `decrypter`, `publisher`, and `vcvalidator` all currently use the same blanket `model.NewBadReqErr`/`NewSignValidationErr` pattern as the other plugins audited in this EPIC, defaulting to `"Bad Request"`/`"Unauthorized"` regardless of actual cause.

Classify each plugin's failure modes into the correct `ErrorCode` value (likely a mix of `SCH_*`, `AUT_*`, `NET_*`, `BIZ_*` depending on the specific failure — determine per plugin during implementation).

## Scope includes
- Unit test updates across all four plugins covering each newly-distinguished failure path, maintaining ≥90% coverage.

## Acceptance criteria
- No remaining blanket `BadReqErr`/`SignValidationErr` usage in these four plugins where a more specific canonical code is determinable from the underlying failure.

nirmay:
Merged into `main` via #887 (epic-level PR for #861, includes post-review fixes).

PR #896, opened 2026-08-01
refactor(model): collapse coded error types into a single CodedErr

### LINKED ISSUE #888: Collapse BadReqErr/SignValidationErr/NotFoundErr into one generic coded-error type with explicit HTTP status
## Background

Follow-up from the #861 epic-level review (PR #887). `BadReqErr`, `SignValidationErr`, and `NotFoundErr` all embed the same `codedErr` (Code storage, `Unwrap()`, `resolveCode` default fallback) and, as of #887, share a `codedErr.becknError(prefix, defaultCode)` helper for building their `*Error` payload. The only thing left distinguishing them as separate Go types is the HTTP status `nackBecknError` (core/module/handler/responsestep.go) picks per type — 400/401/404 — via a type-switch dispatch.

Each type also currently has two constructors that only differ in whether `Code` is set (`NewBadReqErr`/`NewCodedBadReqErr`, and the equivalent pairs for the other two types), even though `resolveCode` already treats an empty `Code` as "use the default" — so the coded/uncoded split per type is redundant too.

## Proposal

1. Collapse each type's two constructors into one, e.g. `NewBadReqErr(code string, err error) *BadReqErr`, called with `code=""` for the existing lazy-default behavior.
2. Go further: collapse `BadReqErr`/`SignValidationErr`/`NotFoundErr` into a single generic coded-error type that takes an explicit HTTP status at construction (not derived from the taxonomy code prefix — `NET_*` alone spans 404/500/502/503 depending on the specific code, so status can't be inferred from prefix alone). This would let any plugin (e.g. vcvalidator, which currently has to pick between `NewCodedBadReqErr` and `NewCodedSignValidationErr` by convention) call one constructor regardless of failure "flavor."
3. `SchemaValidationErr` (multi-cause aggregation across several `Error`s) and `AckNoCallbackErr` (ACK/NACK status, protocol-version-gated 202/500) should stay as separate, distinctly-shaped types — they differ in structure, not just HTTP status, so folding them in would lose real behavior.

## Scope

This touches `pkg/model/error.go`'s type hierarchy and `nackBecknError`'s dispatch switch, plus every call site of the affected constructors — at least 58 non-test call sites across 16 files (signvalidator, opapolicychecker, encrypter, publisher, vcvalidator, schemav2validator, simplekeymanager, registry, decrypter, keymanager, schemavalidator, router, plus core/module/handler). Larger than a single-PR change; do this as its own ticket, not folded into #861.

areebahmeddd:
## Summary

* replaces `BadReqErr`, `SignValidationErr`, and `NotFoundErr` with a single `model.CodedErr`, reducing duplicated error types while preserving existing response behavior
* introduces a more flexible error construction API with explicit status codes, making it easier for plugins to return different HTTP statuses without relying on concrete error types
* keeps wire compatibility unchanged, updates all existing call sites, and extends tests to verify both error codes and HTTP status mappings

Fixes #888


areebahmeddd:
**Deliberately skipped**

* didn't migrate the 59 call sites to `NewCodedErr` since the helper constructors preserve existing message prefixes and changing them would alter NACK text on the wire
* didn't touch `SchemaValidationErr` or `AckNoCallbackErr` since they differ in structure, not just status
* didn't change `vcvalidator` returning 401 for network failures since that's covered by #870 #884
left a note that it's now a one-line change there
* didn't change the `"Signature Validation Error: "` prefix for non-signature causes for the same reason and the original deferral still applies

**One behavior difference**

* the old dispatch selected by type priority (`SignValidationErr` > `BadReqErr` > `NotFoundErr`) regardless of wrapping depth. the new dispatch uses the outermost `CodedErr`
* this only matters if an error chain contains multiple `CodedErr`s. traced every path and none do today, so the behavior is currently unreachable

**Loose thread**

* `NewNotFoundErr` has no production callers, only tests. it may be dead code

nirmay:
One test-coverage gap (not a functional bug): a few pre-existing test suites (keymanager_test.go, simplekeymanager_test.go) still assert only errors.As(err, &model.CodedErr{}) + .Code for sign-validation failures, without also asserting .HTTPStatus() == 401 the way signvalidator_test.go and error_test.go now do. Since all three error flavors collapsed into one type, a future regression that accidentally builds these via the wrong constructor (right code, wrong status) would slip past those two suites. Worth a follow-up, not a blocker.

No correctness, breaking-change, or design issues beyond that — this is a clean, well-scoped execution of an already-agreed refactor.
@areebahmeddd thank you!

areebahmeddd:
sure 🙂

nirmay (approved):
approved.

PR #897, opened 2026-08-13
fix(reqmapper): NACK on JSONata transform failures instead of silent fallback

### LINKED ISSUE #880: reqmapper: surface and classify JSONata transform failures (SCH_*/BIZ_*)
Split out of #867 during scoping — same reasoning as the #866 carve-out from #861: this is new error-handling infrastructure and a real behavior change, not a relabel of an already-coded generic error, so it doesn't fit #861's "remap existing generic code to the new taxonomy" pattern.

## Background

`reqmapper`'s JSONata transform call (`reqmapper.go:99-103`) currently swallows every evaluation failure:

```go
mappedBody, err := s.engine.Transform(ctx, parsed.action, parsed.req, s.role)
if err != nil {
    log.Errorf(ctx, err, "Transformation failed for action %s", parsed.action)
    return body, nil   // original, untransformed body is returned with no error
}
```

A failing transform produces no NACK today — it's logged and falls back to passing the original, untransformed body through. Before this can be classified onto any `Error.code`, the swallow-to-fallback behavior has to become swallow-to-error (or at least error in some cases) — a behavior change beyond what #867's remaining scope covers.

**Scope note found during #881's self-review**: `MappingEngine.Transform` actually has two distinct uncoded failure sites feeding that same swallow-and-fallback in `transformBody`, not one — `json.Marshal(req)` (`reqmapper.go:258`, "failed to marshal request for mapping") in addition to `expr.Evaluate(...)` (`reqmapper.go:264`, "JSONata evaluation failed"). The marshal failure is a different failure mode (a marshalling problem with the parsed request map, not a JSONata-semantic failure) and isn't a JSONata error at all, so it won't carry a `*v206.JSONataError` code — it should get its own classification (likely a `BIZ_GENERIC_ERROR`-style fallback, since it reflects an internal/unexpected condition rather than a shape mismatch) rather than being folded into the `T*`/`D*` JSONata-code bucket below. Make sure whichever fix lands here covers both sites, not just the JSONata-evaluate one.

## What's already known (from investigating #867)

- The JSONata engine actually used (`github.com/jsonata-go/jsonata` → resolves to the `v206` engine via `OpenLatest()`) returns typed, coded errors: `*v206.JSONataError{Code string, ...}`, with codes falling into clear prefix buckets (catalog at `v206/jsonata.go:4514-4611`):
  - `S*` — parser/syntax errors, compile-time only (unreachable per-request given `loadBuiltinMappings` compiles without recovery mode).
  - `T*`/`D*` — runtime type-mismatch / dynamic errors, genuinely data/shape-driven → good fit for `SCH_SCHEMA_ADAPTATION_FAILED`.
  - `U*` — recursion-depth exceeded, timeout, internal panic recovery — resource/internal, not a shape signal → best candidate for `BIZ_GENERIC_ERROR` fallback (arguably NET_*/POL_*-flavored instead, but that's outside SCH_*/BIZ_*'s scope here — worth a second look when this is picked up).
  - The top-level `github.com/jsonata-go/jsonata` wrapper package reqmapper currently imports does **not** re-expose this type; reaching it requires also importing `github.com/jsonata-go/jsonata/v206` directly and using `errors.As`.
- `SCH_INVALID_JSONPATH` (named in #867's original description) has **no real signal today**: JSONata treats a missing/undefined output path as a *successful* result, not an error, and `Transform()` does zero post-transform validation of the output shape. Giving this code real meaning (as opposed to an arbitrary relabel of the same `T*`/`D*` errors) requires adding new post-transform validation — e.g. checking that required output keys are present/non-null against the target mapping's expectations.
- Existing test coverage for the main `reqmapper` package is 75.8% — meeting the epic's ≥90% bar for whichever paths this ticket touches will need substantive new tests regardless of how the classification design shakes out, including new testdata mappings that can genuinely trigger `T*`/`D*` JSONata errors (not just synthetic mock expressions).

## Scope (tentative — needs design discussion when picked up)

1. Decide whether transform failures should NACK at all (behavior change) or remain a soft-fallback for some/all failure kinds.
2. If NACKing: import `v206` directly, classify by `JSONataError.Code` prefix per the buckets above.
3. Also classify the `json.Marshal(req)` failure in `Transform()` (`reqmapper.go:258`) — a non-JSONata failure sharing the same swallow-and-fallback today; don't let it fall through the gap between this ticket and #867.
4. Decide whether `SCH_INVALID_JSONPATH` is worth the new post-transform validation work, or whether `SCH_SCHEMA_ADAPTATION_FAILED` alone is a sufficient bucket for all JSONata evaluation failures.
5. New tests covering each newly-distinguished path, maintaining ≥90% coverage on touched code.

Not part of any EPIC for now — standalone, like #866.

areebahmeddd:
`reqmapper` used to swallow every JSONata transform failure and silently pass the original, untransformed body through with no NACK. This classifies the failure and NACKs instead.

* Evaluate failures are classified by their JSONata error code: shape/type mismatches (`T*` / `D*`) → `SCH_SCHEMA_ADAPTATION_FAILED`, resource/internal errors (`U*`, timeouts, panics) → `BIZ_GENERIC_ERROR`
* The pre-evaluation marshal failure gets its own `BIZ_GENERIC_ERROR` classification
* Added a mapping fixture that triggers a genuine JSONata error, so the test covers a real failure, not just a mock

Closes #880


nirmay (approved):
approved.

PR #898, opened 2026-08-13
fix(schemavalidator): skip validation on empty GET/DELETE body

### LINKED ISSUE #736: fix(schemavalidator): crash on empty body for GET/DELETE requests
## Summary

`schemavalidator.Validate` (the v1 file-based validator) calls `json.Unmarshal(data, &payloadData)` unconditionally. GET and DELETE requests carry no body, so `data` is `[]byte{}`. This produces:

```
failed to parse JSON payload: unexpected end of JSON input
```

The pipeline aborts with a 400 NACK before routing or any downstream step runs.

## Affected file

`pkg/plugin/implementation/schemavalidator/schemavalidator.go` — `Validate()`, lines 57–69

```go
func (v *schemaValidator) Validate(ctx context.Context, url *url.URL, data []byte) error {
    var payloadData payload
    err := json.Unmarshal(data, &payloadData)   // ← fails on empty body
    if err != nil {
        return model.NewBadReqErr(fmt.Errorf("failed to parse JSON payload: %v", err))
    }
    if payloadData.Context.Domain == "" {
        return model.NewBadReqErr(fmt.Errorf("missing field Domain in context"))
    }
    if payloadData.Context.Version == "" {
        return model.NewBadReqErr(fmt.Errorf("missing field Version in context"))
    }
```

## Root cause

Same as `schemav2validator`: the validator was designed for POST-only Beckn flows where every request carries a JSON body. The v1 validator additionally requires `context.domain` and `context.version` from the body, both of which are absent in a GET/DELETE request.

## Expected behaviour

- When `len(data) == 0`, skip JSON unmarshal and body-based context extraction.
- Derive domain and version from query parameters (e.g. `?domain=...&version=...`) or from the URL path, consistent with how Beckn 2.0 GET/DELETE requests encode their context.
- Skip body schema validation for bodyless requests; validate query parameters if a schema exists for them.

## Fix outline

1. Guard `json.Unmarshal` with `if len(data) > 0`.
2. For empty-body requests, extract domain/version from `url.Query()`.
3. Construct the schema lookup key from those values and the endpoint, then validate query params against it (or skip validation if no parameter schema is defined).

## Related

- Counterpart to [#735](https://github.com/beckn/beckn-onix/issues/735) which covers `schemav2validator`.
- Part of the broader GET/DELETE support gap for Beckn 2.0 catalog/subscription verbs.

deepusnath:
Fixes beckn/beckn-onix#736.

## Root cause

validateSchemaStep.Run (core/module/handler/step.go) only rejects an empty body for POST requests before calling SchemaValidator.Validate. A GET/DELETE request with no body falls through to Validate() regardless, which unconditionally ran json.Unmarshal on the body -- so every legitimate bodyless GET/DELETE request failed with "failed to parse JSON payload: unexpected end of JSON input" instead of succeeding.

The sibling schemav2validator already handles this correctly via its OpenAPI-derived bodylessActions index. schemavalidator (v1) has no such per-endpoint index -- it validates against JSON Schema files keyed by domain/version extracted from the payload itself, so there is no way for it to know per-endpoint whether a body is required.

## Fix

Treat an empty body as nothing-to-validate and return nil, relying on the step layer existing guarantee that only non-POST requests ever reach Validate() with an empty body.

## Testing

- Added TestValidator_Validate_EmptyBody covering both nil and []byte{}.
- go test passes (one pre-existing, unrelated failure noted).
- gofmt -l and go vet clean


pkg/plugin/implementation/schemavalidator/schemavalidator.go
```diff
@@ -56,6 +56,16 @@ func New(ctx context.Context, config *Config) (*schemaValidator, func() error, e
 // reqURL.Path holds the Beckn endpoint action already stripped of the module
 // base path by the step layer (e.g. "search") — no leading slash.
 func (v *schemaValidator) Validate(ctx context.Context, reqURL *url.URL, data []byte) error {
+	// An empty body only ever reaches here for GET/DELETE requests — the step
```
areebahmeddd: Small wording point: the comment says an empty body only reaches here for GET or DELETE, but step.go checks specifically for POST, so PUT/PATCH would also pass here if empty. Nothing sends those today, but could we change the comment to "only non-POST requests reach here"?

pkg/plugin/implementation/schemavalidator/schemavalidator.go
```diff
@@ -56,6 +56,16 @@ func New(ctx context.Context, config *Config) (*schemaValidator, func() error, e
...
+	// layer (validateSchemaStep) already rejects an empty-body POST before
+	// calling Validate. This validator has no per-endpoint notion of "requires
+	// a body" (unlike schemav2validator's OpenAPI-derived bodylessActions
+	// index), so it skips content validation instead of failing every bodyless
+	// request with a spurious "failed to parse JSON payload" error.
+	if len(data) == 0 {
+		return nil
```
areebahmeddd, line 68: i think we could add a debug log before returning? The v2 validator does this 🤔

pkg/plugin/implementation/schemavalidator/schemavalidator_test.go
```diff
@@ -193,6 +193,27 @@ func TestValidator_Validate_SchemaErrorDetails(t *testing.T) {
 	}
 }
 
+// TestValidator_Validate_EmptyBody confirms that an empty body — the shape a
+// GET/DELETE request arrives with, since validateSchemaStep only rejects an
+// empty body for POST before calling Validate — is treated as "nothing to
+// validate" rather than a JSON parse failure.
+func TestValidator_Validate_EmptyBody(t *testing.T) {
```
areebahmeddd, line 200: Test looks good. One thought: the actual bug depends on how step.go decides what to let through, and there's already a test there for the empty POST case. A matching one for a bodyless GET would cover the full path, so if someone changes that check later we'd catch it.

deepusnath: @areebahmeddd Thanks for the review. Pushed a follow-up commit addressing all three points:

- Reworded the comment to say "non-POST requests" instead of "GET/DELETE requests", since PUT/PATCH with an empty body also falls through the step layer's POST-only check.
- Added a debug log before the early return, matching the pattern schemav2validator uses for its bodyless-request branch.
- Added TestValidateSchemaStep_Run_EmptyBodyGet_CallsValidator in step.go's test file, covering that a bodyless GET reaches the validator (complementing the existing empty-body-POST rejection test).

Let me know if you'd like anything adjusted further.

pkg/plugin/implementation/schemavalidator/schemavalidator.go
deepusnath: Fixed — reworded to 'non-POST requests' since PUT/PATCH can hit this too.

nirmay (approved):
approved.

PR #901, opened 2026-08-17
feat(catalogstore): CatalogBlobStore plugin + shared index assembler

### LINKED ISSUE #900: feat(catalogstore): define CatalogStateStore plugin + localcatalogstore reference impl
## Background

`catalog-publisher` (branch `catalog-publisher`, not yet merged to `main`) currently has `core/module/handler/catalogPublishHandler.go` call `localstore.Load`/`localstore.Write` directly — concrete filesystem-backed package functions, not a plugin interface. `localstore`'s own doc comment already says this out loud: "this package is one concrete, filesystem-backed way to supply and persist [prior] state, not part of the core plugin's own logic."

`catalogpublisher.Publish` itself is already a pure function (`PublishRequest` in, `PublishResult` out, no I/O) — the file-spec-critical diff/version/sign logic that must stay identical for every publisher. Storage/prior-state-reconstruction is the one piece a real deployment (S3, a database, a CMS-backed catalog) will always want to own differently, and today that means forking `catalogpublisher`'s surrounding wiring rather than swapping one component.

## Scope of this ticket

Build **only** the new capability plugin, independent of the `catalog-publisher` branch:

1. `definition.CatalogStateStore` (naming TBD, propose this or `CatalogArtifactStore`) — a plugin interface, and `CatalogStateStoreProvider` following the existing `KeyManagerProvider`/`SchemaValidatorProvider` convention (`New(ctx, config map[string]string) (CatalogStateStore, func() error, error)`).
   - `LoadPriorState(ctx, catalogIDs []string) (map[string]definition.PriorCatalogState, carryForward []json.RawMessage, error)`
   - `WriteResult(ctx, definition.PublishResult) error`
   - Exact method shapes to be finalized against `definition.PriorCatalogState`/`PublishResult` as they exist on the `catalog-publisher` branch today.
2. `localcatalogstore` — the reference/default implementation, functionally equivalent to today's `localstore` package (filesystem layout: `.well-known/`, `index/`, `catalogs/`, `catalogs/changes/`), but built against the new interface from the start and wired into `manager.go`'s plugin-loading like the other capability plugins.
3. Keep both plugin-system-agnostic where possible: a caller who doesn't want the onix plugin/manager machinery should be able to construct `localcatalogstore` directly and call it as a plain Go type — same as `localstore` already permits today.
4. Unit tests for `localcatalogstore` (port/adapt the existing `localstore_test.go` coverage).

## Explicitly out of scope (follow-up work, not this ticket)

- Wiring `catalogPublishHandler.go` / `catalogpublisherctl` on the `catalog-publisher` branch to use the new plugin instead of `localstore` directly, and deleting the old `localstore` package.
- Any change to `catalogpublisher.go`'s core diff/sign/version logic.
- `KeyManager` interface changes — explicitly not doing this (it's a read-only use today, but it may need to write keys later, so no narrowing).
- Signing/schema-validation structure — already reviewed as correctly structured, no changes needed there.

## Why

Lets a developer swap the storage/prior-state backend per deployment without touching or forking the spec-critical publish core, following the same plugin pattern already used for `KeyManager`/`Registry`/`SchemaValidator`/`PolicyChecker` in this codebase. Landing it as its own plugin first (before touching `catalog-publisher`) also shrinks that branch's footprint once it's rebased to consume this instead of its own bespoke `localstore`.

nirmay:
## Summary

Closes #900.

- `CatalogBlobStore`/`CatalogBlobStoreProvider` (`pkg/plugin/definition/catalogblobstore.go`) — a new, minimal onix plugin surface: read/write bytes at a path, no catalog-file vocabulary at all. This is the actual swap point for a real deployment's storage backend (local disk today; S3/GCS/git/an authenticated CDN root can each drop in later as their own implementation).
- `localcatalogblobstore` — the first (filesystem) implementation, usable directly as plain Go (`New(root)`) or loaded via the plugin manager.
- `catalogstore.Store` (`pkg/plugin/implementation/catalogstore/`) — the one shared, backend-agnostic assembler: understands how a catalog index, its baselines, change files, and "latest" pointers fit together per the decentralized-catalog file spec, built on top of any `CatalogBlobStore`. Deliberately **not** a plugin itself — there is exactly one correct way to do this per spec, so only the blob backend underneath varies. `LoadCatalogs` reconstructs a catalog's current content to diff against; `Publish` merges new/changed signed entries into the existing index and persists everything.
- `pkg/catalogfile` — the change-file-apply helper both `catalogstore` and (later) a rebased `catalog-publisher` branch need; extracted standalone with no plugin dependencies.

## Scope

This is intentionally independent of the (unmerged) `catalog-publisher` branch — see #900 for the full design discussion (why the assembler and the blob-store plugin are two separate interfaces, and why compaction/gzip/publish-latest config stays on the publisher core rather than leaking into `catalogstore`). Wiring `catalog-publisher`'s handler/CLI to consume this instead of its own bespoke `localstore` is deliberately a separate, later piece of work.

Marked draft pending that follow-up discussion.

## Test plan
- [x] `go build ./...` and `go vet ./...` clean
- [x] `go test ./pkg/catalogfile/... ./pkg/plugin/implementation/localcatalogblobstore/... ./pkg/plugin/implementation/catalogstore/...` — all pass
- [ ] Wire into `catalog-publisher` branch (follow-up, not this PR)


nirmay:
Closing in favor of #903 — feat/catalogstore-plugin is already fully merged into feat/catalog-publisher-wired's history, and with the core catalog/store/publisher logic now living in catalog-core, this branch's remaining scope (CatalogBlobStore + localcatalogblobstore) is small enough to review as part of #903 instead of as a separate PR.

PR #903, opened 2026-08-19
feat(catalog): wire catalog-publisher onto pkg/catalog/store + pkg/catalog/publisher

nirmay:
## Summary
- Wires the onix `catalogpublisher` plugin, the `catalogPublish` HTTP handler, and `cmd/catalogpublisherctl` onto the shared `pkg/catalog/store` (assembly/persistence) and `pkg/catalog/publisher` (diff/sign/version/compaction orchestration) libraries built on `feat/catalogstore-plugin`.
- `pkg/plugin/implementation/catalogpublisher` is now a thin wrapper: resolves the signing keyset via `KeyManager`, converts between `definition.*` and the library's own types, delegates everything else.
- Adds debug/info/warn/error logging throughout the library layers via an explicit `*slog.Logger` parameter, bridged into onix's own `pkg/log` (zerolog) so it shows up at whatever level onix is configured for.
- **Latest commit**: repoints `pkg/catalog`, `pkg/catalog/store`, `pkg/catalog/publisher`, and `pkg/security/artifactsigner`/`artifactverifier` at the new `github.com/beckn/catalog-core` module (see companion PR on `feat/catalogstore-plugin` for the extraction) instead of local copies. Currently wired via a local `replace` directive in `go.mod` until `catalog-core` is tagged/released. Also fixes a latent bug where `definition.ErrBlobNotFound` and `catalog-core`'s own not-found sentinel were different values -- now aliased to the same one.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./...` all pass
- [x] `gofmt -l .` clean on touched files
- [x] End-to-end smoke tested via `starter-kit`'s `catalog-publisher-e2e` branch: baseline publish, change-file publish, compaction, gzip, `latest` pointer, and retirement all verified through the fully delegated stack
- [ ] Needs a `catalog-core` release/tag before the `replace` directive can be dropped


nirmalnr:
Can be taken up as follow up issues as well if valid. I believe **1** is worth fixing at least as a follow up issue.

**1. No concurrency control around the index read-merge-write cycle (Medium-High)**
`catalogPublishHandler.ServeHTTP` ([[catalogPublishHandler.go:415-449](https://claude.ai/epitaxy/core/module/handler/catalogPublishHandler.go:415)](core/module/handler/catalogPublishHandler.go:415)) does `catalogStore.LoadCatalogs` → `publisher.Publish` → `catalogStore.Publish` with no lock, and `store.Store.Publish` itself (catalog-core) does a plain read-modify-write: `readIndexEntries` → merge in memory → `blobs.Put(IndexPath(), ...)`, with no mutex anywhere in that path. `http.Handler.ServeHTTP` is invoked concurrently by `net/http` for simultaneous requests, and there's a single shared `*store.Store` instance per handler. Two overlapping `catalogPublish` calls (e.g. an operator script publishing several catalogs in parallel against the same `outputRoot`) can both read the same index snapshot, both write their own merged copy back, and the second write silently discards the first call's index-entry update — its underlying baseline/change files stay on disk but become unreferenced. This isn't mentioned anywhere in the README's design notes. Since fixing it "correctly" means either a mutex in the handler serializing calls to the same `catalogStore`/`outputRoot`, or locking inside catalog-core's `Store` itself, worth deciding where that belongs before this ships — at minimum the handler-side wiring in this PR should serialize its own calls.

**2. All error responses return HTTP 200 (Low-Medium, likely intentional but worth confirming)**
Every failure path in `ServeHTTP` — including `LoadCatalogs` I/O errors, `Publish` errors, and `catalogStore.Publish` write failures — goes through `writePublishJSON`, which never calls `w.WriteHeader`, so the response is always `200 OK` with `status: "FAILED"` in the body. The doc comment on `ServeHTTP` explains this is intentional and matches `catalogPullHandler`'s convention, but that convention was designed around per-catalog business rejections, not infra-level failures (disk write errors, etc.) — worth double-checking that whatever calls this endpoint (operator tooling, CI, monitoring) inspects the body rather than the status code, since a backend outage would otherwise look identical to success at the transport level.

nirmalnr (approved):

PR #906, opened 2026-08-24
fix(catalogpublish): add referential/business validation missing from schemav2validator

### LINKED ISSUE #905: catalogPublish: add referential/business validation missing from schemav2validator
## Context

`schemaValidator` (schemav2validator), when configured against beckn.yaml's `CatalogPublishAction`, only validates per-object shape/enum values. It cannot express cross-references between `catalogs[]` and `publishDirectives[]`/`resourceDirectives[]`, and `schemaTypes` isn't even a field in the real `CatalogPublishAction` schema, so there's currently no way to validate it at all.

See discussion in [PR #903](https://github.com/beckn/beckn-onix/pull/903) review follow-up for the full breakdown.

## Scope

1. **`publishDirectives[].catalogId` must reference a submitted catalog.** Reject (or report as a non-fatal `PublishError`) any directive whose `catalogId` doesn't match a `catalogs[].id` in the same request -- today it's silently unused.
2. **No duplicate `catalogId`** within `catalogs[]` or `publishDirectives[]` in a single request -- a duplicate directive currently silently overwrites the earlier one in the handler's `map[catalogID]publishDirective`.
3. **`resourceDirectives[].resourceId` must reference a resource actually present in that catalog.** `resourceDirectives` isn't wired into `CatalogSubmission` at all yet -- this needs that wiring plus the reference check.
4. **`updateMode` should be enforced or explicitly rejected as unsupported.** It's schema-validated (`FULL`/`MERGE` enum) but has zero effect today -- silently accepting and ignoring a value is worse than rejecting it until implemented.
6. **`schemaTypes` validation via a custom schema sourced from RFC NFH-014** (decentralized-catalog file spec) in the `protocol-specifications-v2` repo -- `schemaTypes` isn't a `publishDirectives` field in the canonical `beckn.yaml`, so it needs its own schema definition from the NFH-014 spec rather than beckn.yaml's `CatalogPublishAction`.

Explicitly out of scope for this ticket:
- Content-based `catalogType` inference when `publishDirectives` is omitted (spec-documented default behavior, needs its own design discussion).
- Flagging/rejecting `CatalogPublishAction`'s `deprecated: true` status at config time.

## Where this lives

- `core/module/handler/catalogPublishHandler.go` (`ServeHTTP`, `publishDirective`/`publishRequest` types)
- `pkg/plugin/implementation/catalogpublisher/storeadapter.go` (`CatalogSubmission` conversion, for `resourceDirectives` wiring)
- Possibly `catalog-core`'s `pkg/catalog/publisher` if `resourceDirectives` validation belongs at the library level rather than the handler

nirmay:
## Summary

Closes #905.

`schemaValidator` (`schemav2validator`) validates per-object shape/enum values against beckn.yaml's `CatalogPublishAction`, but can't express cross-references between arrays in the same document, and `schemaTypes` isn't even a field in the canonical schema. This adds a `validatePublishRequest` layer in `catalogPublishHandler.go` that runs **unconditionally**, independent of whether `schemaValidator` is configured:

1. `publishDirectives[].catalogId` must reference a submitted catalog -- a dangling directive is now a 400, not silently ignored.
2. No duplicate `catalogId` within `catalogs[]` or `publishDirectives[]`.
3. `resourceDirectives[].resourceId` must reference a resource actually present in that catalog, and `extends.masterResourceId` must be present -- `resourceDirectives` is now parsed at all (it wasn't wired into the request type before).
4. `updateMode: "FULL"` is explicitly rejected as unsupported (only `"MERGE"` behavior exists); any other invalid value is rejected too.
6. `schemaTypes` (an NFH-014 addition to the catalog index entry, not a beckn.yaml `CatalogPublishAction` field) is now wired end to end -- parsed from `publishDirectives`, validated against a custom JSON Schema derived from NFH-014's own documented shape, and flows through to `CatalogSubmission.SchemaTypes` and the published index entry.

Items 5 (content-based `catalogType` inference) and 7 (flagging `CatalogPublishAction`'s deprecated status) are explicitly out of scope -- see #905.

Also updates `catalogpublisher/README.md`: fixes two claims left stale by this change (`updateMode`/`resourceDirectives` described as unvalidated; `schemaTypes` described as unwired), documents the new validation layer, and now recommends configuring `schemav2validator` for any catalog publishing (linking its README) since `validatePublishRequest` alone doesn't cover field-level shape/enum constraints.

## Self-review

Ran a self-review pass before opening this PR; found and fixed:
- `resourceDirectives[].extends.masterResourceId` was parsed but never validated (only `resourceId` was) -- now rejected when missing.
- The custom `schemaTypes` JSON Schema was compiled via a package-var initializer that would panic (crashing the whole binary at load time) on any future compile failure -- now compiled lazily via `sync.Once`, surfacing a failure as an ordinary error instead.
- Left as-is: `validatePublishRequest` and the submissions-building loop each unmarshal every catalog's raw JSON independently (same bytes, same purpose) -- kept separate to keep `validatePublishRequest` a pure, independently-testable function rather than threading precomputed probes through it; the double-parse cost is minor for catalog-sized payloads.

## Test plan

- [x] 18 new tests (`validatePublishRequest` unit tests + `ServeHTTP` integration tests) covering all five validation rules, positive and negative cases
- [x] `go build ./...`, `go vet ./...`, `go test ./...` all pass across the whole repo
- [x] `gofmt -l` clean on touched files


nirmay:
Closing in favor of #911 — this PR's base (core/module/handler/catalogPublishHandler.go) was deleted entirely by #910's EndpointHandler generalization. #911 redoes this same validation work (#905) on top of that refactor, inside pkg/plugin/implementation/catalogpublisher/decode.go instead.

PR #909, opened 2026-08-24
feat(catalogcrawler): discover providers via dediregistry instead of direct DeDi endpoints

### LINKED ISSUE #908: catalogcrawler: use dediregistry plugin for discovery instead of direct DeDi endpoints
## Problem

The `catalogcrawler` plugin does not go through the `dediregistry` plugin for provider discovery. Instead it has its own standalone HTTP client (`pkg/plugin/implementation/catalogcrawler/internal/source/dediquery.go`) that calls DeDi's `/query/{networkId}` endpoint directly, with its own base-URL config (`dediRegistryUrl`), its own response-parsing structs, and its own "live" state filtering — duplicating logic that should live in `dediregistry`.

The code contains an explicit acknowledgment of this gap (`dediquery.go`, top-of-file comment): the client was "copied over from the catalog-crawler prototype as-is: the dediregistry plugin doesn't support the /query API yet ... so this is a standalone client for now rather than a reuse of that plugin."

`dediregistry` (`pkg/plugin/implementation/dediregistry/dediregistry.go`) currently only supports single-record lookups:
- `Lookup` — by subscriber_id/key_id
- `LookupRegistry` — by namespace/registryName
- `LookupNode` — by namespace/registry/recordName

None of these query DeDi's bulk `/query/{networkId}` endpoint, so there's no way today to list providers/subscribers for a given network through the plugin.

## Objective

1. Add networkId-based query support to `dediregistry` (new method on `RegistryMetadataLookup`, e.g. `QueryByNetwork(ctx, networkID string) ([]model.SubscriberRecord, error)`), backed by DeDi's `/query/{networkId}` endpoint and reusing existing meta-parsing helpers (`parseMetaFromData`, `extractURLObjectArray`) instead of duplicating them.
2. Update `catalogcrawler` to discover providers via the `dediregistry` plugin instance (through the new method) instead of its own direct HTTP client, and remove the now-redundant `dediRegistryUrl` config key and `dediquery.go`.
3. Update plugin wiring (`CrawlerProvider.New`, `Manager.Crawler` in `pkg/plugin/manager.go`) so the crawler receives a `RegistryMetadataLookup` instance, mirroring how `Manager.ManifestLoader` already does.

## Scope

- `pkg/plugin/definition/registry.go` — new method on `RegistryMetadataLookup` (or new interface)
- `pkg/plugin/implementation/dediregistry/dediregistry.go` — implementation + tests
- `pkg/plugin/implementation/catalogcrawler/internal/source/dediquery.go` — remove
- `pkg/plugin/implementation/catalogcrawler/internal/source/registry.go` — adapt `RegistryClient` to use dediregistry
- `pkg/plugin/implementation/catalogcrawler/catalogcrawler.go` — drop `dediRegistryUrl` config, wire in `RegistryMetadataLookup`
- `pkg/plugin/definition/catalogcrawler.go` — `CrawlerProvider.New` signature
- `pkg/plugin/manager.go` — `Manager.Crawler` signature + call site
- READMEs for both plugins

Breaking change: `CrawlerProvider.New` and `Manager.Crawler` signatures change (additional `RegistryMetadataLookup` param).

nirmay:
## Summary

- Adds `QueryByNetwork(ctx, networkID string) ([]model.SubscriberRecord, error)` to `RegistryMetadataLookup`, implemented in `dediregistry` against DeDi's `/query/{networkId}` bulk endpoint (reusing existing `parseSubscriptionFromData`/`parseMetaFromData` helpers instead of duplicating parsing logic).
- Replaces `catalogcrawler`'s standalone `DediQueryClient` (direct HTTP calls to DeDi, its own response structs, its own `dediRegistryUrl` config) with a small adapter (`registryadapter.go`) backed by the `dediregistry` plugin instance via `RegistryMetadataLookup`.
- Wires `RegistryMetadataLookup` through `CrawlerProvider.New` → `Manager.Crawler` → `cmd/adapter/main.go`'s `initCrawler`, mirroring the existing pattern used for `ManifestLoader`.

Fixes #908

## Breaking changes

- `definition.CrawlerProvider.New` signature: adds a `metadataLookup RegistryMetadataLookup` parameter.
- `Manager.Crawler` signature: adds the same parameter.
- `catalogcrawler` config: `dediRegistryUrl` key removed (no longer read) — discovery now uses the `dediregistry` plugin's own configured `url`.

## Test plan

- [x] `go build ./pkg/... ./cmd/... ./core/...`
- [x] `go vet ./pkg/... ./cmd/... ./core/...`
- [x] `go test ./pkg/... ./cmd/... ./core/...` — all pass, including new `TestQueryByNetwork` (dediregistry) and `TestMetadataLookupClient_*` (catalogcrawler adapter) covering live/non-live filtering, malformed/missing details, and error paths.
- [x] Updated existing mocks (`catalogPublishHandler_test.go`, `manifestloader` tests) that implement `RegistryMetadataLookup` to satisfy the new interface method.
- [x] Updated both plugins' READMEs to document `QueryByNetwork` and drop the stale config/caveat.


nirmalnr:
**Double-encoded catalog_index_urls silently dropped**

parseMetaFromData, now reused by QueryByNetwork for catalogcrawler discovery, drops double-encoded meta.catalog_index_urls (a JSON string containing the array) instead of parsing it, silently losing providers that the deleted DediQueryClient/parseCatalogIndexURLs explicitly handled.
A DeDi registry record double-encodes meta.catalog_index_urls as a JSON string (e.g. "[{\"url\":\"...\"}]") -- a real-world quirk the deleted DediQueryClient.parseCatalogIndexURLs was specifically written and tested (TestDediQueryClient_ParsesDoubleEncodedMeta, deleted with no replacement) to tolerate. parseMetaFromData's type switch only recognizes []any as an array; a string value lands in meta[key] (plain string map), never metaArrays[key]. registryDiscoverer.Discover in catalogcrawler.go only reads rec.MetaArrays["catalog_index_urls"], so that provider's catalog index URL is never emitted as an IndexRef and the provider's catalog is never crawled, even though the exact same record worked via the old direct-DeDi path. No test in the current dediregistry_test.go covers this shape.

Was this intentionally removed?

nirmalnr (approved):

PR #922, opened 2026-09-04
fix(catalogcrawler): don't abort entire poll tick on one bad network ID

### LINKED ISSUE #921: catalogcrawler: single unresolvable network ID aborts entire poll tick, blocking all other networks
## Component

`fidedocker/onix-adapter:1.9.0`, `pkg/plugin/implementation/catalogcrawler`

## Summary

When `plugins.crawler.config.networks` lists multiple network IDs, a single network ID that the DeDi registry can't resolve (404 on `GET /registry/dedi/query/<networkId>`) causes the entire poll tick to abort — not just that one network. Every other correctly-configured network in the list, even ones with providers successfully found in the same tick, gets skipped for that cycle.

## Evidence

With `networks` set to 8 IDs including one unregistered (`ion.id/ion-dc-registry`), the log showed 7 successful `registry lookup succeeded` lines (including real provider counts: `ion-launch=4`, `staging=1`, `staging-test=1`, `prod-test=1`, `ion-apex-sb-reg=1`) followed by the one 404, and then:

```
"message":"catalogcrawler: poll tick failed"
```

No indexes were polled and no catalogs pushed that cycle, despite 5 of 8 networks having live, resolvable providers.

## Impact

Any single misconfigured/unregistered network ID silently blocks catalog discovery for every other network sharing that crawler instance — a config typo or a not-yet-registered network (which is a normal, expected state during rollout) has an outsized blast radius.

## Contrast with old behavior

The prior crawler build (`type: crawl` handler, custom GCR image) logged each network/index result independently and still completed the cycle for everything else, even when one index was unreachable (observed with `ayushmatha.in`'s 404'd ngrok index — that one failure didn't block other catalogs).

## Workaround applied

Removed the unregistered network ID from the `networks` list until it's actually registered in DeDi. Not a real fix — just avoids the trigger condition.

## Suggested fix

The per-network registry lookup loop should catch and log a failed lookup for one network ID without returning an error that aborts the whole tick — continue polling the remaining resolvable networks and only surface the failed one as a warning.

nirmay:
## Summary
- `registryDiscoverer.Discover` (catalogcrawler's registry-backed network discovery) used to return on the very first `QueryByNetwork` error, discarding refs already found for other configured networks and failing the whole poll tick.
- A single unregistered/misconfigured network ID (e.g. a DeDi 404) now only skips that network — logged as a warning — while discovery continues for every other configured network.
- If **every** configured network fails, that's still treated as fatal: logged at error level and returned, so a real registry outage keeps surfacing loudly instead of silently discovering zero indexes every tick. The returned error wraps each individual per-network error (`errors.Join`) instead of just a count, so it's self-describing without correlating against earlier Warn logs.
- `multiSource.Discover` (unions the registry-backed source with any `staticIndexUrls` source) had the identical "one bad thing kills everything" shape one layer up — one source erroring discarded refs a sibling source already found. It now applies the same partial-failure tolerance: one source failing doesn't wipe out another source's results, and only fails when every source fails. When some (not all) sources fail, it logs at Warn (a single failing tick can be transient/self-recovering; a source failing repeatedly escalates via `registryDiscoverer`'s own per-network Warn→Error threshold) so a total registry outage sitting behind a healthy `staticIndexUrls` config isn't invisible.
- A network that keeps failing tick after tick (while other networks stay healthy) escalates from a Warn log to an Error log once it's failed `consecutiveFailEscalateThreshold` (3) ticks in a row, so a permanently broken/decommissioned network doesn't go unnoticed indefinitely. The streak resets the moment that network succeeds again. On-demand crawls (`CrawlRegistry`) intentionally don't share this streak state — each call gets its own fresh discoverer, documented at both call sites.
- The "tolerate partial failure, error only if everything failed" logic (used by both `registryDiscoverer` and `multiSource`) is a single shared `errIfAllFailed` helper, grouped with this file's other small stateless helpers at the bottom, instead of being hand-rolled twice or living inline ahead of its callers.

Fixes #921

## Test plan
- [x] `go test ./pkg/plugin/implementation/catalogcrawler/...` and `go test -race ./pkg/plugin/implementation/catalogcrawler/...` — full suite passes
- [x] `TestRegistryDiscoverer_SkipsFailingNetworkWithoutAbortingOthers` — one network fails, one succeeds → succeeding network's refs still returned, nil error
- [x] `TestRegistryDiscoverer_AllNetworksFailingReturnsError` — every configured network fails → error returned, refs nil
- [x] `TestRegistryDiscoverer_NoNetworksConfiguredIsNotAFailure` — zero configured networks isn't misclassified as "all failed"
- [x] `TestRegistryDiscoverer_TracksConsecutiveFailuresPerNetworkAcrossTicks` — a persistently failing network's streak accumulates across ticks, doesn't leak into an unrelated healthy network, and resets on recovery
- [x] `TestRegistryDiscoverer_EscalatesLogLevelAtThreshold` — captures actual log output to confirm Warn→Error escalation fires exactly at the threshold tick, not just the internal counter
- [x] `TestMultiSource_ToleratesOneSourceFailingWhenAnotherSucceeds` — one source fails, another succeeds → succeeding source's refs still returned, nil error
- [x] `TestMultiSource_AllSourcesFailingReturnsError` — every source fails → error returned, refs nil
- [x] `TestMultiSource_LogsOnPartialSourceFailure` — a partially-failed tick (still "successful") still emits a Warn-level log so it isn't invisible, and does not escalate to Error on a single tick
- [x] `go vet` / `gofmt -l` clean

## Self-review
Ran three rounds of an 8-angle review pass; converged in round 3 with zero medium/high findings remaining.

**Round 1** — fixed: `multiSource` aborting the whole tick on one source's error (same bug class as #921, one layer up); all-networks-failed error losing per-network detail (now `errors.Join`-wrapped); no escalation path for a permanently-broken single network (added consecutive-failure streak → Warn/Error); dead test fixture field removed.

**Round 2** — fixed: a total registry outage could be fully hidden when a `staticIndexUrls` source also succeeded (`multiSource` now logs on partial failure); mutex doc comment cited a concurrency scenario that can't happen given `CrawlRegistry`'s actual wiring (corrected + documented that on-demand crawls don't share failure-streak state); doubly-nested error message when both failure layers fire at once (fixed via shared `errIfAllFailed` + dropped redundant prefix); escalation test only checked a counter, not actual log output (added log-capturing tests); duplicated aggregation logic (extracted to `errIfAllFailed`).

**Round 3** — fixed: the partial-source-failure log fired at Error even for a single self-recovering tick, noisier than warranted (downgraded to Warn — repeated failure still escalates via the existing per-network threshold); `errIfAllFailed` was placed inline ahead of its callers instead of with this file's other small helpers at the bottom (moved). Remaining round-3 findings (three low-severity notes: `multiSource`'s all-failed path not self-logging the way `registryDiscoverer`'s does, the CrawlRegistry fresh-instance behavior being documented in three places, and a negligible unconditional `Sprintf` on the hot path) were assessed as pure polish with no functional impact and left as-is — round 3's Angle G explicitly confirmed no remaining instance of the original bug pattern anywhere in this diff's blast radius, including re-checking `ConfigSource` (flagged as unverified in round 1, confirmed clean).


nirmalnr (approved):
Approved

Issue #1, opened 2024-03-22
ONIX should not remove older installations


vishi24:
## Description

improve the existing Onix-Beckn installation script to ensure it retains previous installations and configurations during updates. The script should provide users with the option to either update the existing system or perform a fresh installation, based on their preference.

#### Approach:
1. **Script Assessment**:
   - Evaluate the current Onix-Beckn installation script to understand its functionalities and limitations.
   - Identify areas where the script removes older installations and configurations.

2. **Enhancement Strategy**:
   - Modify the script to detect existing installations and configurations.
   - Implement logic to prompt users for their preference - update or fresh installation.
   - Ensure that the script retains older configurations if the user chooses to update the system.
   - Develop mechanisms to handle version upgrades smoothly while preserving user configurations.

#### Key Considerations:
- **User Experience**:
  - Prioritize user experience by providing clear prompts and options during script execution.
- **Configuration Retention**:
  - Ensure that the script does not overwrite or remove existing configurations during updates.
- **Version Management**:
  - Implement robust version detection and upgrade mechanisms to facilitate seamless updates.

## Goals

- [ ] Script must retain the older configurations if available.

## Expected Outcome

- The Onix-Beckn installation script will be updated to provide users with the option to update or perform a fresh installation.
- The script will intelligently detect existing installations and configurations and retain them during updates.
- Update will not remove/effect the the existing installation unless the user requests for a fresh installation.

## Acceptance Criteria

- [ ] Script must run properly without getting into issue.
- [ ] The script must execute as mentioned above.

vbabuEM:
This is primarily a discovery task and we should get the final go ahead from Ravi. One other thing to keep in mind is that there should be atleast some way to install all the components. For example older script had an option to install sandbox and webhook sandbox for simulating the seller side software. It should still be possible to install it. 
Also, if we have multiple scripts (guided install vs individual component install), we should use the DRY(Dont Repeat Yourself) principle to structure the script.

faizmagic:
As discussed @vbabuEM to create a individual issue tickets with details. this ticket can be closed, as soon as the individual tickets are in place. please add the the ticket details here

vbabuEM:
Some of the issues already worked up that are related to this are:
https://github.com/beckn/beckn-onix/issues/52
https://github.com/beckn/beckn-onix/issues/34
https://github.com/beckn/beckn-onix/issues/55

vbabuEM:
Marking as duplicate to #2 and closing this

PR #923, opened 2026-09-06
refactor(reqmapper): migrate to translator plugin interface

### LINKED ISSUE #797: refactor(reqmapper): migrate to Translator plugin interface
## Background

The `reqmapper` plugin currently owns its JSONata execution internally. The `Translator` plugin interface introduced in #770 (Epic: schema version mediator) defines a clean, execution-engine-agnostic contract:

\`\`\`go
Translate(ctx context.Context, artifact []byte, payload []byte) ([]byte, error)
\`\`\`

`reqmapper` is architecturally analogous to `SchemaVersionMediator` — both are orchestrators that load a translation artifact from some source and apply it to a payload. The source of the artifact (operator config vs schema.beckn.io) is an outside concern; the execution is the same.

## Goal

Refactor `reqmapper` to depend on the `Translator` plugin interface rather than owning translation execution internally. This makes `reqmapper` translation-engine-agnostic — JSONata today, SHACL or agent-based tomorrow, with no changes to `reqmapper` itself.

## Tasks

- [ ] Inject `Translator` into `reqmapper` via `TranslatorProvider`
- [ ] Replace internal JSONata execution with `Translator.Translate` calls
- [ ] Validate that `jsonatatranslator` (from #770) is a drop-in replacement
- [ ] Update `reqmapper` config and docs accordingly

## Future benefit

Once this lands, `reqmapper` gains support for any future `Translator` implementation (SHACL, agent-based, etc.) purely through config — no code changes required.

## Dependencies

Depends on #770 (schema version mediator — `Translator` interface and `jsonatatranslator` implementation) being merged first.

areebahmeddd:
## Summary

* `reqmapper` now delegates JSONata expression compilation and evaluation to an injected `Translator`, using the same interface introduced by `SchemaVersionMediator` in #770
* adds `jsonatatranslator` as the default `Translator` implementation, so existing `payloadTransformer` configs continue to work unchanged when no `translator` is configured
* malformed JSONata expressions now fail on first use instead of adapter startup since compilation happens behind the `Translator` interface; this is documented in `CONFIG.md`

## Testing

`go build`, `go vet`, and `go test ./...` all pass. Also built both plugins as real `.so` files and ran a live end-to-end smoke test by dynamically loading and wiring them, matching the adapter's runtime behavior

Closes #797


nirmay:
Thanks @areebahmeddd This is an important refactoring.

## Review notes

Ran build/vet/test on the branch — all pass. A few things worth addressing before/after merge:

### 1. Duplicated JSONata expression cache (medium)
[`[jsonatatranslator.go](https://claude.ai/epitaxy/pkg/plugin/implementation/jsonatatranslator/jsonatatranslator.go)`](pkg/plugin/implementation/jsonatatranslator/jsonatatranslator.go) (L16-90) — the `exprCache`/`newExprCache`/`defaultMaxExprCacheEntries`/`compiledExpr` cache is a near-verbatim copy of the existing one in `schemaversionmediator.go` (~L329-352, ~L1029-1049), differing only in the cap (500 vs 200). Any future fix or tuning to the eviction policy now has to be applied in two places.

This also traces back to a scope gap: #770 originally scoped a standalone `JSONataTranslator` plugin (its own `.so`) as a task separate from `SchemaVersionMediator`, but when #770 landed, `schemaversionmediator.go` kept its JSONata compile/cache/execute logic inline rather than depending on a separate `Translator` plugin. So the standalone `jsonatatranslator` didn't actually exist before this PR — #797's assumption ("Validate that `jsonatatranslator` (from #770) is a drop-in replacement") doesn't hold; this PR is the first place it's built as its own artifact.

Given that, I'd like to block on this: the original intent behind #770/#797 was reqmapper and schemaversionmediator sharing one `Translator` implementation, and this PR is the natural place to finish that — retrofit `schemaversionmediator` onto `jsonatatranslator` and delete its inline copy as part of this PR's scope. If there's a reason to defer that (e.g. blast radius on schemaversionmediator, timeline), let's discuss, but by default I think it belongs here rather than as a follow-up.

### 2. `loadTranslator` reinvents the existing generic helper (low)
[`[stdHandler.go:541](https://claude.ai/epitaxy/core/module/handler/stdHandler.go#L541)`](core/module/handler/stdHandler.go#L541) — `loadTranslator` hand-rolls the same nil-check/skip-log/wrap-error/log-success shape that `LoadPlugin[T]` (L430) already provides, and that every other single-config slot in this file uses (Cache, SignValidator, SchemaValidator, Router, Publisher, Signer, TransportWrapper). Suggest `LoadPlugin(ctx, "Translator", translatorCfg, mgr.Translator)` instead, to match convention and drop a duplicate function.

### 3. Startup fail-fast → runtime-only failure for malformed mappings (informational, no action needed)
[`[reqmapper.go:205](https://claude.ai/epitaxy/pkg/plugin/implementation/reqmapper/reqmapper.go#L205)`](pkg/plugin/implementation/reqmapper/reqmapper.go#L205) — `loadBuiltinMappings` no longer compiles JSONata expressions at construction time, so a malformed mapping for a rarely-hit action now only errors on first live use instead of blocking startup. This is intentional and already documented in the PR body, `CONFIG.md`, and the new plugin's README — flagging just for visibility, not asking for a change.

### 4. Minor per-request allocation in cache lookup (low)
[`[jsonatatranslator.go:59](https://claude.ai/epitaxy/pkg/plugin/implementation/jsonatatranslator/jsonatatranslator.go#L59)`](pkg/plugin/implementation/jsonatatranslator/jsonatatranslator.go#L59) — `Translate` converts the artifact `[]byte` to a `string` on every call (even cache hits) before the lookup, copying the full expression text per request. Under high volume with large mapping expressions this is a small but avoidable allocation. Not blocking, just noting.

areebahmeddd:
Sure, addressed.

cc @nirmay

nirmay:
## Follow-up review notes

### `stdHandler.initPlugins` encodes a plugin-selection policy that belongs in config, not in wiring

```go
translatorCfg := cfg.Translator
if translatorCfg == nil && (cfg.PayloadTransformer != nil || cfg.SchemaVersionMediator != nil) {
    translatorCfg = &plugin.Config{ID: defaultTranslatorID}
    cfg.Translator = translatorCfg // keep PluginEntries() in sync
}
translator, err := LoadPlugin(ctx, "Translator", translatorCfg, mgr.Translator)
```

Every other slot in `initPlugins` is a uniform one-liner — `LoadPlugin(ctx, "X", cfg.X, mgr.X)` — because `initPlugins`'s job is to construct whatever the config says, nothing more. This block is different: it decides *what the config should have been* (defaulting `Translator` to the specific ID `jsonatatranslator`, conditioned on two unrelated slots) before handing off to `LoadPlugin`. That's a config-normalization decision, and as written it doesn't scale — every future consumer that wants a default `Translator` needs another `||` clause added here, permanently growing this one function.

`config.go`'s `PluginCfg.Translator` field already documents this default ("Defaults to jsonatatranslator when omitted"), so config.go already claims conceptual ownership — the mechanism just isn't there yet. Suggest moving the defaulting into a method on `PluginCfg`:

```go
// applyTranslatorDefault sets Translator to defaultTranslatorID when a
// consumer needs one but none was explicitly configured.
func (p *PluginCfg) applyTranslatorDefault() {
	if p.Translator == nil && (p.PayloadTransformer != nil || p.SchemaVersionMediator != nil) {
		p.Translator = &plugin.Config{ID: defaultTranslatorID}
	}
}
```

called once during config setup, so `initPlugins` collapses back to the same pattern as every other slot:

```go
translator, err := LoadPlugin(ctx, "Translator", cfg.Translator, mgr.Translator)
```

This keeps `stdHandler` doing only what it's meant to do — load what's configured — and puts the default-selection policy where it's already documented as belonging.

### Confirming `PayloadTransformerProvider` (`step.go`) is intentional, not redundant with `StepProvider`

Flagging this for the record since it came up in review: why not just reuse `StepProvider` for `reqmapper`? Checked, and the answer is that `StepProvider.New(ctx, config)` has no slot for the injected `Translator` dependency `reqmapper` now needs — Go interfaces are structurally typed by exact signature, so a dependency-free provider and a dependency-injecting provider for the same produced type (`Step`) can't be the same interface. This matches the codebase's existing convention: every dependent plugin has its own `*Provider` interface with its dependencies typed directly into `New` (`KeyManagerProvider` takes `RegistryLookup`, `RegistryLookupProvider` takes `Cache`, `ManifestLoaderProvider` takes `Cache` + `RegistryMetadataLookup`, etc.) — `PayloadTransformerProvider` taking `Translator` is the same pattern, not a one-off. Its placement in `step.go` (rather than `translator.go`) is also consistent with the file-per-produced-type convention used everywhere else (providers live with the type they *produce*, not the type they depend on). No change requested here — just confirming this part of the design is sound.

nirmay:
## Blocking: concurrent `Translate` calls on a shared cached expression can crash the process

`jsonatatranslator.Translate` (`jsonatatranslator.go:63`) calls `expr.Evaluate(payload, nil)` on a `jsonata.Expression` that's cached and shared across **every** concurrent caller translating the same artifact text (`compiledExpr`, L72-90). `TestTranslate_ConcurrentAccess` (present since the plugin's first commit, unchanged since) exercises exactly this — 20 goroutines calling `Translate` with the same expression — and it's flaky: it passed 30/30 in one run for me, but fatally crashed the whole process in two other runs, with:

```
fatal error: concurrent map writes
jsonata/v206.(*Frame).bind(...)
jsonata/v206.(*Expression).Evaluate(...)
jsonatatranslator.(*Translator).Translate(...)
```

That's a `fatal error`, not a panic — unrecoverable by `recover()`, it takes down the whole adapter process. I confirmed this isn't new to the latest push — it reproduces identically against the PR's base commit too.

**Two distinct causes, both inside `jsonata-go`'s `Evaluate`, triggered by sharing one compiled `*Expression` across concurrent callers:**

1. When `bindings == nil` (our call), `Evaluate` reuses the compiled expression's own static environment Frame directly and mutates its `bindings` map in place (`v206/jsonata.go` ~L395-397). Concurrent callers race on that shared map — this is the crash above. Passing a non-nil bindings map (even empty) avoids it, by making `Evaluate` allocate a fresh per-call frame instead.
2. Independent of bindings, `Evaluate` also does an unconditional `*e.timestamp = time.Now()` (`v206/jsonata.go:434`) through a pointer field owned by the compiled `Expression` itself. `go test -race` still flags this as a data race even after fix #1 is applied — there's no parameter on `Evaluate` that avoids it. This is a thread-safety gap in `jsonata-go` (`github.com/jsonata-go/jsonata@v0.0.0-20250709164031-599f35f32e5f`) for a *shared, cached* compiled expression, not something fixable from the call site alone.

**Proposed fix:** since #2 can't be worked around from the `Evaluate(...)` call, serialize access per cached expression (not globally, so unrelated expressions still evaluate in parallel) — store a mutex alongside each compiled expression in the cache:

```go
type exprCache struct {
	mu      sync.RWMutex
	entries map[string]*cachedExpr
	max     int
}

// cachedExpr pairs a compiled expression with a mutex serializing Evaluate
// calls on it. jsonata.Expression.Evaluate mutates state owned by the
// compiled Expression itself (see PR discussion) and is not safe for
// concurrent use on the same compiled Expression.
type cachedExpr struct {
	mu   sync.Mutex
	expr jsonata.Expression
}

func newExprCache(max int) *exprCache {
	return &exprCache{entries: make(map[string]*cachedExpr), max: max}
}
```

```go
func (t *Translator) Translate(_ context.Context, artifact []byte, payload []byte) ([]byte, error) {
	ce, err := t.compiledExpr(artifact)
	if err != nil {
		return nil, fmt.Errorf("jsonatatranslator: compile expression: %w", err)
	}
	ce.mu.Lock()
	defer ce.mu.Unlock()
	result, err := ce.expr.Evaluate(payload, map[string]interface{}{})
	if err != nil {
		return nil, fmt.Errorf("jsonatatranslator: evaluate expression: %w", err)
	}
	return result, nil
}

func (t *Translator) compiledExpr(artifact []byte) (*cachedExpr, error) {
	t.exprs.mu.RLock()
	ce, ok := t.exprs.entries[string(artifact)]
	t.exprs.mu.RUnlock()
	if ok {
		return ce, nil
	}

	expression := string(artifact)
	expr, err := t.instance.Compile(expression, false)
	if err != nil {
		return nil, err
	}
	ce = &cachedExpr{expr: expr}

	t.exprs.mu.Lock()
	if len(t.exprs.entries) < t.exprs.max {
		t.exprs.entries[expression] = ce
	}
	t.exprs.mu.Unlock()
	return ce, nil
}
```

Please re-verify with `go test ./pkg/plugin/implementation/jsonatatranslator/... -race -count=50` (or higher) rather than a plain `-count=1` run — the crash is intermittent, so one green run isn't sufficient evidence it's fixed.

Separately, worth filing an issue upstream against `jsonata-go` — root cause is `Expression.Evaluate` mutating the compiled Expression's own state (`e.timestamp`, and the static environment map when `bindings == nil`) instead of confining all mutation to a fresh per-call frame. Not something to block this PR on, just worth reporting.

nirmay (approved):
approved.

